PR #7, opened 2020-09-23
Pass on the HTML chapter

pavpanchekha:
This PR is a complete editing pass on the HTML chapter. Some of the major changes include:

- [x] The parser has been reorganized to more closely match a standard LR parser
- [x] Self-closing elements and doctypes are dealt with in the text instead of the exercises
- [x] Attributes are sort-of justified by the self-closing tags
- [x] The `Layout` object makes the recursive layout algorithm a lot cleaner
- [x] The author errors handled are now the implicit open and close tags: more useful than the prior example and better matching real HTML parsers.

### DIFF
--- a/book/html.md
+++ b/book/html.md
@@ -5,11 +5,6 @@ prev: text
 next: layout
 ...
 
-::: {.todo}
-- HTML attributes are unmotivated
-- I've dropped `<meta>` and `<link>` self-closing tags, probably should put them back somehow
-:::
-
 So far, your web browser sees web pages as a stream of open tags,
 close tags, and text. But HTML is actually a tree, and though the tree
 structure hasn't been important yet, you'll need it to draw
@@ -448,6 +443,10 @@ a slow but a sure process.
 
 HTML attributes
 ===============
+
+::: {.todo}
+- HTML attributes are unmotivated
+:::
 
 HTML tags have tag names but also *attributes*, which are added to an
 open tag to give additional information about that element. For example,
--- a/book/html.md
+++ b/book/html.md
@@ -233,7 +233,8 @@ def __init__(self, text):
 ```
 
 This code assumes all attributes have a value, but in fact the value
-can be omitted! If no value is given, it defaults to the empty string:
+can be omitted, like in `<input disabled>`. In this case, the
+attribute value is supposed to default to the empty string:
 
 ``` {.python indent=8}
 for attrpair in parts[1:]:
--- a/book/html.md
+++ b/book/html.md
@@ -21,6 +21,17 @@ for both open and close tags. But HTML is a tree, and each open and
 close tag pair are one node in the tree, as is each text token. We
 need to convert from tokens to nodes.
 
+We'll need a new `parse` function to do that:
+
+``` {.python}
+if __name__ == "__main__":
+    # ...
+    nodes = parse(lex(body))
+    browser = Browser()
+    browser.layout(nodes)
+    # ...
+```
+
 Let's start by defining the two types of nodes:[^1]
 
 [^1]: In reality there are other types of nodes too, like comments,
--- a/book/html.md
+++ b/book/html.md
@@ -157,10 +157,14 @@ This produces a more reasonable result:
  <link rel="stylesheet" href="../book.css" />]
 ```
 
-Why aren't these open elements closed? Well, most of them (like
+Why aren't these open elements closed?[^4] Well, most of them (like
 `<meta>` and `<link>`) are what are called self-closing: they don't
 need a close tag because they never surround content. Let's add a case
 for that to our parser:
+
+[^4]: Some people put a slash at the end of a self-closing tag (like
+    `<br/>`) but they don't have to: `<br>` is self-closing both with
+    and without that slash.
 
 ``` {.python indent=8}
 # ...
@@ -671,11 +675,3 @@ Exercises
 [^3]: Yes, it's a crazy system, and for a few years in the early '00s
     the W3C tried to [do away with it](https://www.w3.org/TR/xhtml1/).
     They failed.
-
-[^4]: Some people put a slash at the end of a self-closing tag (like
-    `<br/>`) but they don't have to: `<br>` is self-closing with and
-    without that slash.
-
-[^6]: If you implemented earlier exercises like support for `<a>`,
-    `<small>`, and `<big>`, you will likely have more parts to the
-    state.
--- a/book/html.md
+++ b/book/html.md
@@ -111,8 +111,8 @@ tag to begin with.
 
 [html5-after-body]: https://html.spec.whatwg.org/multipage/parsing.html#parsing-main-afterbody
 
-Parsing subtleties
-==================
+Self-closing tags
+=================
 
 Try running this parser on this page, and you'll find that `parse`
 doesn't return anything; let's find out why:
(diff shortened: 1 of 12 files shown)


book/todo.md
```diff
@@ -17,21 +17,22 @@ I am editing the chapters following the first iteration of the course.
...
 - [ ] The [Preliminaries](preliminaries.md) should install Python
-- [ ] [Chapter 3](text.md) should implement multiple text sizes
-- [ ] [Chapter 4](html.md) needs to motivate attributes
+- [x] [Chapter 3](text.md) should implement multiple text sizes
+- [x] [Chapter 4](html.md) needs to motivate attributes
+- [ ] [Chapter 5](layout.md)
 - [ ] [Chapter 7](chrome.md) should add forward buttons
-- [ ] [Chapter 10](reflow.md) needs a rewrite to handle a major bug
+- [ ] [Chapter 8](forms.md) needs minor edits
+- [ ] [Chapter 10](reflow.md) needs to fix a major bug
```
chrishtr, line 25: Is the bug explained somewhere? Write an issue on emberfox and link to it?

compare.py
```diff
@@ -101,14 +101,15 @@ def find_block(block, text):
...
+    failure, count = 0, 0
     for name, block in blocks:
         block = indent(block, name.get("indent", "0"))
         block = replace(block, *[item.split("/", 1) for item in name.get("replace", "/").split(",")])
         cng = find_block(block, src)
         expected = name.get("expected", "True") == "True"
+        count += 1
         if any(l2 for l2, l in cng) == expected:
-            count += 1
+            failure += 1
```
chrishtr, line 112: Comment why it's a failure

src/lab4.py
```diff
@@ -0,0 +1,276 @@
+import socket
+import ssl
+import tkinter
+import tkinter.font
+
```
chrishtr, line 5: Add a comment explaining briefly what this file is.

book/html.md
```diff
@@ -5,408 +5,559 @@ prev: text
...
-So far, our web browser understands web pages as a stream of HTML
-tokens: open tags, close tags, and text. It\'s worked for now, but
-that superficial view of HTML will hold us back when we try to draw
-backgrounds and borders, let alone implement CSS. HTML has a tree
-structure essential to those elements of style. So let\'s change our
-browser to parse HTML, and lay out the page, as a tree.
+So far, your web browser sees web pages as a stream of open tags,
+close tags, and text. But HTML is actually a tree, and though the tree
+structure hasn't been important yet, you'll need it to draw
+backgrounds, add margins, and let alone implement CSS. So this chapter
```
chrishtr, line 11:
```suggestion
backgrounds, add margins, and implement CSS. So this chapter
```

book/html.md
```diff
@@ -5,408 +5,559 @@ prev: text
...
+ <head>,
+ <meta charset="utf-8" />,
+ <meta name="generator" content="pandoc" />,
+ <meta name="viewport" content=... />,
+ <link rel="prev" href="text" />,
+ <link rel="next" href="layout" />,
+ <link rel="stylesheet" href="../book.css" />]
+```
 
-When you do, the best way to get more insight is to print the state of
```
chrishtr, line 130: Why remove this advice?

book/html.md
```diff
@@ -5,408 +5,559 @@ prev: text
...
-`</section>` without closing the `</p>`. Because these errors are so
-common, browsers try to automatically fix them so they do the \"right
-thing\". The full algorithm is pretty complicated (there are multiple
-types of tags and so on) but let\'s implement a simple version of it:
-when a close tag is encountered, we will walk up the tree until we find
-a tag that it closes.
+Attributes
+==========
+
+It doesn't help; why? Because the problem tags aren't just `<meta>`:
```
chrishtr, line 210: What do you mean by "it" here?

book/html.md
```diff
@@ -5,408 +5,559 @@ prev: text
...
+they have attributes, like in `<meta charset="utf-8" />`. HTML
+attributes give additional information about an element; open tags can
+have any number of attributes (though close tags can't have any).
+Attribute values can be either quoted, unquoted, or omitted entirely.
+
+For simplicity, let's stick to unquoted attribute values. Since
+neither tag names nor attribute-value pairs can contain
+whitespace,[^because-unquoted] we can split the tag contents on
```
chrishtr, line 218: The contents of an attribute value can contain spaces. Example: the sandbox attribute.
https://html.spec.whatwg.org/#sandboxing

chrishtr: You could say instead that for simplicity we are going to ignore attribute values that allow spaces, because those require handling quoted text.

book/html.md
```diff
@@ -5,408 +5,559 @@ prev: text
...
-values are optional; \"`<tagname attrname>`\" is perfectly valid and
-sets that attribute to an empty string.
+[html5-doctype]: https://html.spec.whatwg.org/multipage/syntax.html#the-doctype
 
-Let\'s extend `ElementNode` to store attributes:
+This special tag is called a [doctype][html5-doctype], and it's not
+really an element at all. It's not supposed to have close tag, but we
+can't mark it a self-closing tag—it's always the very first thing in
+the document, so there wouldn't be an open element to append it to.
+It's best to throw it away:
```
chrishtr, line 299:
```suggestion
It's best to throw it away for now (a real browser needs to parse it to handle various legacy or obscure parsing modes[1]):
```
[1] https://en.wikipedia.org/wiki/Quirks_mode

chrishtr: Oh I see you refer to quirks mode below. Maybe just foreshadow in a parenthetical?

book/html.md
```diff
@@ -5,408 +5,559 @@ prev: text
...
 
 ``` {.python}
-def layout_open(node):
-    if node.tag == "b":
-        state.bold = True
-    elif node.tag == "i":
-        state.italic = True
+HEAD_TAGS = [
+    "base", "basefont", "bgsound", "noscript",
+    "link", "meta", "title", "style", "script",
```
chrishtr, line 454: script can go in body; I guess it's only when these tags are encountered outside of body/head that this behavior happens?

book/html.md
```diff
@@ -5,408 +5,559 @@ prev: text
...
+        currently_open[-1].children.append(node)
 ```
 
+These rules for malformed HTML may seem arbitrary, and they are: they
+evolved over years of trying to guess what people "meant" when they
+wrote that HTML, and are now codified in the [HTML parsing
+standard](html5-parsing).
+
+::: {.further}
+HTML parsers also have an [algorithm][adoption] to handle misnested
```
chrishtr, line 507:
```suggestion
HTML parsers also have an [algorithm][adoption] to handle mis-nested
```

book/html.md
```diff
@@ -5,408 +5,559 @@ prev: text
...
 
-In this chapter, we taught our browser to understand HTML as a
-structured tree, not just a flat list of tokens, and we\'ve updated
-layout to be a recursive tree traversal instead of a linear pass through
-the document. We\'ve also made the browser much more robust to malformed
-HTML. While these changes don\'t have much impact yet, this new,
-structured understanding of HTML sets us up to implement a layout
-engine in the next chapter.
+This chapter taught our browser that HTML is a tree, not just a flat
+list of tokens:
```
chrishtr, line 520:
```suggestion
list of tokens. We added:
```

book/html.md
```diff
@@ -5,408 +5,559 @@ prev: text
...
+
+- A parser now transforms HTML tokens to a tree
+- Layout operates recursively on the tree
```
chrishtr, line 523:
```suggestion
- Layout operating recursively on the tree
```

book/html.md
```diff
@@ -5,408 +5,559 @@ prev: text
...
+- The lexer recognizes and handles attributes on elements
```
chrishtr, line 524:
```suggestion
- Lexer code to recognize and handle attributes on elements
```

(did you introduce what a "lexer" is?)

book/html.md
```diff
@@ -5,408 +5,559 @@ prev: text
...
+- Some malformed HTML documents are automatically fixed
```
chrishtr, line 525:
```suggestion
- Fixes for some malformed HTML documents
```

book/html.md
```diff
@@ -5,408 +5,559 @@ prev: text
...
+
+The tree structure of HTML is essential to lay out visually complex
```
chrishtr, line 527:
```suggestion
The tree structure of HTML is essential to display visually complex
```

book/html.md
```diff
@@ -5,408 +5,559 @@ prev: text
...
+*Paragraphs:* Since it's not clear what it would mean for one
+paragraph to contain another, the most common reason for this to
+happen in a web page is that someone forgot a close tag. Change the
+parser so that a document like `<p>hello<p>world</p>` results in two
+sibling paragraphs instead of one paragraph inside another.
+
+*Scripts:* JavaScript code embedded in a `<script>` tag uses the left
+angle bracket to mean less-than. Modify your lexer so that the
+contents of `<script>` tags are treated specially: no tags are allowed
+inside `<script>`, except the `</script>` close tag.
```
chrishtr, line 558: Fun fact: there are some weird restrictions on presence of the string "</script>" in an inline script, for reasons of preventing the HTML parser from needing to understand script syntaxes.

book/html.md
pavpanchekha: Yeah, it's for simplicity. In an earlier draft I pushed a 7-state parsing automaton to handle quoted values, but it was just too long. I'll clarify that.

book/html.md
pavpanchekha: Worth footnoting.

book/html.md
pavpanchekha: Yeah, but if you encounter script before either head or body is opened, HTML5 parsing will create a head node. HTML5 more or less enforces the requirement that every DOM tree starts with HTML, HEAD, and BODY, and that HEAD only contains one of those elements (except under noscript), but as you say BODY can also contain script and noscript.

book/html.md
pavpanchekha: Technically it's just `</script` followed by a space, tab, \v, \r, slash, or greater than sign. Yeah, I've run into this once or twice and it's a fun edge case. I talk about it in a video.

https://html.spec.whatwg.org/multipage/parsing.html#script-data-end-tag-name-state

book/todo.md
pavpanchekha: The bug is now issue #8, and means that reflow does not actually satisfy the reflow invariant when heights change!

book/html.md
pavpanchekha: You know, looking over the HTML5 spec again, maybe the right move here is to only drop whitespace text. HTML5 mandates that in standards mode (https://html.spec.whatwg.org/multipage/parsing.html#the-before-html-insertion-mode). And then implicit tags later in the chapter handle the case of text being inserted here.

pavpanchekha: Oh, sorry, I got the context wrong here. I'll add a footnote.

book/html.md
pavpanchekha: Good point! I call the function `lex` but don't think I ever called it a lexer.

book/html.md
```diff
@@ -517,14 +520,14 @@ Summary
 =======
 
 This chapter taught our browser that HTML is a tree, not just a flat
-list of tokens:
+list of tokens. We added:
 
 - A parser now transforms HTML tokens to a tree
```
chrishtr, line 525:
```suggestion
- A parser transforming HTML tokens to a tree
```

chrishtr (commented):
Ready for re-review?

pavpanchekha:
The last unclosed comment here is about debugging advice. I agree that the debugging advice is useful but in retrospect I think we should do a fuller job giving debugging advice, not just in this chapter but in prior ones as well. The existing debugging advices (deleted in this PR) is kind of vague, and I don't think students realistically ever executed it.

I opened #11 with some ideas on what debugging advice we should give.

pavpanchekha:
@chrishtr does the last comment address the debugging advice issue? Should we merge this?

book/html.md
```diff
@@ -5,408 +5,579 @@ prev: text
...
-[^4]: Some people put a slash at the end of a self-closing tag (like
-    `<br/>`) but they don\'t have to: `<br>` is self-closing with and
-    without that slash.
-
-[^5]: Where `href` stands for \"hypertext reference\".
-
-[^6]: If you implemented earlier exercises like support for `<a>`,
-    `<small>`, and `<big>`, you will likely have more parts to the
-    state.
+*Entities:* Implement support for `&amp;`, `&lt;`, `&gt;`, and
```
chrishtr, line 549: Rework this paragraph to refer back to the exercise I added in the other chapter, and
make this exercise about adding the other entities not in the other exercise (if the
student had already done the other exercise).

pavpanchekha: Great point. I ended up eliminating this exercise and replacing it with something I think is more interesting: turning your lexer into a syntax highlighter which you'd run for `view-source:` URLs.

chrishtr (approved):

PR #12, opened 2020-10-02
Pass on the Layout chapter

pavpanchekha:
This is a kind of first attempt at improving the layout chapter, but I'm not thrilled with it. The biggest changes from the last time around are:

- [x] Adding a conceptual section to the top of the page
- [x] Making the API of layout objects more similar to what we'll eventually need in the Reflow chapter.
- [x] Switching from borders to backgrounds as the main use of layout info
- [x] Removing the discussion of the box model

But this puts a lot of stuff in limbo.

- [ ] The conceptual section is a blunt instrument—ideally this info would be more clearly laced throughout the text, so the reader needn't keep the whole conceptual section loaded as they read.
- [ ] I don't really like parents setting childrens' positions before calling their layout method (should that be passed as an argument?)
- [ ] The box model still has to go somewhere; the most obvious place is either the Style or Chrome chapter.


book/styles.md
```diff
@@ -292,124 +418,141 @@ on the page.
...
 Downloading styles
 ==================
 
-This moves some code out of our browser into a plain CSS file, which
-is nice. But the goal is to go beyond just the browser styles; to do
-that, our browser needs to find website-specific CSS files, download
-them, and use them as well. Web pages call out their CSS files using
-the `link` element, which looks like this:
+Browsers get CSS code from two sources. First, each browser ships with a
+*browser style sheet*, which defines the default styles for all sorts of
```
chrishtr, line 506:
```suggestion
*user agent style sheet*, which defines the default styles for all sorts of
```

chrishtr: Did you say yet what a "user agent" is?

book/styles.md
```diff
@@ -440,117 +597,105 @@ slashes that come after `http:` in the URL, while in the last case,
...
+text nodes don't have any styles at all. How can that work?
 
-Here `x[0]` refers to the selector half of a rule, and I\'m calling
-the new `score` method. In Python, the `sort` function is *stable*,
-which means that things keep their relative order if possible. This
-means that in general, a later rule, from a later `<link>` or just
-later in a file, will override an earlier one, which is what CSS does
-as well.
+[^inline-margins]: Margins, borders, and padding can be applied to
+    inline layout objects in a real browser, but they work kind of a
```
chrishtr, line 643:
```suggestion
    inline layout objects in a real browser, but they work in a kind of
```

book/styles.md
```diff
@@ -560,115 +705,137 @@ def style(node, rules):
...
+"normal" with the Tk value "roman", and the `font-size` needs to be
+converted from points to pixels.[^72ppi]
 
-``` {.python}
-return node.tag in ["i", "b"]
-```
+[^72ppi]: Normally you think of points as a physical length unit (one
+    72^nd^ of an inch) and pixels as a digital unit (dependent on the
+    screen) but in CSS, the conversion is fixed at exactly 75% (or 96
+    pixels per inch). I'm not sure why, it seems weird and it does
```
chrishtr, line 735: The reason is to ensure existing sites are readable on all devices.

book/styles.md
```diff
@@ -560,115 +705,137 @@ def style(node, rules):
...
 ```
 
-With inheritance and the `display` property, we can move some more
-code into the CSS file.
+Another place where the code depends on specific tag names is
+`has_block_children`, which relies on a list of inline elements. The
+CSS `display`, which can be either `block` or `inline`, replaces that
+mechanism.^[Modern CSS adds way more values, like `run-in` or
+`inline-block` or `flex` or `grid`, and it has layout modes set by
+other properties, like `float` and `position`. Design matters.]
```
chrishtr, line 753: Not sure what you are trying to get across by saying "design matters".

book/styles.md
```diff
@@ -560,115 +705,137 @@ def style(node, rules):
...
+most of its domain-specific knowledge of tags is moved to the browser
+style sheet.
 
-::: {.todo}
-I think I'd like nodes to have a `font()` method that returns the font
-to use.
-:::
+That style sheet is easier to edit, since it's independent of the rest
+of the code. And while sometimes moving things to a data file means
+maintaining a new format. Here we get to reuse a format, CSS, that our
```
chrishtr, line 780:
```suggestion
maintaining a new format, here we get to reuse a format, CSS, that our
```

book/styles.md
```diff
@@ -560,115 +705,137 @@ def style(node, rules):
...
-    multiple space-separated selectors, like `ul strong`, which
-    selects all `<strong>` elements^[This is basically a `<b>` tag but
-    with a hipper name. The idea was to drop all visual aspects from
-    element names.] with a `<ul>` ancestor. Implement descendent
-    selectors, both in parsing and with a new `DescendentSelector`
-    class. Scoring for descendent selectors works just like in
-    `AndSelector`. Make sure that something like `section .warning`
-    selects warnings inside sections, while `section.warning` selects
-    warnings that *are* sections.
+*Shortcuts*: CSS "shortcut properties" set multiple related CSS
```
chrishtr, line 801: s/shortchut/shorthand/

book/styles.md
```diff
@@ -560,115 +705,137 @@ def style(node, rules):
...
+properties at the same time; `margin`, `padding`, `border-width`, and
+`font` are all popular shortcuts. Implement these four shortcut
```
chrishtr, line 803: Add examples, like "margin: 3px 4px 5px 6px" means top right bottom left.

book/styles.md
```diff
@@ -560,115 +705,137 @@ def style(node, rules):
...
+properties. If you do it in the `body` parsing function, you won't
+need to change the rest of the code. Start with the case where all of
+the subproperties are specified, then add default values.
```
chrishtr, line 806: by "default values", do you mean supporting the "auto" keyword?

book/styles.md
```diff
@@ -560,115 +705,137 @@ def style(node, rules):
...
+
+*Comma*: CSS allows a rule to have multiple, comma-separated
```
chrishtr, line 808: Is there an official terminology for this we can use instead?

book/styles.md
```diff
@@ -560,115 +705,137 @@ def style(node, rules):
...
+selectors, which is basically the same as multiple rules with the same
+body. Implement this and use it to shorten and shorten the browser
```
chrishtr, line 810: Give an example.

book/styles.md
```diff
@@ -560,115 +705,137 @@ def style(node, rules):
...
+`auto`, in which case the existing layout algorithm is used.
+
+*Percentages*: Most places where you can specify a pixel value in CSS,
+you can also write a percentage value like `50%`. When you do that for
+`margin`, `border`, or `padding` properties, it's relative to the
+layout object's width, while when you do it for `font-weight` it's
+relative to the parent's font size. Implement percentage values for
+all of these properties.
+
+*Combinations*: Sometimes you want to select an element by tag *and*
```
chrishtr, line 825: Same question here - better to use official terminology.

book/styles.md
```diff
@@ -560,115 +705,137 @@ def style(node, rules):
...
+class. You do this by concatenating the selectors without anything in
```
chrishtr, line 826:
```suggestion
class. You do this by concatenating the selectors without any characters (including whitespace) in
```

book/styles.md
```diff
@@ -560,115 +705,137 @@ def style(node, rules):
...
+between: `span.announce` selects elements that match both `span` and
+`.announce`. Implement a new `AndSelector` class to represent these
```
chrishtr, line 828: Add an example.

book/styles.md
```diff
@@ -560,115 +705,137 @@ def style(node, rules):
...
+and modify the parser to parse them. Sum priorities.[^lexicographic]
+
+[^lexicographic]: You're supposed to use lexicographic scoring for
+    these `AndSelector` things, but sums will work fine as long as no
+    one strings more than 16 selectors together.
+
+*Descendants*: When multiple selectors are separated with spaces, like
```
chrishtr, line 835:
```suggestion
*Descendant selectors*: When multiple selectors are separated with spaces, like
```

book/styles.md
```diff
@@ -560,115 +705,137 @@ def style(node, rules):
...
+`ul b`, that selects all `<b>` elements with a `<ul>` ancestor.
+Implement descendent selectors; scoring for descendent selectors works
+just like for comnbination selectors. Make sure that something like `section
+.warning` selects warnings inside sections, while `section.warning`
+selects warnings that *are* sections.
```
chrishtr, line 840: Another fun exercise:

*Ancestor selectors*: There is no way in CSS at present to do the inverse of descendant selectors. For example, there is no way to have a CSS rule that matches "all ul tags that contain a b in the subtree". Can you guess why? Try to implement ancestor selectors and analyze the speed of your algorithm.

book/styles.md
chrishtr: Historically, it seems to have gone like this:

1. most/all monitors had similar pixel densities
2. sites were build-and-forget
3. CSS introduced the 'zoom' property, which allowed scaling outside of the monitor, and laid bare the non-physicality of "physical" units
4. High-DPI/"retina" displays futher moved way from true "phyisical"

book/layout.md
```diff
@@ -5,529 +5,632 @@ prev: html
...
+
+![A tree of layout objects. The root is a block for the `<html>`
+    element; it has one child, a block for the `<body>` element; it
+    has three children, each blocks as well (for the `<h1>` and two
+    `<p>` elements); and each of those have a single child, an
+    inline.](/im/layout-modes.png)
+
+Finally: layout. With the layout tree created, how do we compute the
+size and position of each layout object? The general rule is that a
+block, like a paragraph, should take up as much horizontal room as it
```
chrishtr, line 71: say "greedy" somewhere in here

book/layout.md
```diff
@@ -5,529 +5,632 @@ prev: html
...
+can, and should be tall enough to contain everything inside it. That
+means that a layout objects's width is based on its *parent*'s width,
+while its *height* is based on its *children*'s height:
+
+![The flow of information through layout. Width information flows down
+    the tree, from parent to child, while height information flows up,
+    from child to parent.](/im/layout-order.png)
+
+This suggests a step-by-step approach to layout. First, an element
+must compute its width, based on its parent's width. That makes it
```
chrishtr, line 81: .. and its own style, which may override the width

book/layout.md
```diff
@@ -5,529 +5,632 @@ prev: html
...
+possible to lay out the children, which comes next. Finally, the
+children's heights are now available, so the element's height can be
+calculated.
```
chrishtr, line 84: Before python, I suggest some pseudocode:

```
Layout(parent_width) {
  my_width = ComputeWidth(parent_width, my_style)
  my_childrens_height = sum of Layout(children)
  my_height = ComputeHeight(my_childrens_height, my_style)
  return my_height
}

ComputeWidth(parent_width, my_style) {
  if (my_style prescribes a width)
   return width from my_style
  else
   return parent_width
}

ComputeHeight(my_childrens_height, my_style) {
  if (my_style prescribes a height)
    return height from my_style
  else
    return my_childrens_height
```

book/layout.md
```diff
@@ -5,529 +5,632 @@ prev: html
...
 
-Our `layout` function doesn\'t do this yet, so let's fix it.
+class BlockLayout:
+    # ...
+```
 
-Let's start by defining a data structure to store an area where text
-can be laid out. I\'m going to call this a `Block`:
+These layout objects will be the nodes of the layout tree, and in fact
+some browsers call it a layout tree. To make it a tree, we'll want
```
chrishtr, line 114: Remove this second part of the sentence? Not sure it adds value and avoids confusion.

book/layout.md
```diff
@@ -5,529 +5,632 @@ prev: html
...
-        if isinstance(node, ElementNode):
-            self.open(node)
-            for child in node.children:
-                self.layout(child)
-            self.close(node)
-        else:
-            self.text(node)
+    self.x = -1
+    self.y = -1
+    self.w = -1
```
chrishtr, line 154: s/w/width/

s/h/height/

book/layout.md
```diff
@@ -5,529 +5,632 @@ prev: html
...
-careful with this. Python in particular will not warn you if you
-assign to an undeclared variable, because it has no notion of variable
-declaration as separate from assignment!]
-
-As we're doing this, we need to update the hard-coded constants that
-used to define the page boundaries with references to `parent`. So
-instead of...
+Unfortunately, `InlineLayout` already uses the `x` and `y` fields for
+the location of the next word. To avoid a *very* annoying conflict,
+let's rename those fields to `cx` and `cy`. Instead of initializing
```
chrishtr, line 160: Choose something better than cx and cy. block_x_position?

book/layout.md
```diff
@@ -5,529 +5,632 @@ prev: html
...
+    child.x = self.x
+    child.y = y
+    child.layout()
+    y += child.h
 ```
 
-Since we're now looking for data in the parent, let's add a new type
-of node to be the root node:
+Note that the call to the child's `layout` method not only asks it to
+compute its size, but also to build more of the layout tree.
```
chrishtr, line 327: Build more fo the tree? Or recursively compute its layout?

book/layout.md
```diff
@@ -5,529 +5,632 @@ prev: html
...
-maxh = layout.h
-display_list = layout.display_list()
+class DocumentLayout:
+    def layout(self):
+        self.w = WIDTH
+        # ...
+        child.x = self.x = 0
+        child.y = self.y = 0
+        child.layout()
+        self.h = child.h
```
chrishtr, line 360: Or the height of the viewport instead. This makes it easier to explain how backgrounds of the root elements work, as well as clipping and scrolling of the "DocumentLayout".

book/layout.md
```diff
@@ -5,529 +5,632 @@ prev: html
...
+pass the list itself in as an argument, and have the method just
+append to that list instead of returning anything. For
+`DocumentLayout`, which only has one child, it looks like this:
 
 ``` {.python}
-if isinstance(node, TextNode):
-    return not node.text.isspace()
-# ...
+class DocumentLayout:
+    def draw(self, to):
```
chrishtr, line 400:
```suggestion
    def draw(self, display_list):
```

book/layout.md
```diff
@@ -5,529 +5,632 @@ prev: html
...
+        self.children[0].draw(to)
```
chrishtr, line 401:
```suggestion
        self.children[0].draw(display_list)
```

book/layout.md
```diff
@@ -5,529 +5,632 @@ prev: html
...
 ```
 
-We\'ll also need to skip these empty `TextNode` objects when we create
-layouts for each child of a block:
+For `BlockLayout`, which has multiple, `draw` is called on each child::
```
chrishtr, line 404:
```suggestion
For `BlockLayout`, which has multiple children, `draw` is called on each child::
```

book/layout.md
```diff
@@ -5,529 +5,632 @@ prev: html
...
 
 ``` {.python}
-for child in node.children:
-    if isinstance(child, TextNode): continue
-    # ...
+class BlockLayout:
+    def draw(self, to):
```
chrishtr, line 408: s/to/display_list/ etc.

book/layout.md
```diff
@@ -5,529 +5,632 @@ prev: html
...
         )
 ```
 
-Here I\'ve passed `width=0` to `create_rectangle`, because otherwise
-`create_rectangle` would draw a one-pixel black border.
-
-Now `InlineLayout.layout` will create `DrawText` objects, while
-`BlockLayout.display_list` will create `DrawRect` objects for the
-borders.
+We do still want to clip out graphics commands that only occur
```
chrishtr, line 537:
```suggestion
We do still want to omit graphics commands that only occur
```

book/layout.md
```diff
@@ -5,529 +5,632 @@ prev: html
...
+offscreen. `DrawRect` already contains a `y2` field we can use, so
+let's add the same `DrawText`:
```
chrishtr, line 539:
```suggestion
let's add the same to `DrawText`:
```

book/layout.md
```diff
@@ -5,529 +5,632 @@ prev: html
...
-    _ol, _ot = self.x, self.y
-    _or, _ob = _ol + self.w, _ot + self.h
-    _il, _it = _ol + self.bl, _ot + self.bt
-    _ir, _ib = _or - self.br, _ob - self.bb
-    if self.bl: dl.append(DrawRect(_ol, _ot, _il, _ob))
-    if self.br: dl.append(DrawRect(_ir, _ot, _or, _ob))
-    if self.bt: dl.append(DrawRect(_ol, _ot, _or, _it))
-    if self.bb: dl.append(DrawRect(_ol, _ib, _or, _ob))
-    return dl
+    self.y2 = y1 + font.metrics("linespace")
```
chrishtr, line 544: Rename y2 to a more descriptive name, etc.

book/layout.md
```diff
@@ -5,529 +5,632 @@ prev: html
...
 ```
 
-Here I define the variables `_ol`, `_il`, `_ot`, `_it`, `_or`, `_ir`,
-`_ob`, and `_ib` for the outer and inner left, top, right, and bottom
-coordinates. The underscore is because otherwise `or` would conflict
-with a keyword.
+::: {.quirks}
+On some systems, the `measure` and `metrics` commands are awfully
```
chrishtr, line 548: This is definitely true. Might be interesting to note that real browsers go to a lot of lengths to add extra caches and avoid doing work when not necessary because of this problem.

book/layout.md
```diff
@@ -5,529 +5,632 @@ prev: html
...
+        if cmd.y1 > self.scroll + HEIGHT: continue
+        if cmd.y2 < self.scroll: continue
+        cmd.draw(self.scroll, self.canvas)
 ```
 
-This setup also makes it fairly easy to change the appearance of
-elements. For example, a few lines will indent code blocks 8 pixels,
-surround them with a one pixel border, and then add another 8 pixels
-between the border and the code itself:
+Not only have access to the height of any element, but also of the
```
chrishtr, line 569:
```suggestion
We not only have access to the height of any element, but also of the
```

book/layout.md
```diff
@@ -5,529 +5,632 @@ prev: html
...
+variable:
 
 ``` {.python}
-elif node.tag == "pre":
-    self.mr = self.ml = 8
-    self.bt = self.br = self.bb = self.bl = 1
-    self.pt = self.pr = self.pb = self.pl = 8
+def layout(self, tree):
+    # ...
+    self.max_y = document.h
```
chrishtr, line 577: Isn't it height minus viewport height?

book/layout.md
```diff
@@ -5,529 +5,632 @@ prev: html
...
+gray bar naming the chapter and offering back and forward links. It is
+enclosed in a `<nav class="links">` tag. Have your browser give this
+links bar the same light gray background that any other browser does.
+
+*Hidden Head*: There's a good chance your browser is still showing
+scripts, styles, and page titles at the top of every page you visit.
+Make it so that the `<head>` element and its contents are never
+displayed.
+
+*Bullets*: Add bullets to list items, which in HTML are `<li>` tags.
```
chrishtr, line 622: This one is probably ok, but display:list-item is yet another beast in reality...

chrishtr:
Let me know when it's ready for re-review.

pavpanchekha (commented):
As I mentioned in our meeting, I think this branch is dead, but I've implemented most of the suggestions and will migrate the rest to our Google Doc before deleting the pull request.

PR #15, opened 2020-10-10
Update forms.md to fix some nits

chrishtr:


pavpanchekha (approved):
This looks good, but is the intention to totally remove the discussion of replaced elements, or to move them to a "Go Further" block, or to move it elsewhere?

chrishtr:
I think we should cover replaced elements elsewhere or in a Go Further. Filed an issue to track in #16.

PR #18, opened 2020-10-13
Ignore whitespace text nodes in the HTML parser

pavpanchekha:
This modifies the HTML parser to ignore all whitespace nodes. It's not a big change and it's a little weird, but I think it avoids other weird special cases later in the book.

### DIFF
--- a/book/html.md
+++ b/book/html.md
@@ -316,7 +316,7 @@ all text nodes that only contain whitespace:[^ignore-them]
 [^ignore-them]: Real browsers retain whitespace nodes: whitespace is
     significant inside `<pre>` tags and in some other cases. But our
     browser already renders `He<b>llo</b>` as two words, so let's just
-    say it's not a focus. Plus, ignoring all whitespace tags simplifies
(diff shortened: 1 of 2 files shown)


book/html.md
```diff
@@ -310,13 +310,20 @@ Furthermore, text can appear after the doctype declaration but before
...
 between `<!DOCTYPE html>` and `<html>` in the HTML source, so it's
-silly to have the parser crash here. Let's hand it by simply skipping
-text nodes when there aren't any currently-open elements:
+silly to have the parser crash here. Let's handle it just skipping
+all text nodes that only contain whitespace:[^ignore-them]
+
+[^ignore-them]: Real browsers retain whitespace nodes: whitespace is
+    significant inside `<pre>` tags and in some other cases. But our
+    browser already renders `He<b>llo</b>` as two words, so let's just
+    say it's not a focus. Plus, ignoring all whitespace tags simplifies
```
chrishtr, line 319:
```suggestion
    ignore that complication. Plus, ignoring all whitespace tags simplifies
```

chrishtr (approved):
Approved with suggestion.

PR #21, opened 2020-10-16
Code outlines

pavpanchekha:
This PR adds automated code outlines to book chapters. As the browser grows it starts getting confusing which classes and methods are supposed to exist, especially as stuff gets renamed and refactored. An outline of the code—at least including all the functions and all the classes—would help a lot.

- [x] Wrote Python script to convert chapter code to JSON outlines
- [x] Wrote Lua code run that Python script when building the book
- [x] Preliminary CSS

The way it all works is that a Lua filter searches for Markdown blocks with the `cmd` class, which it expects to contain a single line of code. It runs that code as a process, and inserts the results back into the document, either as raw HTML or as a code snippet. To get outlines, I use this functionality to run a little Python script, which uses Python's `ast` module to parse the "teacher's copy" of the chapter's code and convert it into a simple HTML structure. Then some CSS takes care of column breaking and indentation.

I'd like to explore styling options a little bit more. It would also be nice to get a list of fields on each class, and or it may be nice to merge identical classes.


chrishtr (commented):
Could you update http://browser.engineering/draft/ with this so I can see the output?

pavpanchekha:
Alright, I learned some more Lua, and the process here is a lot nicer. You can see these outlines live on the first three chapters in the draft mode, like here: https://browser.engineering/draft/text.html#outline

It's already helped me detect some unused constants.

chrishtr (approved):

PR #22, opened 2020-10-24
Fix makefiles to auto-create the www/blog/ and www/draft/ directories.

chrishtr:


pavpanchekha (approved):
I made the make rules more idiomatic by using the file name itself as the target.

Makefile
```diff
@@ -3,15 +3,18 @@ blog := $(patsubst blog/%.md,www/blog/%.html,$(wildcard blog/*.md))
 FLAGS=
 
 all: $(html) $(blog)
-draft: $(patsubst www/%.html,www/draft/%.html,$(html))
+draft: draftdir $(patsubst www/%.html,www/draft/%.html,$(html))
```
chrishtr, line 6: You can remove "draftdir" now.

pavpanchekha: Did it in `master`, good catch.

PR #24, opened 2020-12-04
Merge this year's changes for CS 6968

pavpanchekha:
This PR basically contains a semester's worth of changes to the book, which unfortunately might not really be reviewable. From memory,

- Chapter 5 (Layout) was totally rewritten, moving box models to the next section and changing the layout API to have parents set their children's x and y fields before doing layout.
- Chapter 6 (Styles) was totally rewritten, with the parser simplified and the box model stuff moved in from the previous chapter
- Chapter 7 (Chrome) gained a discussion of keyboard focus so that you could actually type into the address bar
- Chapter 8 (Forms) now had needed to support keyboard focus as well
- Chapter 10 (Reflow) was rewritten to fix the major bug it had before
- Chapter 9 and 11 (Scripts.& Security) received only the minimal changes needed to mesh the rest of the book.

Unfortunately, these changes are probably big enough that the best way to review them is rereading the relevant chapters, not working on the diff. Let's discuss over video chat.

### DIFF
--- a/book/chrome.md
+++ b/book/chrome.md
@@ -25,15 +25,17 @@ will define `handle_click` and bind it to `<Button-1>`, button number
 
 
 ``` {.python}
-window.bind("<Button-1>", handle_click)
+class Browser:
+    def __init__(self):
+        self.window.bind("<Button-1>", self.handle_click)
 
-def handle_click(e):
-    pass
+    def handle_click(self, e):
+        pass
 ```
 
 Inside `handle_click`, we want to figure out what link the user has
 clicked on. We'll need to look at the `e` argument, which contains an
-\"event object\". This object has `x` and `y` fields, which refer to
+"event object". This object has `x` and `y` fields, which refer to
 where the click happened, relative to the corner of the browser
 window. Since the canvas is in the top-left corner of the window,
 those are also the *x* and *y* coordinates relative to the canvas. To
@@ -41,19 +43,28 @@ get coordinates relative to the web page, we need to account for
 scrolling:
 
 ``` {.python}
-x, y = e.x, e.y + scrolly
+def handle_click(self, e):
+    x, y = e.x, e.y + self.scroll
 ```
 
 The next step is to figure out what links or other elements are at
-that location. Naively, this seems like it should be easy. We already
-have a tree of layout objects, each of which records size and
-position. We could use those to find the element clicked on, something
-like this:
+that location. To do that, we need to keep the tree of layout objects
+(which right now we create and then throw away) available as a field
+on the `Browser` class:
 
 ``` {.python}
-def handle_click(e):
-    x, y = e.x, e.y + scrolly
-    elt = find_element(x, y, mode)
+def layout(self, tree):
+    self.document = DocumentLayout(tree)
+    # ...
+```
+
+Now `handle_click` can walk this tree, checking each layout object's
+size and position to find the element clicked on:
+
+``` {.python}
+def handle_click(self, e):
+    # ...
+    elt = find_element(x, y, self.document)
     print(elt.tag)
 ```
 
@@ -68,48 +79,49 @@ def find_element(x, y, layout):
     if layout.x <= x < layout.x + layout.w and \
        layout.y <= y < layout.y + layout.h:
         return layout.node
-
 ```
 
 In this code snippet, I am checking the children of a given node
-before checking the node itself. That\'s because if you click on a
-link, you want to click on the link, not the paragraph that it's in.
-I search the children in reverse order in case children overlap; the
-last one would be "on top".[^2]
+before checking the node itself. That's because you want the most
+specific element: if you click on a link, you want to click on the
+link, not the page `<body>`. I search the children in reverse order in
+case children overlap; the last one would be "on top".[^2]
 
 [^2]: Real browsers use what are called *stacking contexts* to resolve
     the overlapping-elements question while allowing the order to be
     controlled with the `z-index` property.
 
-Let\'s test it---but actually, first, let\'s handle a silly omission:
-we don\'t have any special style for links! Let\'s quickly add support
+Let's test it---but actually, first, let's handle a silly omission:
+we don't have any special style for links! Let's quickly add support
 for text color, which is controlled by the `color` property:
 
--   First, I add `color` to `INHERITED_PROPERTIES`, and changed `style`
-    so that its default is `black`.
--   Next, I add the default style `a { color: blue; }` to our browser
-    style sheet to color links blue.
--   Finally, I add a `color` field to `DrawText` and fill it out in
-    `InlineLayout.text`.
--   Modified `DrawText.draw` to use the color for the `fill` parameter
-    of `create_text`.
+- First, add `color` to `INHERITED_PROPERTIES` with the default
+  value `black`.
+- Next, add `a { color: blue; }` to the browser style sheet so that
+  links (`<a>` tags) are colored blue.
+- Add a `color` field to `DrawText` and modify `DrawText.draw` to
+  use it for the `fill` parameter of `create_text`.
+- In `InlineLayout` pass the color from the `text` method, through the
+  `line` field, then into the `display_list` in `flush`, and finally
+  into the `DrawText` constructor.
 
 Once links have colors, you can actually *find* them on the page. So
 try clicking on them!
 
-... but it won\'t work. There is no layout object corresponding to a
-link. The link text is laid out by `InlineLayout`, but each
-`InlineLayout` handles a whole paragraph of text, not a single link.
-We\'ll need to do some surgery on `InlineLayout` to fix this.
-
 Adding line and text layout
 ===========================
 
-Here\'s how I want inline layout to work, at a high level:
+Unfortunately, if you click on a link you won't see `a` printed in the
+console. That's because there is no layout object corresponding to a
+link. The link text is laid out by `InlineLayout`, but each
+`InlineLayout` handles a whole paragraph of text, not a single link.
+We'll need to do some surgery on `InlineLayout` to fix this.
 
--   `InlineLayout`\'s children are a list of `LineLayout` objects. This
+Here's how I want inline layout to work, at a high level:
+
+-   `InlineLayout`'s children are a list of `LineLayout` objects. This
     list replaces `InlineLayout` keeping track of a `y` cursor position.
--   `LineLayout`\'s children are a list of `TextLayout` objects, one per
+-   `LineLayout`'s children are a list of `TextLayout` objects, one per
     word. Each `TextLayout` object has a `node`, which is always a
     `TextNode`. This replaces the `x` cursor position
 -   Both `TextLayout` and `LineLayout` objects have a `w` and an `h` fields.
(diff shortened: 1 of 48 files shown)


book/chrome.md
```diff
@@ -25,91 +25,102 @@ will define `handle_click` and bind it to `<Button-1>`, button number
...
+```
+
+Here the `find_layout` function is a straightforward variant of code
 we've already written a few times:
 
 ``` {.python}
-def find_element(x, y, layout):
-    for child in reversed(layout.children):
-        result = find_element(x, y, child)
+def find_layout(x, y, tree):
```
chrishtr, line 74: why rename this?

book/chrome.md
```diff
@@ -119,267 +130,178 @@ Here\'s how I want inline layout to work, at a high level:
...
+decide the parent later, in a separate `append` method, which I'll
+define below.
 
 [^4]: You may want to use inheritance to group all the `Layout`
-    classes into a hierarchy, but I\'m trying to stick to some kind of
+    classes into a hierarchy, but I'm trying to stick to some kind of
     easily-translatable subset of Python.
 
-[^5]: Because of line breaking. Note that in real browsers, even words can
-have line breaks in them, via hyphenation.
```
chrishtr, line 156: Why remove this?

book/chrome.md
```diff
@@ -119,267 +130,178 @@ Here\'s how I want inline layout to work, at a high level:
...
 Next, since each `TextLayout` corresponds to a particular `TextNode`, we
-can immediately compute its width and height. I\'m going make the height
-of a `TextNode` be just the line-space for its font, with the 0.2
-linespace added in `LineLayout`.[^6]
+can compute its font and based on that its width and height:[^6]
 
-[^6]: Make sure you measure `text`, not `node.text`, which contains
-    multiple words! That\'s an easy and confusing bug.
+[^6]: Make sure you measure `word`, not `node.text`, which contains
+    multiple words! That's an easy and confusing bug.
```
chrishtr, line 169:
```suggestion
    multiple words! That's an easy-to-make and confusing bug.
```

book/chrome.md
```diff
@@ -119,267 +130,178 @@ Here\'s how I want inline layout to work, at a high level:
...
-            self.text(node)
+        self.children.pop()
 ```
 
-This makes room for a new `layout` function, which calls `recurse` to
-create children and then lays them out:
+Now that words and lines lay themselves out, a lot of stuff disappears
+from`InlineLayout`. The `style`, `size`, `weight` fields are no longer
+needed. Neither is `line` or `cx`. And the display list is no longer
+needed: instead of addings words to a display list, they can print
```
chrishtr, line 262: Print themselves?

book/chrome.md
```diff
@@ -119,267 +130,178 @@ Here\'s how I want inline layout to work, at a high level:
...
+========================
 
 *Phew*. That was a lot of surgery to `InlineLayout`. But as a result,
-`InlineLayout` should now look a lot like the other layout classes.
-And, we now have individual layout objects corresponding to each word
-in the document. The `handle_click` function should now be working
-correctly: when you click on a link `find_element` should return the
-exact `TextNode` that you clicked on, from which you could get a link:
+`InlineLayout` should now look a lot like the other layout classes,
+and we now have individual layout object corresponding to each word in
```
chrishtr, line 280:
```suggestion
and we now have an individual layout object corresponding to each word in
```

book/chrome.md
```diff
@@ -389,155 +311,102 @@ That would mean:
...
 Browser chrome
 ==============
 
-Now that we are navigating between pages all the time, it\'s easy to
-get a little lost and forget what web page you\'re looking at.
-Browsers solve this issue with an address bar that shows the URL.
-Let\'s implement a little address bar ourselves.
+Now that we are navigating between pages all the time, it's easy to
+get a little lost and forget what web page you're looking at.
+Browsers solve this issue by with an address bar that shows the URL.
```
chrishtr, line 346:
```suggestion
Browsers solve this issue with an address bar that shows the URL.
```

chrishtr (commented):
Reviewed chrome.md

book/forms.md
```diff
@@ -13,349 +13,250 @@ out form information and then send that form to the server. Web forms
...
-Both `<input>` and `<textarea>` elements are inline content, like
-text, laid out in lines. So to support inputs we\'ll need a new kind
-of layout object, which I\'ll call `InputLayout`. It\'ll need to
-support the same kind of API as `TextLayout`, namely `attach` and
-`add_space`, so that it won't confuse `InlineLayout`:
-
-``` {.python}
+are for long, multi-line text. I'll implement `<input>` only, because
+because `<textarea>` has a lot of strange properties.[^sig-ws]
+Usually, web browsers communicate with the operating system and ask
```
chrishtr, line 29: Chrome does not do this..

book/forms.md
```diff
@@ -13,349 +13,250 @@ out form information and then send that form to the server. Web forms
...
+    def __init__(self, node):
         self.node = node
-        self.space = 0
-        self.multiline = multiline
+        self.children = []
+```
 
-    def layout(self, x, y):
-        pass
+These `InputLayout` objects need a `layout` method needs to compute
```
chrishtr, line 51:
```suggestion
These `InputLayout` objects need a `layout` method to compute
```

book/forms.md
```diff
@@ -771,48 +706,40 @@ is better with friends!
...
-    Clicks elsewhere should clear that field. Next, bind the `<Key>`
-    event in Tkinter and use the event's `char` field in the event
-    handler to determine the character the user typed. Add that
-    character the value of the element in `Browser.focus`. If there's
-    no focused element, don't do anything.[^4]
-
-[^4]: You can implement more features if you'd like, but it quickly
-    gets difficult. Backspace: doable; arrow keys: hard; selection:
-    crazy!
+*Check boxes*: Add check boxes. In HTML, check boxes `<input>`
```
chrishtr, line 709:
```suggestion
*Check boxes*: Add checkboxes. In HTML, checkbox `<input>`
```

book/forms.md
```diff
@@ -771,48 +706,40 @@ is better with friends!
...
+elements with the `type` attribute set to `checkbox`. The check box is
```
chrishtr, line 710:
```suggestion
elements with the `type` attribute set to `checkbox`. The checkbox is
```

book/forms.md
```diff
@@ -771,48 +706,40 @@ is better with friends!
...
+checked if it has the `checked` attribute set, and unchecked
+otherwise. Submitting check boxes in a form is a little tricky,
```
chrishtr, line 712:
```suggestion
otherwise. Submitting checkboxes in a form is a little tricky,
```

book/forms.md
```diff
@@ -771,48 +706,40 @@ is better with friends!
...
+though. A check box named `foo` only appears in the form encoding if
```
chrishtr, line 713:
```suggestion
though. A checkbox named `foo` only appears in the form encoding if
```

chrishtr (commented):
forms.md and layou.md

book/chrome.md
pavpanchekha: I think it used to return the element you clicked on, and now returns the layout object you clicked on. We need the layout object because reflow takes a layout object; not sure how it worked before.

book/chrome.md
pavpanchekha: It wasn't removed, just moved elsewhere in the document. My initial Markdown-ification of the book was via Pandoc, which puts footnotes at the bottom. I've tried to move them near where they're used instead.

book/forms.md
pavpanchekha: What's the right way to qualify this? "Many browsers..."? Maybe that's not even the right generalization here. Perhaps it should be

> Many applications want their input areas to match input areas for all other applications, so use operating-system libraries to draw an input area directly. That's possible in Tk, but browsers need a lot of control over application styling, so we'll be drawing input areas ourselves.

We can add a footnote saying that Safari and maybe Firefox do still sometimes use OS libraries for input areas, but need to detect when a web style makes that impossible.

book/chrome.md
pavpanchekha: Yeah, this needs rewording. Something like,

> And the display list is no longer needed: instead of an `InlineLayout` drawing all the words in a paragraph, each word now corresponds to a `TextLayout` object that can draw it.

book/chrome.md
chrishtr: How about find_layout_object?

book/forms.md
chrishtr: New text LGTM

book/chrome.md
chrishtr: LGTM

chrishtr (commented):
Reviewed: reflow.md, scripts.md.

book/chrome.md
pavpanchekha: That's probably better, but for code organization purposes I'd rather eliminate all of these `find_X` methods. Let's discuss next meeting how to get rid of them.

book/security.md
```diff
@@ -373,47 +398,48 @@ sent to all other websites. This means that if you log in on our guest
...
+cookie, they'd overwrite each other and you'd constantly be getting
 logged out!
 
 Web browsers use the *same origin policy* to determine which cookies
-are sent where. The rule is: if a cookie is set on one origin---where
-the origin is the scheme, host, and port---it can only be sent to
-servers on the same origin. Let\'s update our cookie policy to do
-this. I\'ll change the `cookies` field so it stores a map from origins
-to key-value pairs:[^10]
+are sent where. The rule is: if a cookie is only sent in HTTP requests
```
chrishtr, line 407:
```suggestion
are sent where. The rule is: a cookie is only sent in HTTP requests
```

book/security.md
```diff
@@ -636,80 +670,82 @@ This script waits for the user to click the button, and then replaces
...
-that don\'t come with this secret value. The nonce is like a login
-token, but instead of getting it on login, you get a new one every time
-you\'re presented with a form to fill out.
+that don't come with this secret value.
 
-Let\'s implement nonces in our guest book. First, we\'ll generate our
-secret nonce:
+A nonce is like a login token, but instead of getting it on login, you
+get a new one every time you're presented with a form to fill out. In
+some sense, it's like a cookie, except that it's not stored on a
```
chrishtr, line 691:
```suggestion
some sense, it's like a cookie, except that it's stored on a
```

chrishtr (commented):
Reviewed security.md

pavpanchekha:
I think that's all of them. Shall I merge? These chapters are far from done!

chrishtr:
Will finish reviewing tonight.

pavpanchekha:
Ok, no hurry, I probably missed something to review!

book/styles.md
```diff
@@ -21,85 +21,158 @@ The simplest mechanism for that is the `style` attribute on elements.
...
 
-Remember to write out the code to access the other 11 properties; the
-border one is called `border-top-width`, not `border-top`, but other
-than that, they're very repetitive.
+Remember to write out the code to access the other 9 properties, and
+don't forget that the border one is called `border-X-width`, not
+`border-X`.[^because-colors]
+
+[^because-colors]: Because borders have not only widths but also
+    colors and styles, while paddings are margins are thought of as
```
chrishtr, line 113:
```suggestion
    colors and styles, while paddings and margins are thought of as
```

book/styles.md
```diff
@@ -21,85 +21,158 @@ The simplest mechanism for that is the `style` attribute on elements.
...
+``` {.python}
+class InlineLayout:
+    def layout(self):
+        self.w = self.parent.w - self.parent.pl - self.parent.pr \
+            - self.parent.bl - self.parent.br
+```
+
+It's now possible to indent a single element by giving it a `style`
+attribute that adds a `margin-left`. But while that's good for one-off
+changes, but is a tedious way to change the style of, say, every
```
chrishtr, line 166:
```suggestion
changes, it is a tedious way to change the style of, say, every
```

book/styles.md
```diff
@@ -109,115 +182,177 @@ and `center`.] CSS was invented to improve on this state of affairs:
...
-browsers, for example.] Crashing on a parse error would mean web pages
-can't use a feature until all browsers support it, while skipping
-parse errors means a feature is useful once a single browser supports
-it. This is variously called "Postel's Law",^[After a line in the
-specification of TCP, written by Jon Postel] the "Digital
-Principle",^[After a similar idea in circuit design.] or the
-"Robustness Principle": produce maximally supported output but accept
-unsupported input.
+may support different syntaxes for property values.[^like-parens]
+Thanks to silent parse errors, web pages can't use features that only
```
chrishtr, line 342:
```suggestion
Thanks to silent parse errors, web pages can use features that only
```

book/styles.md
```diff
@@ -251,165 +386,209 @@ class IdSelector:
...
-We now want to apply the rules to change the `style` field of the
-`ElementNode` objects on the page; let's call the function that does
-that `style(tree, rules)`. Its logic is pretty simple:
+You can write `matches` for `TagSelector` and `IdSelector` on your
+own.
 
--   Recursively operate on every `ElementNode` in the tree
--   For each rule, check if the rule matches
--   If it does, go through the property/value pairs and assign them
+Now that we know which rules applies to an element, we need use their
```
chrishtr, line 470:
```suggestion
Now that we know which rules apply to an element, we need use their
```

book/styles.md
```diff
@@ -251,165 +386,209 @@ class IdSelector:
...
 Downloading styles
 ==================
 
-This moves some code out of our browser into a plain CSS file, which
-is nice. But the goal is to go beyond just the browser styles; to do
-that, our browser needs to find website-specific CSS files, download
-them, and use them as well. Web pages call out their CSS files using
-the `link` element, which looks like this:
+Browsers get CSS code from two sources. First, each browser ships with a
+*browser style sheet*, which defines the default styles for all sorts of
```
chrishtr, line 543:
```suggestion
*user agent style sheet*, which defines the default styles for all sorts of
```

book/styles.md
```diff
@@ -423,252 +602,305 @@ def find_links(node, lst):
     return lst
 ```
 
-Then, to turn a relative URL into a full URL:
+For each link, the `href` attribute gives a location for the
+stylesheet in question. The browser is expected to make a GET request
+to that location, parse the stylesheet, and use it. Note that the
+location is not a full URL; it is something called a *relative URL*,
+which can come in three flavors:^[There are more flavors, including
```
chrishtr, line 609:
```suggestion
which can come in three flavors:^[There are even more flavors, including
```

book/styles.md
```diff
@@ -423,252 +602,305 @@ def find_links(node, lst):
...
+query-relative and scheme-relative URLs, which I'm skipping.]
+
+-   A normal URL, which specifies a scheme, host, path, and so on
+-   A host-relative URL, which starts with a slash but reuses the
+    existing scheme and host
+-   A path-relative URL, which doesn't start with a slash and is
+    instead tacked onto the current URL (up but not past its last slash)
```
chrishtr, line 616: What does "not past its last slash" mean?

chrishtr (approved):
Approved % remaining suggestions.

PR #28, opened 2021-01-01
Adjust chapter 1 to make it more inclusive.

chrishtr:


pavpanchekha:
All good catches. I edited further. In the chapter head, I tried to retain some of the ironic tone—tell me if you think the new version is passable. For the comment on WIMP environment, I agree with killing the footnote (who cares!) but edited the text since it no longer needs to set up the footnote.

pavpanchekha:
Go ahead and merge if you agree with the edited text.

PR #29, opened 2021-01-01
Add subsection about mobile devices to the graphics chapter.

chrishtr:


pavpanchekha (approved):
Looks great! One question, though: should we put this at the start of the chapter, or at the end? The differences listed might make make more sense once they've talked at least a bit about implementation. Like for example when we say there's a touch screen instead of a keyboard, at the end of the chapter after scrolling is implemented we can refer to that directly. "Instead of scrolling with the arrow keys on mobile people scroll by dragging the screen."

chrishtr:
Good idea about moving to the end - done. Merging.

PR #39, opened 2021-01-19
Add simple Node script to allow debugging comparison with a real browser

chrishtr:


chrishtr:
If this approach LGTY, we can then follow up with instructions to students that they should match the formatted output
of this script, so that it's easy to unittest their browser by comparing with a real one.

pavpanchekha:
Ok, I got this running, fixed various bugs, and compared it to the HTML parser students write. Only major differences on the WBE pages (which are all well behaved and syntactically correct) are:

1. HTML entities
2. A weird bug where console.log sometimes cuts off printing?

But basically it produces output that can be directly compared to the book's parser, and it mostly passes.

pavpanchekha (approved):

PR #41, opened 2021-01-27
New HTML parser

pavpanchekha:
This closes most of the issues we had with the HTML parser, and reorganizes the entire thing.

### DIFF
--- a/book/html.md
+++ b/book/html.md
@@ -16,18 +16,15 @@ HTML parser and converts the layout engine to use it.
 A tree of nodes
 ===============
 
-Right now, the browser sees web pages as a flat sequence of tags and
-text, which is why the `Layout` object's `token` method includes code
-for both open and close tags. But HTML is a tree, with one node for
-each open and close tag pair and each span of text.[^1] Tokens need to
-evolve into nodes.
+The HTML tree has one node for each open and close tag pair and for
+each span of text.[^1] So for our browser to be a tree, tokens need to
+evolve into nodes. That means adding a list of children and a parent
+pointer to each one. Here's the new `Text` class:
 
 [^1]: In reality there are other types of nodes too, like comments,
     doctypes, and `CDATA` sections, and processing instructions. There
     are even some deprecated types!
 
-To make tokens into a tree, we need to add a list of children and a
-parent pointer to each. For `Text` nodes that looks like this:
 
 ``` {.python}
 class Text:
@@ -37,8 +34,8 @@ class Text:
         self.parent = parent
 ```
 
-We'll do the same for `Tag`. Since it takes two tags (the open and the
-close tag) to make a node, let's rename the `Tag` class to `Element`:
+Since it takes two tags (the open and the close tag) to make a node,
+let's rename the `Tag` class to `Element`, and make it look like this:
 
 ``` {.python expected=False}
 class Element:
@@ -48,12 +45,8 @@ class Element:
         self.parent = parent
 ```
 
-Constructing a tree of nodes from source code is called parsing, and
-it's a little more complex than `lex`, so we're going to want to break
-the process into several functions. To store the tree, and to keep all
-those functions organized, let's create a new `HTMLParser` class.
-
-A parser builds a tree one element or text node at a time. But that
+Constructing a tree of nodes from source code is called parsing. A
+parser builds a tree one element or text node at a time. But that
 means the parser needs to store an *incomplete* tree. For example,
 suppose the parser is has so far read this bit of HTML:
 
@@ -63,20 +56,27 @@ The parser has seen five tags (and one text node). The rest of the
 HTML will contain more open tags, close tags, and text; but no matter
 which tokens it sees, no new nodes will be added to the `<head>` tag,
 which has already been closed. So that node is "finished". But the
-other nodess are unfinished: more children can be added to the
-`<html>`, `<body>`, and `<h1>` nodes. Since the parser reads the HTML
-file from left to right, these unfinished tags have a certain
-structure. The unfinished tags have always been *opened* but not yet
-closed; they are always the *to the right* of the finished nodes; and
-they are always *children of other unfinished tags*.
+other nodes are unfinished: more children can be added to the
+`<html>`, `<body>`, and `<h1>` nodes, depending on what HTML comes
+next.
 
-To leverage these facts, let's represent an incomplete tree by storing
-a list of unfinished tags, ordered with parents before
-children.[^touch-last] Let's store this list in a new `unfinished`
-field:
+Since the parser reads the HTML file from left to right, these
+unfinished tags are always in a certain part of the tree. The
+unfinished tags have always been *opened* but not yet closed; they are
+always the *to the right* of the finished nodes; and they are always
+*children of other unfinished tags*. To leverage these facts, let's
+represent an incomplete tree by storing a list of unfinished tags,
+ordered with parents before children. The first node in the list is
+the root of the HTML tree; the last node in the list is the most
+recent unfinished tag.[^touch-last]
 
 [^touch-last]: In Python, and most other languages, it's faster to add
     and remove from the end of a list, instead of the beginning.
+
+Parsing is a little more complex than `lex`, so we're going to want to
+break it into several functions, organized in a new `HTMLParser`
+class. That class can also store the source code it's analyzing and
+the incomplete tree:
 
 ``` {.python}
 class HTMLParser:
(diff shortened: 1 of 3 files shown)


book/html.md
```diff
@@ -142,66 +224,130 @@ sibling to [troff][troff], now used for Linux man pages. The
...
+           '\n  '
 ```
 
-Ok, that's not too helpful. Python needs a method called `__repr__` to
-be defined to print things a little more reasonably:
+Immediately a couple of things stand out. Let's start at the top, with the `<!DOCTYPE html>` tag.
 
-``` {.python}
-class ElementNode:
+This special tag is called a [doctype][html5-doctype] that's s always the very first thing in an HTML document. But it's not really an element at all, nor is it supposed to have close tag. Our toy browser won't be using the doctype for anything, so it's best to throw it away:[^quirks-mode]
```
chrishtr, line 282:
```suggestion
This special tag is called a [doctype][html5-doctype] that's always the very first thing in an HTML document. But it's not really an element at all, nor is it supposed to have close tag. Our toy browser won't be using the doctype for anything, so it's best to throw it away:[^quirks-mode]
```

book/html.md
```diff
@@ -142,66 +224,130 @@ sibling to [troff][troff], now used for Linux man pages. The
...
- <html lang="en-US" xml:lang="en-US">,
- <head>,
- <meta charset="utf-8" />,
- <meta name="generator" content="pandoc" />,
- <meta name="viewport" content=... />,
- <link rel="prev" href="text" />,
- <link rel="next" href="layout" />,
- <link rel="stylesheet" href="../book.css" />]
-```
+Just throwing out doctypes isn't quite enough though---if you run your parser now, it will crash. That's because after the doctype comes a newline, which our parser treats as text and tries to insert into the tree. Except there isn't a tree, since there haven't been any open tags. For simplicity, let's just have our browser skip whitespace-only text nodes to side-step the problem:[^ignore-them]
```
chrishtr, line 297:
```suggestion
Just throwing out doctypes isn't quite enough though--if you run your parser now, it will crash. That's because after the doctype comes a newline, which our parser treats as text and tries to insert into the tree. Except there isn't a tree, since there haven't been any open tags. For simplicity, let's just have our browser skip whitespace-only text nodes to side-step the problem:[^ignore-them]
```

src/lab4.py
```diff
@@ -54,124 +54,125 @@ def request(url):
...
+        self.tag = tag
+        self.attributes = attributes
+        self.children = []
+        self.parent = parent
 
-class Tag:
-    def __init__(self, text):
+    def __repr__(self):
+        attrs = [k + "=\"" + v + "\"" for k, v  in self.attributes.items()]
+        return "<" + self.tag + " " + " ".join(attrs) + ">"
```
chrishtr, line 74: Optimize to avoid the extra space before the > or between / and >; looks ugly in the printed output and is not what real browsers or xml printers ever show.

book/html.md
```diff
@@ -142,66 +224,130 @@ sibling to [troff][troff], now used for Linux man pages. The
...
+headers, body = request(sys.argv[1])
+nodes = HTMLParser(body).parse()
+print_tree(nodes)
 ```
-[<__main__.ElementNode object at 0x101399c70>,  ...]
+
+Run it on this web page, and you'll see something like this:
+
+``` {.example}
+ <!DOCTYPE html >
```
chrishtr, line 268: Change code to get rid of these extra spaces before the >; left another comment in the code file for this as well.

book/html.md
```diff
@@ -142,66 +224,130 @@ sibling to [troff][troff], now used for Linux man pages. The
...
+    backwards-incompatible change in table cell vertical layout. Yes.
+    I don't need to make these up!
+
+[limited-quirks]: https://hsivonen.fi/doctype/
+
+Self-closing tags
+=================
+
+Elements like `<meta>` and `<link>` are what are called self-closing: you don't
+ever write `</meta>` or `</link>`, because these tags don't surround content. To get a reasonable-looking tree out of our parser, we'll need special support for them. In HTML, there's a [fixed list][html-void-elements] of these self-closing tags:[^void-elements]
```
chrishtr, line 350:
```suggestion
ever write `</meta>` or `</link>`, because these tags don't surround content. To get a reasonable-looking tree out of our parser, we'll need special support for them. In HTML, there's a [fixed list][html5-void-elements] of these self-closing tags:[^void-elements]
```

book/html.md
```diff
@@ -474,101 +584,95 @@ browsers insert them automatically. To do so, they compare the current
 token's tag to the list of currently open elements; that reveals
 whether any additional elements need to be created:
```
chrishtr, line 585:
```suggestion
whether any additional elements need to be created.
```

book/html.md
pavpanchekha: This one is correct, actually. `pandoc` converts triple-dash to an em dashes (for pauses in sentences) and double-dash to en dashes (for ranges).

pavpanchekha:
I think I addressed all of your comments, @chrishtr. Here's a few things it would be good to have your comment on before calling this done:

- Right now `print_tree` is not actually a pretty-printer—it adds quotes and doesn't print close tags. I should rewrite it to actually pretty print, right?
- Is the syntax highlighting exercise still doable? I like it quite a bit, but the lexer is now bound very tightly to the parser, so I'm not sure it's still doable.
- I'm going to do some automated testing on the parser using your Chrome debugging tools branch before committing. There might be easy ways to improve compliance.

chrishtr:
If the comments are addressed then LGTM!

Pretty print: Yes it should pretty-print and match the output from the node script.

Syntax highlighting: why is it not still doable exactly? I think it should be possible with a little more refactoring or some choice lines of code?

pavpanchekha:
It's doable but I think you'll need to copy-and-paste the lexer. And if you want to highlight attribute values separately then you'll need to merge it with `get_attributes`. Instead of reusing the lexer directly. That might be fine—I do like the syntax highlighting exercise.

chrishtr:
I think it's ok; let's put a task for us to do the exercises ourselves and verify how hard it is. We could make some small changes to the chapter code to make it a bit easier.

pavpanchekha:
Ok, I've done several passes over the text and it looks good. I've also tested against the Chrome parser and it works well enough on well behaved pages—that's probably all we can promise. I'm ready to ship.

book/html.md
```diff
@@ -8,205 +8,378 @@ next: layout
...
+print_tree(nodes)
 ```
-[<__main__.ElementNode object at 0x101399c70>,  ...]
+
+Run it on this web page, and you'll see something like this:
+
+``` {.example}
+ <!DOCTYPE HTML>
+   '\n'
+   <HTML LANG="EN-US" XML:LANG="EN-US">
```
chrishtr, line 283: All uppercase? sigh.

book/html.md
```diff
@@ -8,205 +8,378 @@ next: layout
...
+
+``` {.example}
+<HTML LANG="EN-US" XML:LANG="EN-US">
+   <HEAD>
+     <META CHARSET="UTF-8" />
+       <META NAME="GENERATOR" CONTENT="PANDOC" />
+         <META NAME="VIEWPORT" CONTENT="WIDTH=DEVICE-WIDTH,INITIAL-SCALE=1.0,USER-SCALABLE=YES" />
+           <META NAME="AUTHOR" CONTENT="PAVEL PANCHEKHA &AMP; CHRIS HARRELSON" />
+             <LINK REL="STYLESHEET" HREF="BOOK.CSS" />
+               <LINK REL="STYLESHEET" HREF="HTTPS://FONTS.GOOGLEAPIS.COM/CSS?FAMILY=VOLLKORN%7CLORA&DISPLAY=SWAP" />
```
chrishtr, line 346: URLs are case-sensitive. I think you will need to remove the upppercase code in the parser.

book/html.md
```diff
@@ -296,168 +472,140 @@ if "=" in attrpair:
...
-to handle poorly-written, confusing, boilerplate-less HTML.
-
-In fact, modern HTML parsers are capable of transforming *any* string
-of characters into an HTML tree, no matter how confusing the markup.[^3]
-The full algorithm is, as you might expect, complicated beyond belief,
-with dozens of ever-more-special cases forming a taxonomy of human
-error, but one of the nicer time-saving innovations is *implicit* tags.
+The parser now handles HTML pages correctly—at least when the HTML is
+written by the sorts of goody-two-shoes programmers who remember the
+`<head>` tag, close every open tag, and make their bed. Mere mortals
```
chrishtr, line 595:
```suggestion
`<head>` tag, close every open tag, and make their bed in the morning. Mere mortals
```

pavpanchekha:
The upper-case code was in the pretty-printer, not the parser, luckily. I added it so I could test against the headless Chrome, but in retrospect it's better for the case-conversion code to live in that script, not in the book's code.

pavpanchekha:
Please merge if you think this is ready!

PR #43, opened 2021-02-06
Fix types and shorten subsection names in the "Browsers and the Web" chapter

chrishtr:


chrishtr:
Simple enough to merge without review, can fix suggestions post-hoc.

PR #48, opened 2021-02-17
Rename remaining numbered footnotes

chrishtr:


chrishtr:
This is mechanical, so just committing without review.

PR #49, opened 2021-02-17
Should this be here?

pavpanchekha:
This paragraph was added in #29 but is probably a mistake (it is unfinished). @chrishtr should we delete it, or do you have the other half of the paragraph?

### DIFF
--- a/book/http.md
+++ b/book/http.md
@@ -583,13 +583,6 @@ not recommended. It's very difficult to write a custom TLS
 implementation that is not only correct but secure.
 :::
 
-A reference implementation
-==========================
-
-There are of course many implementations already of the HTTP protocol. You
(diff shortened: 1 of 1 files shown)


book/http.md
```diff
@@ -583,13 +583,6 @@ not recommended. It's very difficult to write a custom TLS
 implementation that is not only correct but secure.
 :::
 
-A reference implementation
-==========================
-
-There are of course many implementations already of the HTTP protocol. You
-should take advantage of this to use one of the existing implementations
-as a *reference implementation*, in order to test the whether your
```
chrishtr, line 591: Oops. this was meant to be finished off with statements like comparing your implementation with curl. Let's just delete it for now.

PR #50, opened 2021-02-19
Change layout to reference the previous sibling

pavpanchekha:
This changes how the layout tree is linked and the protocol for computing layout. Basically, in addition to a parent and a list of
children, each layout object now also has a pointer to its previous sibling. That way, to compute its Y-position, a layout object can reference its previous sibling's Y-position and height.

This might seem more complex than having the parent keep track of a vertical cursor for children, but the advantage of this approach is that computing the layout of each layout object is totally self-contained to a single `layout` function, instead of
responsibilities being split between that node's and its parent's `layout` functions. It also means that the recursive calls to
`layout` are purely that---calling `layout` on each child---instead of being interlaced with code that tracks a vertical cursor.

I've also added a paragraph that talks about dependencies and why the order is so important—this is easier to discuss now that it's purely an ordering problem within a single function. That could set us up to talk about dependencies better in the reflow chapter.

### DIFF
--- a/book/layout.md
+++ b/book/layout.md
@@ -123,7 +123,6 @@ def __init__(self, node, parent, previous):
     self.parent = parent
     self.previous = previous
     self.children = []
-    # ...
 ```
 
 The `InlineLayout` constructor also initializes `weight`, `style`, and
@@ -176,7 +175,7 @@ of layout object:
 
 ``` {.python}
 class BlockLayout:
-    def __init__(self, node, parent):
+    def __init__(self, node, parent, previous):
         self.node = node
         self.parent = parent
         self.children = []
@@ -399,7 +398,8 @@ tree and lay it out:
 
 ``` {.python}
 class Browser:
-    def layout(self, tree):
+    def layout(self, body):
+        tree = HTMLParser(body).parse()
         document = DocumentLayout(tree)
         document.layout()
 ```
@@ -446,7 +446,7 @@ variable:
 
 ``` {.python}
 class Browser:
-    def layout(self, tree):
+    def layout(self, body):
         # ...
         self.display_list = []
         document.draw(self.display_list)
@@ -594,7 +594,7 @@ past the bottom of the page. In `layout`, store the height in a
 `max_y` variable:
 
 ``` {.python}
-def layout(self, tree):
+def layout(self, body):
     # ...
     self.max_y = document.h - HEIGHT
 ```
@@ -606,7 +606,7 @@ bottom of the page:
 def scrolldown(self, e):
     self.scroll = self.scroll + SCROLL_STEP
     self.scroll = min(self.scroll, self.max_y)
-    self.scroll = max(0, self.scroll))
+    self.scroll = max(0, self.scroll)
     self.render()
 ```
 
--- a/book/layout.md
+++ b/book/layout.md
@@ -113,8 +113,8 @@ class BlockLayout:
 ```
 
 These layout objects will be the nodes of a *layout tree*. So we'll
-want layout objects to have children, a parent, and an HTML element
-that they correspond to. Each layout object will also have a layout to
+want layout objects to have references to their children, a parent, and an HTML element
+that they correspond to. Each layout object will also have a reference to
 its previous sibling in the tree.
 
 ``` {.python}
@@ -144,7 +144,7 @@ def layout(self):
     self.flush()
 ```
 
-Besides the references to other layout object, each layout object also
+Besides the references to other related layout objects, each layout object also
 needs a size and position, which we'll store in the `w`, `h`, `x`, and
 `y` fields.
 
@@ -331,7 +331,7 @@ object can be computed:
 self.h = sum([child.h for child in self.children])
 ```
 
-Note how important the order is here. A block layout's width depends
+Note how important the order is here. A block layout object's width depends
 on the parent's width; so by symmetry the width must be computed
 before laying out the children. The position is the same: it has to be
 computed before recursing so it's already available when the children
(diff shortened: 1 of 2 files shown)


book/layout.md
```diff
@@ -114,14 +114,15 @@ class BlockLayout:
 
 These layout objects will be the nodes of a *layout tree*. So we'll
 want layout objects to have children, a parent, and an HTML element
-that they correspond to:
+that they correspond to. Each layout object will also have a layout to
```
chrishtr, line 117:
```suggestion
that they correspond to. Each layout object will also have a reference to
```

book/layout.md
```diff
@@ -114,14 +114,15 @@ class BlockLayout:
 
 These layout objects will be the nodes of a *layout tree*. So we'll
 want layout objects to have children, a parent, and an HTML element
```
chrishtr, line 116:
```suggestion
want layout objects to have references to their children, a parent, and an HTML element
```

book/layout.md
```diff
@@ -143,18 +144,9 @@ def layout(self):
...
-
-``` {.python}
-def __init__(self, node, parent):
-    # ...
-    self.x = -1
-    self.y = -1
-    self.w = -1
-    self.h = -1
-```
+Besides the references to other layout object, each layout object also
```
chrishtr, line 147:
```suggestion
Besides the references to other related layout objects, each layout object also
```

book/layout.md
```diff
@@ -233,8 +220,7 @@ whether it should have block or inline contents:
 def has_block_children(self):
     for child in self.node.children:
         if isinstance(child, TextNode):
-            if not child.text.isspace():
```
chrishtr, line 236: Why did these two lines get removed?

book/layout.md
```diff
@@ -246,11 +232,12 @@ create child layout objects:
 ``` {.python indent=4}
 def layout(self):
     if self.has_block_children():
+        previous = None
         for child in self.node.children:
-            if isinstance(child, TextNode): continue
```
chrishtr, line 250: Not skipping TextNode any more?

book/layout.md
```diff
@@ -295,56 +282,87 @@ Size and position
...
 
 ``` {.python}
-self.h = y - self.y
+self.h = sum([child.h for child in self.children])
 ```
 
-That settles `BlockLayout`; let's now work on `InlineLayout`. Its `x`
-and `y` will be set by its parent, so its `layout` just needs to set
-`w` and `h`:
+Note how important the order is here. A block layout's width depends
```
chrishtr, line 334:
```suggestion
Note how important the order is here. A block layout object's width depends
```

book/layout.md
```diff
@@ -295,56 +282,87 @@ Size and position
...
+of its children.
 
 ``` {.python}
 class InlineLayout:
     def layout(self):
+        self.x = self.parent.x
         self.w = self.parent.w
+
+        if self.previous:
+            self.y = self.previous.y + self.previous.h
```
chrishtr, line 360: Where did self.y get set in the old code? Don't see any line deletions in your PR for that.

book/layout.md
```diff
@@ -354,10 +372,10 @@ document always starts in the same place it's pretty simple:
 ``` {.python}
 class DocumentLayout:
     def layout(self):
-        self.w = WIDTH
         # ...
-        child.x = self.x = 0
-        child.y = self.y = 0
+        self.w = WIDTH - 2*HSTEP
```
chrishtr, line 376: Why the change here?

src/lab5.py
```diff
@@ -54,143 +54,149 @@ def request(url):
...
+            parent = self.unfinished[-1] if self.unfinished else None
+            node = Element(tag, attributes, parent)
+            self.unfinished.append(node)
+
+    HEAD_TAGS = [
+        "base", "basefont", "bgsound", "noscript",
+        "link", "meta", "title", "style", "script",
+    ]
+
+    def implicit_tags(self, tag):
```
chrishtr, line 154: Why all these changes in parsing? Did this PR get mixed up with the previous one you wrote somehow?

book/layout.md
pavpanchekha: Thanks to there no longer being whitespace-only TextNodes, after changes in Ch4.

book/layout.md
pavpanchekha: There no won't be whitespace-only TextNodes, which is what this was skipping

book/layout.md
pavpanchekha: In the previous version, `x` and `y` were set by one's parent, and since `InlineLayout` doesn't have any children (at this stage of the book) there isn't code for `y` to delete. Honestly that was super confusing.

book/layout.md
pavpanchekha: This is unrelated to the others, actually, but without padding there seems to be some kind of border that Tk adds by default which obscures some letters, and students wondered if that was a bug in their code.

pavpanchekha: The padding comes back next chapter anyway.

chrishtr: OK can you add a comment explaining what's going on?

book/layout.md
chrishtr: Got it, makes sense. A nice improvement.

src/lab5.py
pavpanchekha: I should have separated this, but yes, I had to "forward-port" the ch4 changes to the ch5 code snapshot before I made further changes.

chrishtr (approved):

Issue #20, opened 2020-10-16
Introduce Browser.load earlier


pavpanchekha:
The early chapters of the book thread together lexing, parsing, and so on in the entry point. This is confusing to some readers and also means that all of that is renamed to a method on `Browser` midway through the book. It seems like instead we should introduce `Browser.load` in Chapter 2, and pass the page source itself as a field on `Browser`s instead of as an argument to `layout`.

PR #56, opened 2021-02-25
Rename Browser.layout to Browser.load

pavpanchekha:
This forward parts Ch6 (Styles) needs a `load` method to download both content and stylesheets. But as an added bonus, renaming this method also makes sure that there are never two `layout` methods (on `Browser` and on `SomethingLayout`) that do different things.


chrishtr (approved):

pavpanchekha:
This closes #20

PR #58, opened 2021-02-26
Refactor `has_block_children` in the Layout chapter

pavpanchekha:
One confusing in the current version of Chapter 5 is how the browser chooses to use InlineLayout versus BlockLayout. The way it's implemented now, the browser uses `InlineLayout` as long as any child has `display: inline`. That's a problem because it means an extra character somewhere can convert a whole HTML subtree into a single confusing paragraph. Real browsers use anonymous block boxes to do this right, but they're hard to implement here because our layout objects refer to a single element, not a sequence of them like anonymous block boxes.

So this commit just inverts the heuristic from "any child has `display: inline`" to "all children have `display: inline`". Along the way it changes how `InlineLayout` objects work. In the current version a `BlockLayout` either has several `BlockLayout` children, or a single `InlineLayout` child. In the PR's version a `BlockLayout` has any number of children, each of which can be an `InlineLayout` or a `BlockLayout`.

This still mishandles cases like this:

    <p>This <span>is</span> <div>a block</div> inside a paragraph</p>

With the PR's version, there will effectively be four paragraph ("This", "is", "a block", and "inside a paragraph"). The current version will have one paragraph instead. The correct rendering has three paragraphs ("This is", "a block", "inside a paragraph"). I think it'll be less confusing, though I'm willing to hear that we just need to implement anonymous block boxes instead.

### DIFF
--- a/book/layout.md
+++ b/book/layout.md
@@ -195,14 +195,22 @@ objects. Let's focus on `BlockLayout`, since `InlineLayout` won't have
 children for now.[^but-later]
 
 [^but-later]: Later, in [Chapter 7](chrome.md), we'll make inline
-    layouts contain lines, which themselves contain words.
+    layouts contain lines, which will themselves contain words.
 
-Usually, a block layout has one block child per child in the element
-tree. But when you get to something like a paragraph, the children are
-like text, so the child layout object is a single inline layout. We
-can tell the difference by examining the children of the node we are
-laying out: some elements are only used inside running text, other
-elements are only used as blocks:
+When creating child layout objects, the main question is whether each
+child should be a `BlockLayout` or an `InlineLayout`. Basically,
+elements that just contains text, or maybe formatted text, should have
+an `InlineLayout`, but containers like `<div>` or `<header>` should
+have a `BlockLayout`. But what happens if an elements contains both
+text and something like a `<div>`?
+
+In real browsers, a somewhat complicated mechanism known as [anonymous
+block boxes][anon-block] is used, but in our toy browser we'll
+implement something a little simpler. We'll start with a list of
+elements used for formatting text, which should be inside an
+`InlineLayout`:
+
+[anon-block]:https://developer.mozilla.org/en-US/docs/Web/CSS/Visual_formatting_model#anonymous_boxes 
 
 ``` {.python}
 INLINE_ELEMENTS = [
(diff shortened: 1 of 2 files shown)


book/layout.md
```diff
@@ -195,14 +195,26 @@ objects. Let's focus on `BlockLayout`, since `InlineLayout` won't have
...
-tree. But when you get to something like a paragraph, the children are
-like text, so the child layout object is a single inline layout. We
-can tell the difference by examining the children of the node we are
-laying out: some elements are only used inside running text, other
-elements are only used as blocks:
+When creating child layout objects, the main question is whether each
+child should be a `BlockLayout` or an `InlineLayout`. Basically,
+elements that just contains text, or maybe formatted text, should have
+an `InlineLayout`, but containers like `<div>` or `<header>` should
+have a `BlockLayout`. But what happens if an elements contains both
```
chrishtr, line 204:
```suggestion
have a `BlockLayout`. But what happens if an element contains both
```

book/layout.md
```diff
@@ -195,14 +195,26 @@ objects. Let's focus on `BlockLayout`, since `InlineLayout` won't have
...
+text and something like a `<div>`?
+
+In real browsers, a somewhat complicated mechanism known as [anonymous
+block boxes][anon-block] is used, but in our toy browser we'll
+implement something a little simpler.[^how-anon] We'll start by
+listing the elements used for formatting text:
+
+[^how-anon]: Anonymous block boxes handle cases like an element
+that contains text, formatted text, and also a container. You can
```
chrishtr, line 213:
```suggestion
that contains text, formatted text, and also a block. You can
```

book/layout.md
```diff
@@ -213,31 +225,49 @@ INLINE_ELEMENTS = [
 ]
 ```
 
-A block layout object looks at its children's tags to determine
-whether it should have block or inline contents:
+To determine whether an element has an `InlineLayout` or a
+`BlockLayout` we compare its children against that list. Let's add a
+top-level `layout_mode` function that function looks through the
```
chrishtr, line 230:
```suggestion
top-level `layout_mode` function that looks through the
```

book/layout.md
```diff
@@ -213,31 +225,49 @@ INLINE_ELEMENTS = [
...
+children of a node and categorizes them as either "text", including
+formatting tags, or containers:
```
chrishtr, line 232: Oh. Is "containers" the terminology you're using for block elements? Is that a good name to use?

chrishtr (commented):
I'm pretty sure the right decision is *not* to implement anonymous boxes in this first explanation.
If there ends up being an "advanced layout" chapter later, it can go there.

Let's also add an exercise digging into it. Not sure, but maybe it's not impossibly hard for the
reader to try their hand at it?

book/layout.md
pavpanchekha: It's not. Do you have better ideas? Here's what I'm trying to do—I want to define what "block" and "inline" layout modes are for, and if I say "block layout mode is for block elements" it'll be confusing why this is even a question... "Containers" is pretty weird though, maybe it's best to talk about "blocks" and "text"?

chrishtr: Ok just add a sentence defining "container" earlier in the chapter, in the conceptual section?

book/layout.md
```diff
@@ -38,52 +38,53 @@ layout tree and then computing those sizes and positions.
...
+laid out with an inline layout mode, while the `<div>` will be laid
+out with a block layout mode. What we've been implementing so far in
+this book is the inline layout mode, so in this chapter we'll
+implement block layout mode.
+
+Second: the layout tree. Layout modes can nest: a web page can have an
+`<html>` in block layout that contains a `<body>` in block layout that
+contains a `<p>` in inline layout. So tree-based layout starts by
+traversing the HTML tree to build a _layout tree_, where each layout
+object in the tree is one of the layout modes.
```
chrishtr, line 60:
```suggestion
object in the tree is in one of the layout modes.
```

book/layout.md
```diff
@@ -38,52 +38,53 @@ layout tree and then computing those sizes and positions.
...
 ![The flow of information through layout. Width information flows down
     the tree, from parent to child, while height information flows up,
     from child to parent.](/im/layout-order.png)
 
 This suggests a step-by-step approach to layout. First, an element
-must compute its width, based on its parent's width. That makes it
-possible to lay out the children, which comes next. Finally, the
-children's heights are now available, so the element's height can be
-calculated.
+compute its position and width, based on its parent. Then it lays out
```
chrishtr, line 85:
```suggestion
computes its position and width, based on its parent. Then it lays out
```

book/layout.md
```diff
@@ -195,14 +196,24 @@ objects. Let's focus on `BlockLayout`, since `InlineLayout` won't have
...
+
+When creating child layout objects, the main question is whether each
+child should be a `BlockLayout` or an `InlineLayout`. Basically,
+elements that just contains text, or maybe formatted text, should have
+an `InlineLayout`, but containers like `<div>` or `<header>` should
+have a `BlockLayout`.
+
+What happens if an element contains both text and something like a
+`<div>`? In some sense, this is an error on the part of the web
+developer. Just like with implicit tags in [Chapter 4][html.md],
```
chrishtr, line 209:
```suggestion
developer. And just like with implicit tags in [Chapter 4][html.md],
```

chrishtr: Not grammatical, but follows the tone of the book?

book/layout.md
```diff
@@ -195,14 +196,24 @@ objects. Let's focus on `BlockLayout`, since `InlineLayout` won't have
...
+browsers use a repair mechanism to make sense of the situation. In
+real browsers, the mechanism is known as [anonymous block
```
chrishtr, line 211:
```suggestion
real browsers, the mechanism known as [anonymous block
```

chrishtr:
LGTM with those nits fixed.

PR #63, opened 2021-03-01
Third section fixes

chrishtr:


chrishtr (commented):
Ditto - committing now.

PR #68, opened 2021-03-02
Pass on "The Web in History"

pavpanchekha:
I tried to reorient this section around a single theme: how the web has changed, and how you can be involved in its evolution. I think this ties it better to the theme of the chapter. It's an important point.

A few big changes:

- I moved the paragraph about how the web relies on a lot of other technologies to be the intro this section. I think it works well.
- This section has a paragraph in here on web apps. I don't think it fits—we might want to move it to one of the later sections about history, but honestly I'm not sure where it fits.

PR #69, opened 2021-03-02
Totally reorganize the introduction

pavpanchekha:
This PR is parallel to an email I sent @chrishtr which proposes a reorganization of the intro. You can see the revised intro here:

https://browser.engineering/draft/intro.html

Unfortunately browsing the diff on Github is mostly uninformative. In short, this reorg tries to fix two problems:

- The four history sections aren't obviously connected to the rest of the chapter.
- The "Browser and me" section feels too long and a bit unfocused.

To fix these, the reorg:

- Refocuses "The web in history" to focus on change and how reading the book puts the reader into this historical process. I have a separate PR (#68) to do only this.
- Splits "the browser and me" in two. The paragraphs starting at "Even in real browsers" discuss the software engineering that goes into a web browser. Those are merged with the "How browsers evolve" section, which also talks about software engineering. That merged section is placed after the "The web in history" section, which introduces the idea of browsers changing, core principles like openness, and the reader's role in the process.
- Splits the remaining two history sections ("Technological precursors" and "The browser ecosystem") into their own chapter. Right now, editing these sections is an uncomfortable tension between shortening (because these sections are so detailed and historically focused that they lose the emotional pull of the chapter) and lengthening (because the history is interesting and there are lessons to elaborate on).

In short, the proposal is to reorganize the introductory chapter to look like this (https://browser.engineering/draft/intro.html) and to expand the history chapter, at least by moving many of the footnotes in line but also expanding them where it's interesting.

### DIFF
--- a/book/browsersandweb.md
+++ b/book/browsersandweb.md
@@ -270,9 +270,9 @@ form once all the pieces came together.
 
 In the early days, the Internet was a world wide network of computers, largely
 at universities, labs, and major corporations, linked by physical cables and
-communicating over bespoke, application-specific protocols. The early web built
-on this foundation. Web pages were files stored on specific computers, and web
-browsers used an application-specific protocol to request them. URLs for web
+communicating over application-specific protocols. The early web built on this
+foundation. Web pages were files in a specific format stored on specific
+computers, and web browsers used a custom protocol to request them. URLs for web
 pages named the computer and the file, and early servers did little besides read
 files from a disk. The logical structure of the web mirrored its physical
 structure.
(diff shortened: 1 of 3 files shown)


chrishtr (commented):
Seems good to split out. I agree that the parts you left in the intro suffice to make clear the central point I didn't want top lose - evolution of the web along with technology over time, and the reader's place in it. The detailed parts about the origin story are fun and interesting, but can be in a second chapter that avoids getting in the weeds and allowing the reader to forget the main point of the intro chapter.

chrishtr (approved):
Let's merge these two PRs and then I will read again and assess.

I think history.html is probably incomplete still as-is, but let's merge it and then iterate.

PR #76, opened 2021-03-10
Refactor the layout conceptual section

pavpanchekha:
This PR refactors the conceptual section of the layout chapter. Basically, the conceptual section focuses entirely on the idea of a layout tree, and the discussion of layout modes is moved into the next section. I also did substantial rearrangement in the size and position sections to highlight the dependency order idea.


book/layout.md
```diff
@@ -340,33 +314,36 @@ else:
...
-for child in self.children:
-    child.layout()
-```
-
-Note that the call to the child's `layout` method not only asks it to
-compute its size, but also to its own children to the layout tree and
-recursively call `layout` on them.
+These three computations have to go before the loop that calls
+`layout` on each child. After all, a block layout object's width
+depends on the parent's width; so by symmetry the width must be
```
chrishtr, line 319:
```suggestion
depends on the parent's width; so the width must be
```

pavpanchekha:
I redid the first paragraph in the "Layout Modes" section to clarify what "container" means and also to just be less repetitive.

book/layout.md
```diff
@@ -8,152 +8,132 @@ next: styles
...
+Different layout object lay out their children differently. Some
+elements contain text, like the paragraph above, and lay that text out
+horizontally in lines. But other elements contain those containers
+themselves. In the example bove, the body element contains a heading
+and some paragraphs, which are themselves containers for text. Those
+containers are stacked vertically, from top to bottom. Abstracting a
+bit, there are two *layout modes* here, two ways an element can be
+laid out relative to its children.[^or-equivalently] They're called
+block layout, for laying out containers, and inline layout, for laying
+out text, so we say the body element is in block layout mode (because
```
chrishtr, line 73:
```suggestion
out text.

 In the example, the body element is in block layout mode (because
```

book/layout.md
```diff
@@ -8,152 +8,132 @@ next: styles
...
+inline layout mode (because their children are text). In real browsers
+there are a lot of other layout modes too, but in this chapter we'll
+keep it simple.
+
+[^or-equivalently]: Or, equivalently, how an element is laid out
+relative to its siblings and parent. CSS tends to this equivalent
+perspective, though in some cases (like the `inline-block` display
+mode) it also defines the relationship to children.
+
+Let's have one class for each layout mode. Inline layout should be
```
chrishtr, line 84:
```suggestion
Let's have one class for each layout mode. InlineLayout should be
```

book/layout.md
```diff
@@ -8,152 +8,132 @@ next: styles
...
+familiar: it's what we've been implementing up until this point and
+corresponds to how text inside a paragraph are laid out in lines, left
+to right.[^in-english] Block layout is new to this chapter: it
```
chrishtr, line 87:
```suggestion
to right.[^in-english] BlockLayout is new to this chapter: it
```

book/layout.md
```diff
@@ -8,152 +8,132 @@ next: styles
...
-described by an [attribute grammar](wiki-atgram). Attribute grammar
-engines analyze dependencies between different attributes to determine
-the right order to calculate each attribute.
-:::
 
-[wiki-atgram]: https://en.wikipedia.org/wiki/Attribute_grammar
+For example, consider a web page with a body that contains a heading
+and three paragraphs:
+
+    <body>
```
chrishtr, line 41: I think it would be great, and instructive, to make this a real example page, perhaps in an iframe so that can also be loaded on its own. I will try this out in a PR once you've landed this one.

pavpanchekha: This is a great idea!

chrishtr (approved):

PR #77, opened 2021-03-14
Rename w to width, h to height

chrishtr:
Also, fix a bug in the `render()` function - it was referring to `y1` and `y2` instead of `top` and `bottom`.


chrishtr:
Merge conflicts resolved. Merge now?

PR #78, opened 2021-03-15
Add example pages demonstrating block and inline layout

chrishtr:


chrishtr:
Updated the example to use nested iframes, as we discussed.

chrishtr:
Rebased on top of your latest chapter updates.

pavpanchekha:
I've played around with this a bunch and these are great. We do need to think about styling for these examples.

- Why not have the code and the rendering side-by-side, with the caption on top? We'd need to do some responsive CSS for phones but I think it's worth it.
- Why have comments for "inline" and "block" in the second example?
- Perhaps move the "caption" out of the iframe? Then the "click to open in a full tab" text could be integrated into the caption, and plus the caption would have the same font and font size as the rest of the page.

I was also trying to figure out if there's a way to make it so that we don't need to make separate "pre" and "normal" versions of each page. What do you think of having a JavaScript widget generate the iframe from the source? I'm not sure this is actually a good idea though—it could be brittle and opening in a new tab would be hard to support. And doing the reverse (iframe contents to pre tag contents) has same-origin issues. (I tried using `srcdoc` but it's pretty cumbersome.)

What are your thoughts on examples vs widgets? The upside of runnable widgets is students could change the source code of the example, and see the new layout tree. With the example the input is fixed. With a widget it's hard to "open in a browser tab", though. We're still manually assigning colors to elements (instead of the browser determining the layout modes), so what are readers going to see from opening the example in their browser?

chrishtr:
> We do need to think about styling for these examples

Yes. I sort of intentionally kept it simple for now during first review.
 
> * Why not have the code and the rendering side-by-side, with the caption on top? We'd need to do some responsive CSS for phones but I think it's worth it.

Good ideas. I'll try that.

> * Why have comments for "inline" and "block" in the second example?

Sure I can remove that. Also remove from the first example I suppose?

> * Perhaps move the "caption" out of the iframe? Then the "click to open in a full tab" text could be integrated into the caption, and plus the caption would have the same font and font size as the rest of the page.

I think what you're suggesting here is to put the text like `This is a real
example! Play with it in <a href="layout-block-container-example.html" target=_blank>another
browser tab</a> to see how it behaves.</span>` in a regular markdown sentence right? Not sure what you mean about integrated, isn't the link already integrated?

> I was also trying to figure out if there's a way to make it so that we don't need to make separate "pre" and "normal" versions of each page. What do you think of having a JavaScript widget generate the iframe from the source? I'm not sure this is actually a good idea though—it could be brittle and opening in a new tab would be hard to support. And doing the reverse (iframe contents to pre tag contents) has same-origin issues. (I tried using `srcdoc` but it's pretty cumbersome.)

I think maybe I can make it work, will give it a shot.

> What are your thoughts on examples vs widgets? The upside of runnable widgets is students could change the source code of the example, and see the new layout tree. With the example the input is fixed. With a widget it's hard to "open in a browser tab", though. We're still manually assigning colors to elements (instead of the browser determining the layout modes), so what are readers going to see from opening the example in their browser?

Widget SGTM, will try it.

chrishtr:
I updated to a side-by-side widget that has some responsive design layout for phones. The link to new tab feature also keeps around the current state. Looks good? If so I'll do the other example and clean it up a bit.

chrishtr:
I went ahead and did the cleanup, and also upgraded the second example to a widget. How does it look now?

PR #80, opened 2021-03-19
Add an interactive, web-based version of the chapter 2 lab.

chrishtr:
1. Adds a proxy server for obtaining resources over HTTP to work around CORS
2. Ported lab2.py to lab2.js via canvas and a testbed HTML file.


chrishtr:
Updated to remove the proxy server and make the testbed interactive - it will print out results as it goes.

PR #81, opened 2021-03-23
Major refactor of layout chapter

pavpanchekha:
This PR combines some basic cleanups for the layout chapter with one big reorganization. The basic cleanups are:

- [x] Change from a list of inline elements to a list of block elements.
- [x] Fix text to refer to `width` and `height` fields
- [x] Many general rephrasings and improvements.

The big reorganization is moving `BlockLayout` ahead of `InlineLayout`. This changes the structure of the text quite a bit, but I think makes it cleaner, shorter, and easier to read.

### DIFF
--- a/book/layout.md
+++ b/book/layout.md
@@ -14,7 +14,7 @@ tree of elements is transformed into a tree of *layout objects*
 corresponding to the visual elements of the page. In the process,
 we'll add support for backgrounds to make our web pages more colorful.
 
-The Layout Tree
+The layout tree
 ===============
 
 Right now, our browser lays out an element by separately handling its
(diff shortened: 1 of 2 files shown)


pavpanchekha:
I did not end up making separate PRs because I remembered how that was confusing last time. Best way to review is probably to open the [draft page](https://browser.engineering/draft/layout.html) and read it anew.

book/layout.md
```diff
@@ -25,117 +25,124 @@ width and height, is never computed. That makes it pretty hard to draw
...
+        for child in self.node.children:
+            next = BlockLayout(child, self, previous)
+            self.children.append(next)
+            previous = next
 ```
 
-So this renamed `layout` method initializes `weight`, `style`, and
-`size` fields, as well as the `display_list`:
+The difference between _nodes_ and _layout objects_ can get confusing
+here. Each `child` comes from `node.children`, so it is a HTML node.
```
chrishtr, line 74: Your comments here might be creating as much confusion as they aim to solve, e.g. because self.children actually is layout objects whereas self.node.children are nodes. I was confused myself for about 30 seconds. How about instead putting\ some comments in the code? Something like:

def layout(self); 
  previous = None
  for child in self.node.children:
    // Note: child is a node, not a layout object! Likewise self.node.children is an array of nodes
    next = BlockLayout(...)
    self.children.append(next)
    previous = next

book/layout.md
```diff
@@ -25,117 +25,124 @@ width and height, is never computed. That makes it pretty hard to draw
...
 ```
 
-I've changed this code to initialize `cursor_x` and `cursor_y` from
-`x` and `y` instead of `HSTEP` and `XSTEP`. Make sure to make that
-change inside the `flush` method also.
+So we're now building a layout tree with one layout object per HTML
+node, with an extra layout object at the root. But before we move on
+to computing sizes and positions, we have to face an important truth:
+different HTML elements are laid out differently. They need different
+kinds of layout objects.
```
chrishtr, line 117:
```suggestion
kinds of layout objects!
```

book/layout.md
```diff
@@ -146,46 +153,51 @@ class InlineLayout:
...
+        self.cursor_y = self.y
+        self.line = []
+        self.recurse(self.node)
+        self.flush()
 ```
 
-Unlike with inline layout, the first job of `BlockLayout` is creating
-layout objects for all the containters inside it.
+I've also initialized `cursor_x` and `cursor_y` from `x` and `y`
+instead of `HSTEP` and `XSTEP`. Make that changes, and make a similar
```
chrishtr, line 171:
```suggestion
instead of `HSTEP` and `XSTEP`. Make those changes, and similar
```

book/layout.md
```diff
@@ -146,46 +153,51 @@ class InlineLayout:
...
+change inside the `flush` method too:
```
chrishtr, line 172:
```suggestion
changes inside the `flush` method too:
```

book/layout.md
```diff
@@ -146,46 +153,51 @@ class InlineLayout:
...
+        # ...
+```
 
-What happens if an element contains both text and something like a
-`<div>`? In some sense, this is an error on the part of the web
-developer. And just like with implicit tags in [Chapter 4][html.md],
-browsers use a repair mechanism to make sense of the situation. In
-real browsers, "[anonymous block boxes][anon-block]" are used, but in
-our toy browser we'll implement something a little simpler. 
+Inline layout object aren't going to have any children [for
```
chrishtr, line 182:
```suggestion
Inline layout objects aren't going to have any children [for
```

book/layout.md
```diff
@@ -146,46 +153,51 @@ class InlineLayout:
...
+now](chrome.md), so we don't need any code for that in `layout`. And
+let's again leave actually computing `x` and `y` and `width` and
```
chrishtr, line 184:
```suggestion
just as with block layout, let's leave actually computing `x` and `y` and `width` and
```

book/layout.md
```diff
@@ -201,113 +213,72 @@ BLOCK_ELEMENTS = [
...
+                return "block"
+        return "inline"
+    else:
+        return "block"
 ```
 
-`BlockLayout` can now create child layout objects by looping over its
-children, calling `layout_mode` to determine what type of layout
-object to create:
+The cases make sure text nodes get inline layout while empty elements
```
chrishtr, line 233: Not sure if this sentence is understandable. Is it necessary either? Maybe remove?

book/layout.md
```diff
@@ -201,113 +213,72 @@ BLOCK_ELEMENTS = [
...
+layout mode to use for each element:
 
 ``` {.python}
 class BlockLayout:
     def layout(self):
         previous = None
         for child in self.node.children:
             if layout_mode(child) == "inline":
-                previous = InlineLayout(child, self, previous)
+                next = InlineLayout(child, self, previous)
```
chrishtr, line 243: Is this fixing a bug in the old code?

book/layout.md
```diff
@@ -336,31 +304,25 @@ self.height = sum([child.height for child in self.children])
 
 But note that the height of a block layout depends on the height of
 its *children*. So, it must be computed after recursing, after the
-heights of its children are computed.
+heights of its children are computed. Getting this dependency order
+right is crucial: get it wrong, some layout object will try to read a
```
chrishtr, line 308:
```suggestion
right is crucial: get it wrong, and some layout object will try to read a
```

book/layout.md
```diff
@@ -473,26 +430,23 @@ class Browser:
...
+complex, and we're about to add more features. Stable foundations make
+for comfortable houses.
 
 Backgrounds
 ===========
 
-Tree-based layout gives every layout object a size and position. This
-capability is foundational[^for-what] but this already-complex chapter
-demands a simple and visually compelling demonstration: backgrounds.
+The layout tree is the central[^for-what] but one simple and visually
```
chrishtr, line 440: Needs grammar fixing.

chrishtr: Or rephrasing..

book/layout.md
```diff
@@ -576,28 +527,21 @@ class DrawRect:
         )
 ```
 
-We do still want to skip graphics commands that occur offscreen, and
-`DrawRect` already contains a `bottom` field we can use, so let's add
-the same to `DrawText`:
+By default `create_rectangle` draws a one-pixel black border, which
```
chrishtr, line 530:
```suggestion
By default, `create_rectangle` draws a one-pixel black border, which
```

book/layout.md
```diff
@@ -620,30 +575,29 @@ class Browser:
...
 def scrolldown(self, e):
-    self.scroll = self.scroll + SCROLL_STEP
-    self.scroll = min(self.scroll, self.max_y)
-    self.scroll = max(0, self.scroll)
+    self.scroll = min(self.scroll + SCROLL_STEP, self.max_y)
     self.render()
 ```
 
-Make sure those `max` and `min` calls happen in the right order!
+Thank you tree-based layout! In fact, as we'll see in the next two
```
chrishtr, line 587: Why thank you? Is there a convenient feature you're referring to?

If, not, perhaps "Well that's tree-based layout!"

book/layout.md
pavpanchekha: No—it's just clarifying. Before you create a new layout object by referring to the previous node, then store the new node in the `previous` variable. Now, you store it in a new `next` variable which is later saved into `previous`.

book/layout.md
chrishtr: WDYT of my comments?

book/layout.md
chrishtr: ditto

chrishtr (commented):
Some of my comments are still open, assuming you're going to get to them soon?

pavpanchekha:
The comments are good, I'm working on them.

chrishtr (approved):

PR #95, opened 2021-04-08
Provide link to GitHub on the index page

### LINKED ISSUE #92: Github Project Link on home page
It is common place for documentation to place a link to it's repo on the home page. currently it only exists on the preface.

DanielRosenwasser:
Fixes #92.

### DIFF
--- a/book/index.md
+++ b/book/index.md
@@ -14,6 +14,8 @@ This book is a [work in progress](todo.md). Use at your own risk.
 
 Follow this book's [blog](blog/) or
 [Twitter](https://twitter.com/browserbook) for updates.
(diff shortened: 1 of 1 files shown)


DanielRosenwasser:
Ah, I just saw this:

> I prefer to receive typos and small comments on the text using the book's built-in feedback tools, which you can enable with `Ctrl+E`.

I can close this if appropriate.

pavpanchekha:
Yeah, we would prefer feedback via the built-in tools and purposely don't advertise the Github link. We don't mind people looking over the source code and whatnot, but books written by committee are drier and less focused, which we want to avoid.

PR #99, opened 2021-04-11
Add summary conclusion and one exercise.

chrishtr:


pavpanchekha:
Love it this is great and very much in spirit of both the chapter and the book.

PR #100, opened 2021-04-13
Bugfixes in labs

chrishtr:
1. Handle <html> elements not preceded by other tokens correctly; in these cases, currently_open will be an empty array
2. Handle URLs not containing a / after the domain
3. s/self/this/ in runtime9.js
4. Output "comment9.js" instead of "comment.js" in server9.js
5. Output the id=errors div in server9.js, which is expected by the script
6. Wrong order of arguments to ElementNode constructor in lab6.py

### DIFF
--- a/book/html.md
+++ b/book/html.md
@@ -211,7 +211,7 @@ This is *almost* a complete parser, but it doesn't quite work at the
 beginning and end of the document. The very first open tag is an edge
 case without a parent:
 
-``` {.python}
+``` {.python indent=4}
 def add_tag(self, tag):
     # ...
     else:
@@ -557,7 +557,7 @@ class Layout:
 Now we need the `Layout` object to walk the node tree, calling `open`,
 `close`, and `text` in the right order:
 
-``` {.python}
+``` {.python indent=4}
 def recurse(self, tree):
     if isinstance(tree, Text):
         self.text(tree.text)
--- a/src/lab3.py
+++ b/src/lab3.py
@@ -14,7 +14,15 @@ def request(url):
     assert scheme in ["http", "https"], \
         "Unknown scheme {}".format(scheme)
 
-    host, path = url.split("/", 1)
+    host = ""
+    path = ""
+    if (url.find("/") >= 0):
+        host, path = url.split("/", 1)
+        print(host)
+    else:
+        host = url
+        print(host)
+
     path = "/" + path
     port = 80 if scheme == "http" else 443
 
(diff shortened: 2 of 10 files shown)


chrishtr:
The comparison script for lab4 seems to have gone rogue. It is recording diffs that aren't in my CL...any suggestion or idea?

pavpanchekha:
Yeah the differ could stand a rework now that I've built all this other infra for compiling and outlining. Anyway, fixed it by reminding it that these lines should be indented. I don't understand why it doesn't do that automatically but it helps sometimes.

pavpanchekha:
For the "URLs not containing a slash" I would write that code like this:

```
if (url.find("/") >= 0):
    host, path = url.split("/", 1)
else:
    host, path = url, ""
```

Is there a comparable part of Chapter 1 that we should update?

pavpanchekha:
The "elements not preceded by other ..." that doesn't happen in Ch4, right? The code for labs 6 onward aren't up to date with the changes so far in the book. (Sometimes the changes would be large which is why I don't want to do it before we get to those chapters.)

pavpanchekha:
Other things look good.

pavpanchekha (approved):

chrishtr:
> For the "URLs not containing a slash" I would write that code like this:
> 
> ```
> if (url.find("/") >= 0):
>     host, path = url.split("/", 1)
> else:
>     host, path = url, ""

Doesn't this have a scoping problem for the variables host and path?

> ```
> 
> Is there a comparable part of Chapter 1 that we should update?

I already fixed it in a previous PR.

pavpanchekha:
There's not a scoping problem in Python—"if" and "for" and so on don't introduce scopes. (I just implemented this in the compiler, since JS semantics are different!)

chrishtr:
> The "elements not preceded by other ..." that doesn't happen in Ch4, right? The code for labs 6 onward aren't up to date with the changes so far in the book. (Sometimes the changes would be large which is why I don't want to do it before we get to those chapters.)

Right. It was only broken for the later labs.

I came across this because I'm working on a lab for my new chapter 13, which needs to riff on chapter 9's code.

pavpanchekha:
Ok sounds good happy to merge this. If you'd like I can bring Ch9 up to date.

chrishtr:
I upda

> There's not a scoping problem in Python—"if" and "for" and so on don't introduce scopes. (I just implemented this in the compiler, since JS semantics are different!)

Got it. I fixed the code then.

PR #102, opened 2021-04-14
Fix href attribute for sequential link elements

anthony-geoghegan:
I use a browser extension (Tridactyl) that makes it easy to navigate backwards and forwards using the URLs specified by `<link>` elements. This commit add the `.html` suffix for `href `attributes of external resource links when the relationship indicates sequential access, i.e., “prev”, “next”.

### DIFF
--- a/book/template.html
+++ b/book/template.html
@@ -5,8 +5,8 @@
   <meta name="generator" content="pandoc" />
   <meta name="viewport" content="width=device-width,initial-scale=1.0,user-scalable=yes" />
 $for(author-meta)$  <meta name="author" content="$author-meta$" />$endfor$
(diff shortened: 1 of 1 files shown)


pavpanchekha:
Good catch, also what's the Tridactyl keybinding for this, please advise.

anthony-geoghegan:
> Good catch, also what's the Tridactyl keybinding for this, please advise.

It's `[[` and `]]` to go backwards and forwards, respectively.

I vaguely recall keyboard short-cuts for the same feature being available for browsers in the late 90s / early 2000s. I’m thinking of Netscape but it could have been Konqeror or Opera.

I was interested to see how you used `make` with Pandoc to generate the HTML so I was happy to be able to help make it better.

PR #107, opened 2021-04-24
Add images and acknowledgements for wikipedia and web builders

chrishtr:


pavpanchekha:
Love the acks. Let's merge this and maybe I'll work on better styling for images (maybe we can make them expandable? Perhaps they could live in the margins?) later.

PR #109, opened 2021-04-25
Add first pieces of the rendering pipeline chapter

chrishtr:
This is the first part of the chapter. There is more to come after this.


pavpanchekha:
This is a great start to the chapter, and the TODO at the end felt like a cliffhanger 😄. Can't wait the read the rest. Here are my quick notes—we can discuss more on Friday.

The highest order bit is, this is a really great chapter. My take-away from the chapter so far is that the "tasklist" architecture is the central organizing principle for the chapter. A task list, multiple kinds of tasks, flags for what sorts of tasks need to be done, and careful logic around copying and clearing the queue are all interesting technical parts of that task. Combining that with multithreading in the second half of the chapter will work very naturally.

Some smaller comments:

* I think introducing the "tasklist" architecture is a really strong central organizing principle for the chapter
  * The discussion of RAF calls from RAF callbacks (in a footnote) should move to the main text
  * It's interesting and subtle stuff and it directly relates to the "tasklist" architecture
* Would be nice to move some of this out of the Browser code
  * Perhaps a set of Task classes?
  * Perhaps instead of the needs_XXX flags we prioritize the list of tasks?
* The flashback to Ch2 at the beginning feels unmotivated
  * Basically the first half of the chapter is about the architecture, which doesn't directly contribute to speed.
  * This will be easy to reorganize once the second half the chapter is in
* The network thread also seems unmotivated. My guess, when the rest of the chapter is done:
  * The task list + set of task classes will be the first half of the chapter
  * The network thread will move to be near a discussion of the compositor thread and be part of a threading discussion
  * Perhaps we should add async download of CSS and re-lay-out when it's downloaded? Would stress both threading and tasklist.
* I think this would make more sense as one chapter, but with architecture + threading + performance it could grow too long.
  * I don't think we can make that decision until you've written the second half—it'll probably be obvious then.

And the smallest comments:

* We should make the "render" -> "draw" change throughout the book. Consistent terminology!
* Should we put units into constant names? WIDTH_PX, REFRESH_RATE_MS and similar?
* Datetime "now" function is not introduced on the Python side

pavpanchekha:
By the way feel free to merge and deal with the above comments later—we can also move them into our longer doc. Merging keeps everything simple and sane.

chrishtr:
Your larger comments make sense and I agree that it'll be a lot easier to reorder the chapter once the whole picture is there.

> * We should make the "render" -> "draw" change throughout the book. Consistent terminology!

Agreed, that had been on my list to do already.

> * Should we put units into constant names? WIDTH_PX, REFRESH_RATE_MS and similar?

Yes. Done.

> * Datetime "now" function is not introduced on the Python side

It is exposed - see js_now. Did you mean that it isn't mentioned in the chapter text?

PR #117, opened 2021-05-11
Rendering architecture chapter part 2

chrishtr:
This PR completes the text for a first draft of the rendering architecture chapter. I fully expect to do a bunch of refactoring later to improve the code, order of explanation of concepts, etc.


pavpanchekha:
I read through it and it looks really great. I'm very excited about adding this chapter to the book and look forward to talking about it on Friday.

My highest level bit of feedback was that splitting the browser into two threads goes by pretty fast and is confusing. This might be helped by making a `TaskList` with `Task`s on it early in the chapter; later in the chapter `TaskList` could become `MainThreadRunner` later on.

I think the discussion of performance here is good, but I'm not totally convinced we can delay the font cache this long. It becomes painful almost immediately in Chapter 3, and we probably need to go back and add it then.

chrishtr:
> I read through it and it looks really great. I'm very excited about adding this chapter to the book and look forward to talking about it on Friday.

Cool. If you don't mind I'll commit this text for now and refactor it.

> My highest level bit of feedback was that splitting the browser into two threads goes by pretty fast and is confusing. This might be helped by making a `TaskList` with `Task`s on it early in the chapter; later in the chapter `TaskList` could become `MainThreadRunner` later on.

Agree.

> I think the discussion of performance here is good, but I'm not totally convinced we can delay the font cache this long. It becomes painful almost immediately in Chapter 3, and we probably need to go back and add it then.

Sure, ok. I'm sure I can find some other way to demonstrate the performance problems.

PR #118, opened 2021-05-12
(WIP) Pass over the style chapter

pavpanchekha:
This PR is a pretty big revision to the style chapter. Some highlights include:

- [x] Moving the box model discussion to the next chapter. Another option is to move it to an exercise.
- [x] Instead focusing on `background-color` and `color` as the styles of interest
- [x] Implementing descendant selectors instead of class selectors
- [x] Moving discussion of the `display` property to an exercise
- [ ] Adding support for percentage sizes

There are also a lot of good small changes, like a discussion of parser debugging, fixes to `br` elements, an exercise on faster selectors, and a fix for stylesheets that 404. I've also tweaked the book style sheet to work better in our browser.


book/styles.md
```diff
@@ -6,189 +6,124 @@ prev: layout
...
-- Add the block's property values to those elements' `style` fields.
+Once one block can apply to many elements, the possibility exists for
+several blocks apply to a single element. So browsers have a
+*cascading* mechanism to resolve conflicts in favor of the most
+specific rule. Cascading also means browsers can ignore rules they
+don't understand---the cascade will apply the next-most-specific rule
+that it understands.
+
+Let's add support for CSS to our browser. We'll need to parse CSS
+files into selectors, blocks, property-values pairs; to figure out
```
chrishtr, line 103:
```suggestion
files into selectors, blocks, and property-values pairs; to figure out
```

book/styles.md
```diff
@@ -230,32 +165,37 @@ condition is false.[^add-a-comma]
...
 ```
 
 This function takes index `i` pointing to the start of the value and
-returns index `j` pointing to its end. It computes `j` by advancing
-through letters, numbers, and minus and period characters (which might
-be present in numbers), and returns all the text it iterated through
-as the parsed data. Also note the check: if `j` didn't advance, that
-means `i` didn't point at a word to begin with.
+increments it through any word characters,[^word-chars] much like
+`literal`. But unlike `literal`, it also returns the word as parsed
```
chrishtr, line 181: What do you mean by "literal"?

book/styles.md
```diff
@@ -268,51 +208,74 @@ def pair(self, i):
...
+```
+
 Skipping parse errors is a double-edged sword. It hides error
-messages, so debugging CSS files becomes more difficult, and also
+messages, so debugging CSS files becomes more difficult, and it also
 makes it harder to debug your parser.[^try-no-try] This makes
 "catch-all" error handling like this a code smell in most cases.
 
-[^try-no-try]: Try debugging without the `try` block first.
+[^try-no-try]: Try removing the `try` block when debugging.
```
chrishtr, line 278:
```suggestion
[^try-no-try]: I suggest removing the `try` block when debugging.
```

book/styles.md
```diff
@@ -421,33 +400,72 @@ class CSSParser:
...
-both the index and parsed value at the end. You'll get a lot of
-output, but if you step through it by hand, you will find your mistake.
-
-Once we've parsed a CSS file, we need to apply it to the elements on
-the page.
-
-Selecting styled elements
-=========================
+two chapters back. There are a couple of different kinds of errors you
+might run into:
```
chrishtr, line 404: How about moving these debugging tips into a footnote?

book/styles.md
```diff
@@ -469,13 +487,11 @@ def style(node, rules):
...
-CSS (just the elements that wrap the text).
-
-Note that we skip properties that already have a value. That's because
-`style` attributes are loaded into the `style` field first, and should
-take priority. But it means that it matters what order you apply the
-rules in.
+Note that this code skips properties that already have a value. Since
+`style` attributes are filled in first (when the `Element` is
+created), that means `style` attributes take priority over CSS
+stylesheet, which is how it is supposed to work. But it's also a hint
```
chrishtr, line 493:
```suggestion
stylesheets, which is how it is supposed to work. But it's also a hint
```

book/styles.md
```diff
@@ -485,36 +501,46 @@ sort the rules in priority order, with higher-priority rules first.
 
 So let's add a `priority` method to the selector classes that return
 this priority. In this simplest implementation the exact numbers don't
-matter if they sort right, but with an eye toward the future let's
-assign tag selectors priority `1`, class selectors priority `16`, and
-id selectors priority `256`:
+matter if they sort right so let's assign tag selectors priority `1`
```
chrishtr, line 504:
```suggestion
matter, so long as they sort right, so let's assign tag selectors priority `1`
```

book/styles.md
```diff
@@ -564,40 +584,56 @@ their CSS files using the `link` element, which looks like this:
...
+is mandatory. And the `href` attribute describes the CSS file's URL.
 
 [link-types]: https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types
 
-To find these links, we'll need another recursive function:
+[^like-canonical]: For example, `rel=canonical` names the "true name"
+    of a page and is used by search engines to track pages that appear
+    at multiple URLs.
+
+We could definitely write a little recursive function to find ever
```
chrishtr, line 595:
```suggestion
We could definitely write a little recursive function to find every
```

book/styles.md
```diff
@@ -612,85 +648,128 @@ def relative_url(url, current):
...
+usually links are blue. But there's a catch: `color` affects text, but
+text nodes don't have any styles at all. How can that work?
+
+::: {.further}
+Each browser has its own browser style sheet ([Chrome][blink-css],
+[Safari][webkit-css], [Firefox][gecko-css]). [Reset
+stylesheets][mdn-reset] are often used to overcome any differences.
+This works because web page style sheets take precedence over the
+browser style sheet, just like in our browser, though real browsers
+[fiddling with priorities][cascade-order] to make that
```
chrishtr, line 715:
```suggestion
[fiddle with priorities][cascade-order] to make that
```

book/styles.md
```diff
@@ -612,85 +648,128 @@ def relative_url(url, current):
...
+
+Alright: we've got background colors that can be configured by web
+page authors. But there's more to web design than that! At the very
+least, if you're changing background colors you might want to change
+foreground colors as well---the CSS `color` property. For example,
+usually links are blue. But there's a catch: `color` affects text, but
+text nodes don't have any styles at all. How can that work?
+
+::: {.further}
+Each browser has its own browser style sheet ([Chrome][blink-css],
```
chrishtr, line 710:
```suggestion
Each browser has its own browser style sheet ([Chromium][blink-css],
```

book/styles.md
```diff
@@ -612,85 +648,128 @@ def relative_url(url, current):
...
+[Safari][webkit-css], [Firefox][gecko-css]). [Reset
```
chrishtr, line 711:
```suggestion
[WebKit][webkit-css], [Gecko][gecko-css]). [Reset
```

book/styles.md
```diff
@@ -724,104 +806,137 @@ def style(node, rules):
...
+    # ...
+    for x, y, word, font, color in self.display_list:
+        display_list.append(DrawText(x, y, word, font, color))
 ```
 
-And then read that in `has_block_children`:[^default-block]
+Phew! That was a lot of coordinated changes, so please test everything
+and make sure it works. You should now see links on this page appear
+in blue---and you'll also notice that the rest of the text has become
+slightly lighter.^[Check out [the book's stylesheet](book.css) to see
```
chrishtr, line 882:
```suggestion
slightly lighter.^[Check out the book's stylesheet](book.css) to see
```

book/styles.md
```diff
@@ -724,104 +806,137 @@ def style(node, rules):
...
+===============
 
-``` {.python}
-def has_block_children(self):
-    for child in self.node.children:
-        # ...
-        elif child.style.get("display", "inline") == "inline":
-            return False
-    return True
+Well---since we're already mucking around with `InlineLayout`, let's
```
chrishtr, line 888:
```suggestion
Since we're already mucking around with `InlineLayout`, let's
```

book/styles.md
```diff
@@ -724,104 +806,137 @@ def style(node, rules):
...
+also modify it to look up the font size, weight, and style information
+via CSS instead of using the `style`, `weigth`, and `size` fields on
+`InlineLayout`:
+
+::: {.todo}
+How do we resolve percentage values for nodes? And the `font-size`
+needs to be converted from pixels to points.[^72ppi]
+:::
+
+[^72ppi]: Normally you think of points as a physical length unit (one
```
chrishtr, line 898: This differs between Windows and Mac.

book/styles.md
```diff
@@ -724,104 +806,137 @@ def style(node, rules):
...
+    size = node.style["font-size"] * .75 # ???
+    font = tkinter.font.Font(size=style, weight=weight, slant=style)
+    # ...
 ```
 
-With these changes, `InlineLayout` can lose its `open` and `close`
-methods, becoming a small, self-contained engine for line layout while
-most of its domain-specific knowledge of tags is moved to the browser
-style sheet.
+Note that for `font-style` we needs to translate CSS "normal" to Tk
```
chrishtr, line 914:
```suggestion
Note that for `font-style` we need to translate CSS "normal" to Tk
```

book/styles.md
```diff
@@ -881,17 +1017,11 @@ priorities.[^lexicographic]
...
-guess why? Hint: try to implement ancestor selectors and analyze the speed of
-your algorithm.
+*Ancestor Selectors*: an ancestor selector is the inverse of a
+descendant selector - it styles an ancestor according to the presence
+of a descendant. This feature is one of the benefits provided by the
+[`:has` syntax](https://drafts.csswg.org/selectors-4/#relational). Try
+to implement ancestor selectors. As I write this, no browser has
+actually implemented `:has`; why? Hint: analyze the asymptotic speed
+of your implementation. There is a clever implementation that is
+*O(1)* amortized per element---can you find it?
```
chrishtr, line 1027: What is the solution to this you have in mind?

book/styles.md
```diff
@@ -881,17 +1017,11 @@ priorities.[^lexicographic]
...
-[`:has` syntax](https://drafts.csswg.org/selectors-4/#relational). However, you
-will find that `:has` is not implemented in any real browser as yet. Can you
-guess why? Hint: try to implement ancestor selectors and analyze the speed of
-your algorithm.
+*Ancestor Selectors*: an ancestor selector is the inverse of a
+descendant selector - it styles an ancestor according to the presence
+of a descendant. This feature is one of the benefits provided by the
+[`:has` syntax](https://drafts.csswg.org/selectors-4/#relational). Try
+to implement ancestor selectors. As I write this, no browser has
+actually implemented `:has`; why? Hint: analyze the asymptotic speed
```
chrishtr, line 1025:
```suggestion
actually implemented `:has`; why do you think that is? Hint: analyze the asymptotic speed
```

book/styles.md
```diff
@@ -838,33 +953,54 @@ files. That means we:
...
+=======
+
+The complete set of functions, classes, and methods in our browser 
+should look something like this:
+
+::: {.cmd .python .outline html=True}
+    python3 outlines.py --html src/lab6.py
+:::
+
 Exercises
```
chrishtr, line 966: How about !important as an exercise?

book/styles.md
```diff
@@ -6,189 +6,124 @@ prev: layout
...
-``` {.python}
-def px(s):
-    if s.endswith("px"):
-        return int(s[:-2])
-    else:
-        return 0
+Here we're adding the `style` field in the `ElementNode` constructor,
+based on the `style` attribute,[^python-get] and filling it by parsing
+that attribute's value. Now we can use that information when we do
+layout:
```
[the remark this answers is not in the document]
pavpanchekha: Let's discuss common rendering terms soon so this is easier (paint vs draw vs render).

book/styles.md
pavpanchekha: I'm guessing there's some convention to this. I avoided saying Blink, Webkit, Gecko here because those terms aren't as widely known, even though it's more accurate. Is that better? Is there some legal guideline on using "Chrome"?

book/styles.md
pavpanchekha: I thought the difference here would be handled by Tk. We should test.

book/styles.md
pavpanchekha: Suppose you have a selector, `S1:has(S2)`, where we assume `S2` doesn't have a `:has` inside. Then as a pre-pass, match each element against `S2` and store that as a bit on the element. As a second pass over all the elements, store a second bit for whether any descendant matches `S2`, computed by propagating the first bit upwards. That second bit now represents `:has(S2)` for all elements, so you can match `S1:has(S2)` using it.

That doesn't work for some reason? Perhaps I should have thought more about it.

book/styles.md
pavpanchekha: Great idea!

book/styles.md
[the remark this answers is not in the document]
chrishtr: cf glossary.md (which we haven't yet added to the published book)

book/styles.md
chrishtr: Yes it does work.

This is a good learning moment for the student, as to why even this is not good enough. I suggest adding a sentence saying that in real browsers, even O(1) amortized time is not good enough - iterating twice over the tree is too much overhead.

Literally today we are discussing this:

https://github.com/mozilla/standards-positions/issues/528#issuecomment-842148898

book/styles.md
chrishtr: Not a legal guideline. I don't want to say Chrome because it implies it's the feature is not in Edge, Brave, Opera, etc.

I'd be happy with just changing Chrome to Chromium.

book/styles.md
chrishtr: This is a blog post that goes into what I'm talking about:

https://tonsky.me/blog/font-size/


Read after "why the number 72?

chrishtr:
How is this PR going? Can we get to a point where it is landed so we can iterate on top of it?

pavpanchekha:
I can merge it. The reason I haven't—and the biggest blocker for this chapter—is that I haven't figured out a good way to do percentage font sizes. I might throw in the towel and go back to normal pixel sizes.

chrishtr:
I see. Just merge it and I'll also give it a read and also think through the percentage font size issue. Maybe one of us can figure something out.

pavpanchekha:
You're right. I'll do the merge, one second.

pavpanchekha:
Ok pushed!

PR #122, opened 2021-05-17
Add lint target that runs compare.py

chrishtr:
This makes a non-backward-compatible change to compare.py, so we'll need to update the github scripts.


chrishtr:
Thanks for fixing. I forgot that it was under the .github subdirectory. I grep -r * 'ed for it but of course * doesn't match hidden files.

PR #123, opened 2021-05-18
Fix lint errors in some chapters; update lint to cover all of them.

chrishtr:


book/reflow.md
```diff
@@ -723,7 +722,7 @@ Next up: the `BlockLayout`. Here `compute_height` looks like this:
 ``` {.python}
 class BlockLayout:
     def compute_height(self):
-        self.h = self.pt + self.pb
+        self.h = 0
```
chrishtr, line 725: Is this correct, or was the book correct and the lab wrong?

pavpanchekha: I think the height should include the padding here. It's a little weird that we add the top and bottom padding at the start. So the original is, I think, correct.

pavpanchekha:
Right now those chapters don't pass linting, but the code is also off anyway, since as we edit and publish chapters we're changing things and not propagating those changes to later chapters. Happy to merge this. As we edit and publish these chapters we'll need to redo the code a bit anyway; when we do that we have to remember to add back the `.python` and remove any `expected=False`, at least if that's not intentional.

PR #124, opened 2021-05-18
Rename render() to draw() throughout

chrishtr:


pavpanchekha (approved):
So I think the summary here is that layout objects `paint` themselves, and the browser then `draw`s them? Sounds good.

PR #126, opened 2021-05-22
Revisions to the rendering architecture chapter

chrishtr:
Changes:

* Introduces the Task and TaskQueue classes, and uses them in the code and chapter
* Removes unnnecessary aside about the networking thread
* Reworks the intro to focus first on the task queues and event loops.
* More typo and editorial fixes to the rest of the chapter.


pavpanchekha:
This is looking better and better and better. We should merge it.

I think focusing on the code is probably the best next step here. Do you think it's possible to move both 1) the locking, and 2) the `needs_XXX` flags, into the TaskList class? That would keep the code more self-contained, and could also let the text build the TaskList step by step instead of dropping the whole thing at the start.

pavpanchekha (approved):

chrishtr:
> I think focusing on the code is probably the best next step here. Do you think it's possible to move both 1) the locking, and 2) the `needs_XXX` flags, into the TaskList class? That would keep the code more self-contained, and could also let the text build the TaskList step by step instead of dropping the whole thing at the start.

Good thoughts, I'll try that in the next PR.

PR #128, opened 2021-05-23
Add feedback button and dialog

chrishtr:
This PR adds a new modal dialog for providing per-chapter feedback. It also adds some upsell language advertising Ctrl-E.

### DIFF
--- a/infra/api.py
+++ b/infra/api.py
@@ -183,7 +183,6 @@ def static(file):
 
 @bottle.route('/')
 def index():
-    print('/')
     return bottle.static_file("index.html", root=".")
 
 @bottle.route("/auth/tools")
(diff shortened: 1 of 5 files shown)


pavpanchekha (approved):

pavpanchekha:
This took me a while to review, because when I started reviewing I had a bunch of improvements I wanted to suggest (for example—if someone has already activated the typo tools, don't ask them their name again, don't show the legalese again). But as I thought about it more—most of these improvements aren't about adding chapter comments, but a broader overhaul of the typo & feedback system.

So here's my take: let's merge this now, and rethink the typo & feedback system separately.

chrishtr:
Ok merging now. Let's be careful about rolling it out though and do some testing on the live site. Also back up the existing db in case.

PR #129, opened 2021-05-26
Some edits to the style chapter

chrishtr:


src/lab6.py
```diff
@@ -322,10 +322,10 @@ def style(node, rules):
     if isinstance(node, TextNode):
         node.style = node.parent.style
     else:
+        node.style = {}
```
chrishtr, line 325: I think there was previously a bug, in the situation where there is a style recalc and it causes properties
to no longer apply. This fixes that bug I think..

pavpanchekha: Ooh, style recalc is actually a problem, because I parse and use the `style` attribute in the `Element` constructor. I think I need to move that out—then this fix will work.

pavpanchekha (approved):

PR #130, opened 2021-05-27
Another pass over the `style` chapter

pavpanchekha:
This chapter is getting better and cleaner! This PR:

- [x] Moves attribute parsing to the `style` function
- [x] Actually implements font sizes

### DIFF
--- a/book/styles.md
+++ b/book/styles.md
@@ -869,7 +869,9 @@ in the `style` function:
 ``` {.python}
 def style(node, rules):
     # ...
-    if node.style["font-size"].endswith("px"):
+    if isinstance(node, Text):
+        node.font_size = node.parent.font_size
+    elif node.style["font-size"].endswith("px"):
         node.font_size = float(node.style["font-size"][:-2])
     elif node.style["font-size"].endswith("%"):
         parent_font_size = node.parent.font_size if node.parent else 16
--- a/book/styles.md
+++ b/book/styles.md
@@ -520,19 +520,18 @@ Downloading styles
 Actually... let's back up. Yes, browsers download web-page-specific
 CSS files. But each browser also ships with a *browser style
 sheet*,[^technically-ua] which defines the default styles for all
-sorts of elements. Let's work on the browser style sheet first.
+sorts of elements. Let's work on the browser style sheet first. It
+might look like this:
 
 [^technically-ua]: Technically called a "User Agent" style sheet. User Agent,
     [like the Memex](history.md).
-
-Our browser's style sheet might look like this:
 
 ``` {.css}
 pre { background-color: gray; }
 ```
 
 Let's store that in a new file, `browser.css`, and have our browser
-load and parse that file when it downloads a page:[^not-correct-browser-styles]
+load and parse that file when it downloads a page:
 
 ``` {.python replace=browser.css/browser6.css}
 class Browser:
@@ -544,29 +543,27 @@ class Browser:
         # ...
 ```
 
-Beyond the browser styles, our browser needs to find website-specific
-CSS files, download them, and use them as well. Web pages name their
-CSS files using the `link` element, which looks like this:
+Our browser needs to find website-specific CSS files, download them,
+and use them as well. Web pages name their CSS files using the `link`
+element, which looks like this:
 
 ``` {.example}
 <link rel="stylesheet" href="/main.css">
 ```
 
-The `rel` attribute here tells that browser that this is a link to a
-style sheet. Browsers mostly don't care about any [other kinds of
-links][link-types], but search engines do,[^like-canonical] so `rel`
-is mandatory. And the `href` attribute describes the CSS file's URL.
+The mandatory `rel` attribute identifies this link as a style
+sheet[^like-canonical] and the `href` attribute has the style sheet
+URL. So we need to fine all links like this. We could do that with a
+little recursive function, but since we'll be doing similar tasks in
+the next few chapters, so let's instead write a more general recursive
+function that turns a tree into a list of nodes:
+
+[^like-canonical]: Browsers mostly don't care about any [other kinds
+of links][link-types], but search engines use relations like
+`rel=canonical`, which names the "true name" of a page and lets search
+engines track pages that appear at multiple URLs.
 
 [link-types]: https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types
-
-[^like-canonical]: For example, `rel=canonical` names the "true name"
-    of a page and is used by search engines to track pages that appear
-    at multiple URLs.
-
-We could definitely write a little recursive function to find every
-`link` element with those exact attributes. But we'll be doing similar
-tasks a lot in the next few chapters, so let's instead write a more
-general recursive function to turn a tree into a list:
 
 ``` {.python}
 def tree_to_list(tree, list):
@@ -576,8 +573,9 @@ def tree_to_list(tree, list):
     return list
 ```
 
-We can later use this function on the layout tree as well. Anyway, 
-now we can grab the URLs for each CSS file with this line:
+I've written this to work on both HTML and layout tree, for later.
+Anyway, now we can use a Python's list expression to grab the URL of
+each linked style sheet:
 
 ``` {.python indent=4}
 def load(self, url):
@@ -591,13 +589,9 @@ def load(self, url):
 ```
 
 It's kind of crazy, honestly, that Python lets you do this. Anyway,
-now we have all the style sheet URLs. The browser is expected to make a
-GET request to that location, parse the style sheet, and use it to
-style the page.
-
-Note that style sheet URLs we have are not full URLs; they are
-something called *relative URLs*, which can come in three
-flavors:^[There are even more flavors, including query-relative and
+the browser has the style sheet URLs now. But these style sheet URLs
+are usually not full URLs; they are something called *relative URLs*,
+such as:^[There are even more flavors, including query-relative and
 scheme-relative URLs, that I'm skipping.]
 
 -   A normal URL, which specifies a scheme, host, path, and so on;
@@ -611,8 +605,7 @@ scheme-relative URLs, that I'm skipping.]
     "directories" are dropped from the current URL; and then the
     relative URL is put at the end.
 
-To turn a relative URL into a full URL, then, we need to figure out
-which case we're in:
+So we need a function that turns a relative URL into a full URL:
 
 ``` {.python}
 def relative_url(url, current):
@@ -630,10 +623,8 @@ def relative_url(url, current):
         return dir + "/" + url
 ```
 
-Let's put it all together. We want to collect CSS rules from each of
-the linked files, and the browser style sheet, into one big list so we
-can apply each of them. So let's add them onto the end of the `rules`
-list:
+Now the browser can request each linked style sheet and add them onto
+the `rules` list:
 
 ``` {.python indent=4}
 def load(self, url):
@@ -641,23 +632,23 @@ def load(self, url):
     for link in links:
         try:
             header, body = request(relative_url(link, url))
-            rules.extend(CSSParser(body).parse())
         except:
             continue
+        rules.extend(CSSParser(body).parse())
 ```
 
-The `try`/`except` here handles the case where downloading a style
-sheet fails---in which case the browser just ignores it. But if you're
+Note that we add the page's style sheets at the end of `rules`, which
+means that user styles take priority over the browser style sheet. The
+`try`/`except` here handles the case where downloading a style sheet
+fails, in which case the browser just ignores it. But if you're
 debugging `relative_url` try removing the `try`/`except` here, which
 can hide errors.
 
-Since the page's style sheets come *after* browser styles, user styles
-take priority over the browser style sheet. With the rules loaded, we
-need only sort and apply them and then do layout, the code for which
-we've already added to `load`. Open this page up again, and you should
-see both gray backgrounds on every code block (thanks to the browser
-style sheet) and light-gold backgrounds on this book's mailing list
-signup form (try the book's main page).
+With the rules loaded, we need only sort and apply them and then do
+layout, the code for which we've already added to `load`. Open this
+page up again, and you should see gray backgrounds on every code block
+(thanks to the browser style sheet) and light-gold backgrounds on this
+book's mailing list signup form (try the book's main page).
 
 Alright: we've got background colors that can be configured by web
 page authors. But there's more to web design than that! At the very
@@ -729,11 +720,11 @@ def style(node, rules):
 Inheriting font size comes with a twist. Web pages can use percentages
 as font sizes: `h1 { font-size: 150% }` makes headings 50% bigger than
 normal text. But what if you had, say, a `code` element inside an `h1`
-tag---would that inherit the `150%` value for `font-size`? Would it be
-another 50% bigger than the rest of the heading text? That seems
-wrong. So in fact, browsers resolve percentages to absolute pixel
-units before storing them in the `style`; it's called
-["computing"][^css-computed] the style.
+tag---would that inherit the `150%` value for `font-size`? Surely it
+shouldn't be another 50% bigger than the rest of the heading text? So,
+in fact, browsers resolve percentages to absolute pixel units before
+storing them in the `style`; it's called ["computing"][^css-computed]
+the style.
 
 [^css-computed]: Full CSS is a bit more confusing: there are
 [specified, computed, used, and actual values][css-computed], and they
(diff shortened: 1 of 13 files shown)


pavpanchekha:
Ok, I implemented font sizes, including percentages. I'm not _thrilled_ with it—the chapter is now longer than the first chapter again, and also it still incorrectly makes font sizes not inheritable in order to avoid implementing computed styles.

pavpanchekha:
Ok, I did another pass on font sizes, this time implementing a minimal "computed style" mechanism. It's still not idea (even more text in a long chapter) but at least the implementation is minimally faithful to reality and works well.

book/styles.md
```diff
@@ -5,55 +5,65 @@ prev: layout
...
-does not support that.] The browser looks at those property-value
-pairs when drawing elements, so this `style` attribute allows web page
-authors to override the default background for `pre` elements.
+This `<div>` element's `style` attribute contains a property/value
+pair matching the property `background-color` to the value
+`lightblue`.^[CSS allows spaces around the punctuation, but our
+simplistic attribute parser does not support that.] Multiple
+property/value pairs, separated by semicolons, are also allowed. The
+browser looks at those property-value pairs when painting the element;
+this way the `style` attribute allows web page authors to override the
```
chrishtr, line 34:
```suggestion
this allows web page authors to override the
```

book/styles.md
```diff
@@ -68,45 +78,50 @@ class BlockLayout:
         # ...
 ```
 
-Put the exact same lines of code inside `InlineLayout` as well, so that
-paragraphs and list items and so on can have backgrounds as well. For now, also
-remove the default gray background from `pre` elements, but we'll put it back
-soon.
+Put the exact same lines of code inside `InlineLayout` as well, so
```
chrishtr, line 81: Your code above uses a static method and brand checking on class names instead...

book/styles.md
```diff
@@ -68,45 +78,50 @@ class BlockLayout:
...
+    across many devices (try this book on mobile!).
 
 ``` {.css}
 selector { property-1: value-1; property-2: value-2; }
 ```
 
-Once one block can apply to many elements, the possibility exists for
-several blocks apply to a single element. So browsers have a
+Since one of these *rules* can apply to many elements, it's possible
+for several blocks to apply to the same element. So browsers have a
```
chrishtr, line 120:
```suggestion
for several rules to apply to the same element. So browsers have a
```

book/styles.md
```diff
@@ -218,36 +231,35 @@ def pair(self):
...
         pairs[prop] = val
         self.whitespace()
         self.literal(";")
         self.whitespace()
-    self.literal("}")
     return pairs
 ```
 
-Another twist to parsing functions is handling errors. So for example,
+One twist to parsing functions is handling errors. So for example,
```
chrishtr, line 254:
```suggestion
One twist to parsing functions is handling errors. For example,
`

File: book/styles.md [hunk 5]
```diff
@@ -434,42 +449,28 @@ class DescendantSelector:
...
-It also means that it matters what order you apply the rules in.
-What's the correct order? In CSS, it's called *cascade order*, and it
-is based on the selector used by the rule. Tag selectors get the
-lowest priority; class selectors one higher; and id selectors higher
-still. But since our CSS parser just has tag selectors, you just count
-the number of tag selectors to put the rules in order:
+Since two rules can apply to the same element, rule order matters. In
+CSS, the correct order is called *cascade order*, and it is based on
+the rule's selector. Tag selectors get the lowest priority; class
+selectors one higher; and id selectors higher still. Since we only
```
chrishtr, line 472: Also note here that order of declaration is a tie breaker.

book/styles.md
```diff
@@ -483,50 +484,47 @@ class DescendantSelector:
...
 
     self.document = DocumentLayout(nodes)
     # ...
 ```
 
-In Python, the `sorted` function is *stable*, which means that things
-keep their relative order if possible. This means that in general, a
-rule that comes later in the CSS file has higher priority, unless the
+Note that Python's `sorted` function keeps the relative order of
+things when possible. This means that in general, a rule that comes
```
chrishtr, line 511:
```suggestion
things when possible. This ensures that a rule that comes
```

book/styles.md
```diff
@@ -483,50 +484,47 @@ class DescendantSelector:
...
+later in the CSS file comes later in cascade order, unless the
 selectors used force something different. That's how real browsers do
```
chrishtr, line 513:
```suggestion
selectors used force something different.
```

book/styles.md
```diff
@@ -483,50 +484,47 @@ class DescendantSelector:
...
-it, too. The `reversed` call is because we want higher-priority rules
-to come first.
+it, too.
```
chrishtr, line 514:
```suggestion
```

book/styles.md
```diff
@@ -701,109 +699,112 @@ color and other font properties are.
...
+    if property == "font-size":
         # ...
+    else:
+        return value
 ```
 
-With all this in place, we can implement text color itself. First,
-let's add a `color` parameter to `DrawText` and pass it to
-`create_text`'s `fill` parameter:
+Compute `font-size` values works differently for pixel and percentages
```
chrishtr, line 757:
```suggestion
Computing `font-size` values works differently for pixel and percentage
```

book/styles.md
```diff
@@ -701,109 +699,112 @@ color and other font properties are.
...
-        # ...
-        self.color = color
+``` {.python indent=8}
+if value.endswith("px"):
+    return value
+elif value.endswith("%"):
+    if node.parent:
+        parent_px = float(node.parent.style["font-size"][:-2])
+    else:
+        parent_px = 16
```
chrishtr, line 767: factor 16 into a global static constant

src/lab6.py
```diff
@@ -274,7 +266,10 @@ def parse(self):
         while self.i < len(self.s):
             try:
                 selector = self.selector()
+                self.literal("{")
```
chrishtr, line 269: Was this a bug previously?

pavpanchekha: No, this is kind of subtle. Before, the `block` production included the curly braces, now it's in the overall `parse` construction. `block` now just has the property/value pairs. I want to see if I can use that to parse style attributes and thereby shorten the chapter a bit and also improve the code.

pavpanchekha:
I still have a few things I want to try in the style chapter to shorten and improve it:

- Perhaps use the CSSParser to parse style attributes. This means the chapter ramps up slowly (bad) but also splits the long parsing section into two (good). It also leads to better code and could be a bit shorter.
- Perhaps move the browser stylesheet up into the applying styles section, which will make that section more motivated and could also shorten stuff.

pavpanchekha:
Another thing to perhaps change: parse and save the browser stylesheet on start instead of every time a page is loaded.

pavpanchekha:
I finished a nit-picky pass through the whole document where I tried to clean up the writing. The chapter is not quite so long, but the parsing section is still tedious. I'm going to now try those three ideas I posted in earlier comments—hopefully that makes it a little less boring to read, and if it shortens the chapter a bit more I won't be too worried about the length any more.

book/styles.md
pavpanchekha: Note sure what you mean here.

pavpanchekha:
I re-requested a review because I did a pretty big re-organization of the chapter. Not much of the content changed, but it got shifted around quite a bit. The end result is that the chapter is quite a bit shorter (just a bit over 4k words, though it's still the chapter with the most code thanks to the parser) and I think a bit clearer and less tedious as well.

I think it definitely needs a fresh read through by both of us just to make sure nothing important got lost.

pavpanchekha:
Ok, I did another pass over the chapter. This time all the changes were pretty minor. I also tested stuff a bit to shake out some bugs. It's still missing two "Go Further" blocks but otherwise is pretty much ready to ship.

book/styles.md
```diff
@@ -689,112 +653,232 @@ happen.[^ours-works]
...
+A web page can now have any number of stylesheets applied to it. And
+since two rules can apply to the same element, rule order matters: it
+determines which rules take priority, and when one rule overrides
+another.
+
+In CSS, the correct order is called *cascade order*, and it is based
+on the rule's selector, with file order as a tie breaker. This system
+allows more specific rules to override more general ones, so that you
+can have a browser style sheet, a site-wide style sheet, and maybe a
+special style sheet for your about page, all co-existing.
```
chrishtr, line 668:
```suggestion
special style sheet for your specific web page, all co-existing.
```

book/styles.md
```diff
@@ -689,112 +653,232 @@ happen.[^ours-works]
...
+    style(nodes, sorted(rules, key=cascade_priority))
+    # ...
+```
+
+Note that before sorting `rules`, it as in file order. Since Python's
+`sorted` function keeps the relative order of things when possible,
+file order thus acts as a tie breaker, as it should.
+
+That's it: we've added CSS to our web browser! I mean---for background
+colors. But there's more to web design than that! At the very least,
```
chrishtr, line 707:
```suggestion
colors. But there's more to web design than that. At the very least,
```

book/styles.md
```diff
@@ -839,75 +912,55 @@ def paint(self, display_list):
...
+            # ...
+            fill=self.color,
+        )
+```
 
-[why-72]: https://tonsky.me/blog/font-size/
-[why-7227]: https://tex.stackexchange.com/questions/200934/why-does-a-tex-point-differ-from-a-desktop-publishing-point
+Phew! That was a lot of coordinated changes, so test everything and
+make sure it works. You should now see links on this page appear in
+blue---and you'll might also notice that the rest of the text has
```
chrishtr, line 933:
```suggestion
blue---and you might also notice that the rest of the text has
```

book/styles.md
```diff
@@ -839,75 +912,55 @@ def paint(self, display_list):
...
-    # ...
-    weight = node.style["font-weight"]
-    style = node.style["font-style"]
-    if style == "normal": style = "roman"
-    size = node.style["font-size"] * .75 # ???
-    font = tkinter.font.Font(size=style, weight=weight, slant=style)
-    # ...
-```
+[^book-css]: The book's main body text [is colored](book.css) `#333`,
+    or roughly 97% black.
```
chrishtr, line 937: It is? 3/15 = 20%, no?

pavpanchekha: Yeah, but then it's gamma corrected. I actually busted out the sRGB converter and checked. Perhaps I should say something like "97% black after gamma correction" and link "gamma correction" to the Wikipedia page?

chrishtr: SGTM

book/styles.md
chrishtr: Looks like my comment is outdated anyway, so disregard.

book/styles.md
```diff
@@ -404,172 +465,103 @@ function and a close parenthesis at the end, you can use your editor's
...
+A parser receives arbitrary bytes as input, so parser bugs are usually
+easy for bad actors to exploit. Parser correctness is thus crucial to
+browser security, as [many][bug-1] [parser][bug-2] [bugs][bug-3] have
+demonstrated. Nowadays browser developers use [fuzzing] to try to find
+and fix such bugs.
+:::
 
-Since descendant selectors wrap other selectors, their `match` method
-is recursive:
+[bug-1]: https://nvd.nist.gov/vuln/detail/CVE-2010-3971
```
chrishtr, line 479: Curious: are these particularly well-known vulnerabilities?

book/styles.md
```diff
@@ -404,172 +465,103 @@ function and a close parenthesis at the end, you can use your editor's
...
+Applying style sheets
+=====================
 
-If a rule applies to an element, the browser needs to add its
-property-value pairs to the element's `style`. The logic is pretty
-simple. First, we need to recurse over all the HTML elements:
+With the parser debugged, the next step is applying the parsed style
+sheet to the web page. Since each CSS rule can style many elements on
+the page, this will require looping over all elements *and* all rules.
+When a rule applies its property/values pairs are copied to the
```
chrishtr, line 490:
```suggestion
When a rule applies, its property/values pairs are copied to the
```

book/styles.md
```diff
@@ -404,172 +465,103 @@ function and a close parenthesis at the end, you can use your editor's
...
+``` {.python indent=8 expected=False}
+def load(self, url):
+    # ...
+    rules = self.default_style_sheet.copy()
+    style(nodes, rules)
+    # ...
+```
+
+The browser style sheet is the default for the whole web. But each web
+sites can also use CSS to set a consistent style for the whole site.
```
chrishtr, line 540:
```suggestion
site can also use CSS to set a consistent style for the whole site.
```

book/styles.md
```diff
@@ -624,50 +610,37 @@ def relative_url(url, current):
...
-signup form (try the book's main page).
-
-Alright: we've got background colors that can be configured by web
-page authors. But there's more to web design than that! At the very
-least, if you're changing background colors you might want to change
-foreground colors as well---the CSS `color` property. For example,
-usually links are blue. But there's a catch: `color` affects text, but
-text nodes don't have any styles at all. How can that work?
+The `try`/`except` ignores style sheets that fail to download, but it
+can also hide bugs in your, so if something's not right try removing
```
chrishtr, line 642:
```suggestion
can also hide bugs in your code, so if something's not right try removing
```

book/styles.md
```diff
@@ -689,112 +662,232 @@ happen.[^ours-works]
...
-        canvas.create_text(
-            # ...
-            color=self.color,
-        )
+The loop that handles `style` attributes should likewise call
+`computed_style`. Note that because the `style` function recurses at
+the end of the function, the node's parent already has a `font-size`
+value stored when `compute_style` is called.
+
+::: {.further}
```
chrishtr, line 842: remove - empty

pavpanchekha: I'm going to fill the two empty ones in with something interesting. Open to suggestions—thinking of talking about rule trees in the "Inheritance" section and promoting the long footnote on pixels v points to the "Font Properties" go further block.

book/styles.md
pavpanchekha: I searched the CVE database—perhaps more recent bugs also exist but don't get CVEs for some reason? It would be nice to link to more recent and more prominent bugs when possible.

book/styles.md
chrishtr: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/css/style-calculation.md doesn't have much..

pavpanchekha:
The "Go Further" blocks have been added. Besides linking better parsing bugs (actual news stories would be ideal!) I don't have any "todos" before publishing this chapter. Please read through it again and we can release it!

book/styles.md
```diff
@@ -283,21 +162,26 @@ def body(self):
...
-On the web, however, there is an unusual benefit: it supports an
-ecosystem of multiple implementations. For example, different browsers
-support different kinds of property values,[^like-parens] so CSS that
-parses in one browser might not parse in another. With silent parse
-errors, browsers just ignore stuff they don't understand. The
-principle (variously called "Postel's Law",[^for-jon] the "Digital
-Principle",[^from-circuits] or the "Robustness Principle") is: produce
-maximally conformant output but accept even minimally conformant
-input.
+But "catch-all" error handling has an unusual benefit on the web . The
```
chrishtr, line 172:
```suggestion
But "catch-all" error handling has an unusual benefit on the web. The
```

book/styles.md
```diff
@@ -309,92 +193,269 @@ input.
...
+            if self.ancestor.matches(node.parent): return True
+            node = node.parent
+        return False
+```
+
+Now, to create these selector objects, we need a parser. In this case,
+that's just another parsing function:[^not-quite-word]
+
+[^not-quite-word]: Once again, using `word` here for tag names is
+actually not quite right, but it is close enough. One tricky side
```
chrishtr, line 366:
```suggestion
actually not quite right, but it's close enough. One tricky side
```

book/styles.md
```diff
@@ -404,172 +465,103 @@ function and a close parenthesis at the end, you can use your editor's
...
+    style(nodes, rules)
+    # ...
 ```
 
-Beyond the browser styles, our browser needs to find website-specific
-CSS files, download them, and use them as well. Web pages name their
-CSS files using the `link` element, which looks like this:
+The browser style sheet is the default for the whole web. But each web
+site can also use CSS to set a consistent style for the whole site.
+To do that, the web site posts a CSS file and references it from each
```
chrishtr, line 541:
```suggestion
To do that, the web site references CSS files and references them from each
```

book/styles.md
```diff
@@ -404,172 +465,103 @@ function and a close parenthesis at the end, you can use your editor's
...
+page using a `link` element:
```
chrishtr, line 542:
```suggestion
page using `link` elements:
```

book/styles.md
```diff
@@ -689,112 +662,231 @@ happen.[^ours-works]
...
-already inherited the correct property value when the children try to
-read it.
+Inheriting font size comes with a twist. Web pages can use percentages
+as font sizes: `h1 { font-size: 150% }` makes headings 50% bigger than
+surrounding text. But what if you had, say, a `code` element inside an
+`h1` tag---would that inherit the `150%` value for `font-size`? Surely
+it shouldn't be another 50% bigger than the rest of the heading text?
+
+So, in fact, browsers resolve percentages to absolute pixel units
+before storing them in the `style`; it's called a ["computed
```
chrishtr, line 780:
```suggestion
before storing them in the `style` and inheriting them to descendants; it's called a ["computed
```

book/styles.md
```diff
@@ -689,112 +662,231 @@ happen.[^ours-works]
...
+            return None
     else:
-        # ...
+        return value
 ```
 
-With all this in place, we can implement text color itself. First,
-let's add a `color` parameter to `DrawText` and pass it to
-`create_text`'s `fill` parameter:
+The case for percentage sizes has to handle a tricky edge case:
```
chrishtr, line 804:
```suggestion
Percentage font sizing also has to handle a tricky edge case:
```

PR #134, opened 2021-06-06
Add lab1 unittests

chrishtr:


pavpanchekha:
I'm open to the approach, but we're going to need to make edits to Chapter 1 to describe `get_text`. I'm also worried about our testing requiring too much modification to the code. To the extent there's a trade-off between readability and testability, we should throw in with readability. In this case we can avoid any code changes by changing `sys.stdout` to a `io.StringIO()` and using its `getvalue()` method to capture the output.

Do you mind if I do a pass on this unit testing code before we merge it?

chrishtr:
> Do you mind if I do a pass on this unit testing code before we merge it?

Sure, go for it. You make some good points.

chrishtr:
Updated to mock sys.stdout and reverted changes to non-test code. Please take a look.

PR #141, opened 2021-06-10
JS runnable browser

pavpanchekha:
This PR adds support for running the full browser (for chapters 2-8) in the browser, with demo browsers for chapters 2, 3, and 4. Besides some bug fixes, the main changes are:

1. Support for `bind` in the runtime. All the keys used in the chapters to date are implemented, and extending this shouldn't be too painful. This also required special-casing `bind` in the compiler to deal with "with".
2. Support for using `fetch` to fetch real web pages. The main change here is special-casing `makefile` to be an async function, so that it can make async requests. Naturally the web page needs to allow CORS to be fetchable.


www/widgets/rt.js
```diff
@@ -99,7 +113,33 @@ static tkinter(options) {
         }
 
         bind(key, fn) {
-            // this.tk.addEventListener(...)
+            if (['<Up>', '<Down>', '<Left>', '<Right>'].indexOf(key) !== -1) {
```
chrishtr, line 116: Is typing in the URL bar supposed to work? I tried building your branch just now and it didn't seem to.

www/widgets/lab2-browser.html
```diff
@@ -0,0 +1,36 @@
+<!doctype html>
```
chrishtr, line 1: These files are all the same - use a pandoc template?

chrishtr, line 1: These files are all the same - use a pandoc template?

infra/compile.py
```diff
@@ -191,7 +191,13 @@ def load_outline(ol):
 def compile_method(base, name, args, ctx):
     base_js = compile_expr(base, ctx)
     args_js = [compile_expr(arg, ctx) for arg in args]
-    if name in LIBRARY_METHODS:
+    if name == "bind": # Needs special handling due to "this"
+        assert len(args) == 2
+        return base_js + ".bind(" + args_js[0] + ", (e) => " + args_js[1] + "(e))"
+    elif name == "makefile":
+        assert len(args) == 3
+        return "await " + base_js + ".makefile(" + ", ".join(args_js) + ")"
```
chrishtr, line 199: clever!

www/widgets/rt.js
pavpanchekha: It doesn't yet. I would like to add support, but I need to make a UI to explain that the web version is limited to same-origin.

www/widgets/lab2-browser.html
pavpanchekha: That's a good idea!

www/widgets/rt.js
pavpanchekha: URL is now configurable, but has to stay on the "browser.engineering" domain. Avoids having to explain this stuff.

www/widgets/lab2-browser.html
pavpanchekha: It templatized it, but we don't have an easy way to link chapter numbers (like `4`) to chapter names (like `html`) so all the templates right now point to the "graphics" chapter. Not ideal, I'll think about ways to fix it.

pavpanchekha:
I believe this is ready to merge

chrishtr (approved):
LGTM % my comments from the other PR

PR #144, opened 2021-06-20
Convert lab2 to use JS modules

chrishtr:
The primary reason for this change is to make it easier for the labs to share code. In a subsequent PR, for example,
I plan to import the `request` function from lab1 into other labs. A second benefit is that this enables easier unittesting of the modularized code.

An ancillary benefit is that this changes the code to be more idiomatic modern JavaScript, from which readers can benefit.

I plan to convert the other labs as well, this is just the first PR. I'm sending it now to make sure we're in agreement about the
approach.


src/lab2.py
```diff
@@ -8,7 +8,7 @@
 import ssl
 import tkinter
 
-def request(url):
+def request(socket, url):
```
chrishtr, line 11: I injected this as a dependency because it resides in a different module.

src/lab2.py
```diff
@@ -65,55 +65,67 @@ def lex(body):
         breakpoint("lex", text)
     return text
 
-WIDTH, HEIGHT = 800, 600
```
chrishtr, line 68: I got rid of these global variables, since they were creating a lot of headaches when trying to set them
from different modules, and upgrade the compiler to match. It seems just as good to pass them as
constructor arguments to Browser.

src/lab2.py
```diff
@@ -65,55 +65,67 @@ def lex(body):
...
-            if y + VSTEP < self.scroll: continue
+            if y > self.scroll + self.height: continue
+            if y + self.vstep < self.scroll: continue
             self.canvas.create_text(x, y - self.scroll, text=c)
 
     def scrolldown(self, e):
         self.scroll += SCROLL_STEP
         self.draw()
 
+def my_breakpoint(arg1=None, arg2=None):
```
chrishtr, line 118: This lab was actually broken before this PR if you tried to run it directly via Python on the commandline,
because the built-in `breakpoint` function does not take any arguments.

chrishtr: I also used arg1 and arg2 because the compiler doesn't seem to like the "*args" syntax.

pavpanchekha: Ah, sorry, yeah. You need to run it with `PYTHONBREAKPOINT=0` set, which disables `breakpoint`.

pavpanchekha:
This is great. The compiler change is small. I imagine eventually `rt-module.js` will just be the normal version of `rt.js`, right? My main concern is whether this requires changes to the code that are unnatural in Python.

The `socket` change—that's because otherwise `socket` isn't defined? We might be able to get around that by collecting all the imports (they are already collected, actually, in `IMPORTS` in the compiler) and adding a JS import for them from `rt-module.js`.

As for turning constants into variables—is it not possible to create module globals that are settable from elsewhere?

I would rather make the compiler a little more complex (readers don't read the compiler!) rather than mangle the Python code.  So for example, we could have the compiler write out a "change_constant" function which changes the value of a given constant, if scoping is a problem.

chrishtr:
> This is great. The compiler change is small. I imagine eventually `rt-module.js` will just be the normal version of `rt.js`, right?

Yes.
 
> The `socket` change—that's because otherwise `socket` isn't defined? We might be able to get around that by collecting all the imports (they are already collected, actually, in `IMPORTS` in the compiler) and adding a JS import for them from `rt-module.js`.

I fixed it by factoring those objects into a named, external js module source in the latest PR revision. (You can't export symbols from an inline module script.)

> As for turning constants into variables—is it not possible to create module globals that are settable from elsewhere?

No. Global variables are always const when exported. IMO the new code is just as readable anyway.

chrishtr:
LGTY now?

pavpanchekha:
So here's why I'm concerned about the constants—we grow more of them as the chapters roll on. In chapter 3, SCROLL_STEP is added; I know in chapter 7 we add CHROME_PX. It's going to look crazy to add more and more of these parameters over time, and readers are not going to enjoy having to thread the variables through, either.

What about modifying the compiler to put all global constants inside an object? So instead of HSTEP, you'd refer to constants.HSTEP, and we could mutate that? I don't think this will be too hard.

chrishtr:
> What about modifying the compiler to put all global constants inside an object? So instead of HSTEP, you'd refer to constants.HSTEP, and we could mutate that? I don't think this will be too hard.

I could do that, or wrap the objects in a new Config object that is passed as a constructor argument to Browser. I don't have a strong opinion, so unless you suggest otherwise I'll do what you suggested.

pavpanchekha:
Let's just do an ugly global object. We can do this entirely in the compiler: it should output an `export const constants = {};`, and then when a variable is compiled (in `compile_expr`) it needs to look up whether that variable is bound in the module context. The best way to do that is to add a method to `Context` objects that recurses. `Context` objects have a string `type` field and at the top level it is (I think) `"module"`, so we can check against that.

chrishtr:
Done.

PR #145, opened 2021-06-20
A different take on the lab1 tests

pavpanchekha:
This PR covers the same ground as #134 but uses Python's `doctest` module so we can organize the tests a little better and also to make the mock-socket library a little more configurable. It is modeled on our JS runtime. The main tests are in `lab1-tests.md` and the `make test` target is set up to run that file. The `test` module provides mocks for the `socket` and `ssl` libraries.


pavpanchekha:
My proposal is that we merge this instead of #134.

chrishtr:
Looks good. I like doing it with doctest, this pairs well with my belief that we should make the code part of the book.

PR #148, opened 2021-06-24
Pass over Chapter 7

pavpanchekha:
This PR does quite a lot, most importantly:

- [x] Redo the LineLayout and TextLayout code to be more in line with the other layout objects.
- [x] Implement multi-tab support (code and text included)
- [x] Remove discussion of the CSS box model

I still have a few issues with the chapter I'd like some feedback on:

- [ ] The LineLayout/TextLayout refactor is really long and gets the chapter off to a slow start.
- [ ] I'm not sure how to order clicking links, multiple tabs, and the address bar

One thing I'm considering is putting multiple tabs first, then the address bar, and then links. I think that leads to the least code changes, but I'm not sure how to motivate multiple tabs before we have anything else. Those tabs would all view the same web page, what's the point?

### DIFF
--- a/book/chrome.md
+++ b/book/chrome.md
@@ -271,10 +271,10 @@ mouse.[^1]
 class Browser:
     def __init__(self):
         # ...
-        self.window.bind("<Button-1>", self.handle_click)
+        self.window.bind("<Button-1>", self.click)
 ```
 
-Inside `handle_click`, we want to figure out what link the user has
+Inside `click`, we want to figure out what link the user has
 clicked on. Luckily, the event handler is passed an "event object",
 whose `x` and `y` fields refer to where the click happened:
 
@@ -286,7 +286,7 @@ relative to the web page, so we need to account for scrolling:
 
 ``` {.python expected=False}
 class Browser:
-    def handle_click(self, e):
+    def click(self, e):
         x, y = e.x, e.y + self.scroll
 ```
 
@@ -295,7 +295,7 @@ that location. To do that, we need search through the tree of layout
 objects:
 
 ``` {.python}
-def handle_click(self, e):
+def click(self, e):
     # ...
     objs = [obj for obj in tree_to_list(self.document, [])
             if obj.x <= x < obj.x + obj.width
@@ -314,7 +314,7 @@ element is on top using the `z-index` property. So real browsers use
 *stacking contexts* to resolve what exactly you actually clicked on.
 
 ``` {.python}
-def handle_click(self, e):
+def click(self, e):
     # ...
     elt = objs[-1].node
 ```
@@ -325,7 +325,7 @@ since we want to know the actual URL the user clicked on, we need to
 climb back up the HTML tree to find the link element:
 
 ``` {.python}
-def handle_click(self, e):
+def click(self, e):
     # ...
     while elt:
         if isinstance(elt, Text):
@@ -335,11 +335,11 @@ def handle_click(self, e):
         elt = elt.parent
 ```
 
-I wrote this in a kind of curious way because this way it's easier to
-add other types of clickable things later---like text boxes and
-buttons in the [next chapter](forms.md). Finally, now that we've found
-the link element itself, we need to extract the URL and direct the
-browser to it. This URL might be a relative URL:
+I wrote this in a kind of curious way, but this way it's easier to add
+other types of clickable things later---like text boxes and buttons in
+the [next chapter](forms.md). Finally, now that we've found the link
+element itself, we need to extract the URL and direct the browser to
+it. This URL might be a relative URL:
 
 ``` {.python}
 # ...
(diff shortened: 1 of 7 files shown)


pavpanchekha:
@chrishtr I think the most important thing to do in your review is to look over the code, specifically, the `LineLayout`, `TextLayout`, `Tab`, and `Browser` classes. Any changes you want to make to these classes or their interfaces we should do now.

book/chrome.md
```diff
@@ -8,531 +8,797 @@ next: forms
...
+        and obj.y <= y < obj.y + obj.height]
+if not objs: return
+```
+
+Now, normally you click on lots of layout objects at once: some text,
+and also the paragraph it's in, and the section that that paragraph is
+in, and so on. We want the one that's "on top", so the last object in
+the list:[^overlap]
+
+[^overlap]: In a real browser you can also have a dialog that overlaps
```
chrishtr, line 312:
```suggestion
[^overlap]: In a real browser you can also have sibling elements that overlap each other, such as a dialog that overlaps
```

book/chrome.md
```diff
@@ -8,531 +8,797 @@ next: forms
...
+some text, or something like that. And real browsers can control which
```
chrishtr, line 313:
```suggestion
some text. And real browsers can control which
```

book/chrome.md
```diff
@@ -8,531 +8,797 @@ next: forms
...
+element is on top using the `z-index` property. So real browsers use
+*stacking contexts* to resolve what exactly you actually clicked on.
```
chrishtr, line 315:
```suggestion
[*stacking contexts*](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context) to resolve what exactly you actually clicked on.
```

book/chrome.md
```diff
@@ -8,531 +8,797 @@ next: forms
...
+while elt:
+    if isinstance(elt, Text):
+        pass
+    elif elt.tag == "a" and "href" in elt.attributes:
+        # ...
+    elt = elt.parent
+```
+
+I wrote this in a kind of curious way, but this way it's easier to add
+other types of clickable things later---like text boxes and buttons in
```
chrishtr, line 338:
```suggestion
other types of clickable things later---like text boxes and buttons---in
```

book/chrome.md
```diff
@@ -8,531 +8,797 @@ next: forms
...
+the [next chapter](forms.md). Finally, now that we've found the link
+element itself, we need to extract the URL and direct the browser to
+it. This URL might be a relative URL:
+
+``` {.python indent=12}
+# ...
+elif elt.tag == "a" and "href" in elt.attributes:
+    url = relative_url(elt.attributes["href"], self.url)
```
chrishtr, line 346: Nit: relative_url is should be called absolute_url. I'll write a PR for this.

book/chrome.md
```diff
@@ -8,531 +8,797 @@ next: forms
...
+
+Fundamentally tabbed browsing means we need to distinguish between the
+browser itself and individual tabs that browse some specific web page.
+Right now the `Browser` class stores both information about the
+browser (like the canvas it draws to) and information about a single
+tab (like the layout tree and display list). We need to tease the two
+apart.
+
+Here's the plan: the `Browser` class will store the window and canvas,
+plus the list of tabs. Everything else goes into a new `Page` class.
```
chrishtr, line 379:
```suggestion
plus a list of `Tab` objects, one for each browser tab. Everything else goes into a new `Page` class.
```

book/chrome.md
```diff
@@ -8,531 +8,797 @@ next: forms
...
+Since the `Browser` stores the window and canvas, it handles all of
+the events, sometimes forwarding it to the active tab.
+
+So the `Tab` constructor looks like this:
```
chrishtr, line 383:
```suggestion
The `Tab` constructor looks like this:
```

book/chrome.md
```diff
@@ -8,531 +8,797 @@ next: forms
...
-now don't need to do anything but recurse on their children.
+The `load`, `scrolldown`, `click`, and `draw` methods all move to
+`Tab`, since that's now where all web-page-specific data lives. But
+there's a change here too.
 
-Navigating between pages
-========================
+Since the `Browser` controls the canvas and handles events, it decides
+when rendering happens and which tab does the drawing. After all, you
+only want one tab drawing its contents at a time! So let's remove the
```
chrishtr, line 398:
```suggestion
only want one tab drawing its contents at a time![^windowed] So let's remove the
```

chrishtr: Plus:

[^windowed] Unless the browser implements multiple windows!

And add an exercise for this.

book/chrome.md
```diff
@@ -8,531 +8,797 @@ next: forms
...
-    # ...
+    def handle_click(self, e):
+        self.tabs[self.active_tab].click(e.x, e.y)
+        self.draw()
 ```
 
-Now, inside `handle_click`, we can convert the link we clicked on to a
-new URL:
+You'll need to modify the arguments `Tab`'s `scrolldown` and `click`
+methods expect. Finally, that `draw` call also just forwards to the
```
chrishtr, line 447: Spell this out?

book/chrome.md
```diff
@@ -8,531 +8,797 @@ next: forms
...
 
-[^10]: Yep, that predates and inspired the name of Google's Chrome
+[^chrome]: Yep, that predates and inspired the name of Google's Chrome
     browser.
 
-To do that, we first have to move the page content itself further down.
-I'm going to reserve 60 pixels for the browser chrome, which we need
-to subtract in `draw`:
+So let's add some code to draw a set of tabs at the top of the browser
+window. Let's try to keep it simple, because this is going to require
```
chrishtr, line 491:
```suggestion
window (and keep it simple, because this is going to require
```

book/chrome.md
```diff
@@ -8,531 +8,797 @@ next: forms
...
+some tedious and mildly tricky geometry. Let's have each tab be 80
```
chrishtr, line 492:
```suggestion
some tedious and mildly tricky geometry). Let's have each tab be 80
```

book/chrome.md
```diff
@@ -8,531 +8,797 @@ next: forms
...
+them independently, and switch tabs whenever you want.
+
+Navigation history
+==================
+
+Now that we are navigating between pages all the time, it's easy to
+get a little lost and forget what web page you're looking at.
+Browsers solve this issue with an address bar that shows the URL.
+Let's implement a little address bar ourselves. The URL it shows is
+the URL for the currently-active tab:
```
chrishtr, line 628:
```suggestion
the one for the currently-active tab:
```

src/lab7.py
```diff
@@ -57,550 +57,618 @@ def request(url):
 
     return headers, body
 
+def relative_url(url, current):
```
chrishtr, line 60: absolute_url

pavpanchekha: You're right. Perhaps `resolve_url`? This is going to need a PR because this function was introduced in the last chapter.

PR #152, opened 2021-07-10
Convert all widgets and js-based browsers to use JS modules

chrishtr:
To do this, I had to change the configuration options bound to tkinter, ssl, and socket to global variables. This is because not all labs use the same options, and otherwise there wasn't a way to set it from the HTML file's module. This also allowed me to remove render.js. A future change can refactor rt.js a bit to remove the need for the tkinter, ssl and socket global variables, as they no longer bind a variable.

### DIFF
--- a/infra/compile.py
+++ b/infra/compile.py
@@ -283,7 +283,7 @@ def compile_function(name, args, ctx):
     if name in RENAME_FNS:
         return RENAME_FNS[name] + "(" + ", ".join(args_js) + ")"
     elif name in OUR_FNS:
-        EXPORTS.append(name)
+     #   EXPORTS.append(name)
         return "await " + name + "(" + ", ".join(args_js) + ")"
     elif name in OUR_CLASSES:
         return "await (new " + name + "()).init(" + ", ".join(args_js) + ")"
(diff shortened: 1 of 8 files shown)


infra/compile.py
```diff
@@ -765,15 +766,14 @@ def compile_module(tree, name, use_js_modules):
 
         imports_str = "import {{ {} }} from \"./{}\";"
 
-        rt_imports_arr = [ 'breakpoint', 'pysplit', 'truthy' ]
-        rt_imports = imports_str.format(",".join(rt_imports_arr), "rt-module.js")
+        rt_imports_arr = [ 'breakpoint', 'comparator', 'filesystem', 'pysplit',
+        'socket', 'ssl', 'tkinter', 'truthy' ]
```
pavpanchekha, line 770: I think here we can do `[ 'breakpoint', 'comparator', 'filesystem', 'pysplit', 'truthy' ] + IMPORTS`. Honestly it does not matter much, but as we write more chapters it'll remove a hassle.

pavpanchekha (approved):
Looks good. I'll execute the minor change I suggested and then merge.

PR #155, opened 2021-07-14
Lab 3 unittests

chrishtr:


chrishtr:
This PR is on top of the lab2 1. It seems github PRs doesn't support dependent PRs. 

I did make a few edits to lab2 to conform to the new test.py changes though.

src/lab3-tests.md
```diff
@@ -0,0 +1,115 @@
...
+
+Testing `lex`
+-------------
+
+The `lex` function in chapter three has been beefed up to return an array
+of `Tag` or `Text` objects, rather than just the stream of characters from the
+input.
+
+    >>> lab3.lex('<body>hello</body>')
+    [Tag: body, Text: hello, Tag: /body]
```
pavpanchekha, line 20: I'm in favor of adding `__repr__` methods to our objects, for testing, but let's not have unquoted user strings here—it could get confusing. Something like `Tag('body')` would be a bit wordier but is less likely to confuse someone accidentally.

src/test.py
```diff
@@ -62,10 +65,105 @@ def wrap_socket(self, s, server_hostname):
...
+        self.weight = weight
+        self.slant = slant
+        self.style = style
+
+    def measure(self, word):
+        return self.size * len(word)
+
+    def metrics(self, name=None):
+        all = {"ascent" : self.size / 2, "descent": self.size / 2,
+            "linespace": self.size * 2}
```
pavpanchekha, line 127: This is all fine but I think in Tk, `linespace = ascent + descent`, it doesn't include suggested interline spacing, so this is a pretty weird font by Tk standards. Also `ascent = descent` is a bit unrealistic; it's usually something like `ascent = size * .75` and `descent = size * .25`. Again, this might not matter much for our tests, since they don't have to be super realistic, but it also doesn't seem like it could hurt?

src/test.py
```diff
@@ -62,10 +65,105 @@ def wrap_socket(self, s, server_hostname):
...
+    if value2:
+        print('breakpoint: name={} value1={} value2={}'.format(
+            name, value1, value2))
+        return
+
+    if value1:
+        print('breakpoint: name={} value1={}'.format(name, value1))
+        return
+
+    print('breakpoint: name={}'.format(name))
```
pavpanchekha, line 161:
```suggestion
def breakpoint(name, **kwargs):
    kwstr = (", " + ", ".join(["{}={}".format(k, v) for k, v in kwargs.items()]) if kwargs else ""
    print('breakpoint: name={}{}'.format(name, kwstr))
```

There's a Python trick here for arbitrary keyword arguments. I have a comment like the one Text/Tag about making the output easy to group by keeping quotes around some of the strings and so on, but it's not that important.

pavpanchekha (approved):
Again, these are really fantastic and I am enthusiastic about merging these!

src/lab3-tests.md
chrishtr: Done

src/test.py
chrishtr: Done.

PR #170, opened 2021-07-28
New focus tracking for Chapter 8

pavpanchekha:
This PR contains *just the text*, so the tests fail, but you should ignore that and just review the text. Overall, the big change here is that we have two levels of focus tracking: once in the `Browser` object and once in the `Tab` object. That was pretty hard to explain well, so please read that part carefully.

One big question I had was whether or not to include blurring; I decided not to, which means if you click in a text box and then in the address bar there will be two cursors onscreen, but I felt that wasn't too bad. The alternative was to add a blur call to tabs, and if we're doing that perhaps we would add explicit focus and blur calls to layout objects, and if we're doing that perhaps text areas should draw their own cursors when focused. This is a cleaner design, but there's more code, and it's not what Chrome actually does, so I decided against.

### DIFF
--- a/book/forms.md
+++ b/book/forms.md
@@ -164,102 +164,147 @@ def input(self, node):
 Try it out: you should now be able to see basic input elements as
 light gray rectangles.
 
+
 Interacting with widgets
 ========================
 
-We've now got input elements rendering, but only as empty rectangles.
-We need the *input* part! Let's 1) allow the user to change that content, and 2) draw the content when it changes.
+We've now got input elements rendering, but when you click on them
+they don't yet do anything! We need to allow the user to type stuff
+into an `input`. Let's make it work the same way the address bar
+does---clicking on an input element will clear it, and then you can
+type into it.
 
-In this toy browser, I'm going to require the user to click on an
-input element to change its content. We detect the click in
-`Browser.handle_click`:
+The clearing part is easy: we need another case inside `Tab`'s `click`
+method:
+
+``` {.python}
+class Tab:
+    def click(self, x, y):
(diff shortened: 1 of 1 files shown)


book/forms.md
```diff
@@ -164,102 +164,147 @@ def input(self, node):
...
+address bar or the active tab:
 
-``` {.python indent=4}
-def keypress(self, e):
-    # ...
+``` {.python}
+class Browser:
+    def handle_key(self, e):
+        # ...
+        elif self.focus == "content":
```
chrishtr, line 245: When changing tabs, you'll have to set self.focus to None

pavpanchekha: I'm not sure I understand this comment, could you elaborate?

Switching tabs happens in `handle_click`, and that sets `Browser.focus` to `None` (see paragraph starting "So, when you click on the web page"). So when you switch tabs keyboard input now goes nowhere.

Or do you mean—when we switch tabs we need to blur the tab we left? As in, do you mean we need to set `Tab.focus` to `None` when we switch tabs? This is a good point—you can switch tabs and the `Tab` has focus somewhere but the `Browser` doesn't have the content focused so typing doesn't go anywhere. I think this is kind of analogous to clicking in the address bar—you can end up with something thinking it's focused when it's not, which is confusing. But at least keyboard entry will go to the right one, and it avoids explicit `blur` calls, which can be moved to an exercise. Does that trade sound bad to you?

chrishtr: I see. Code looks ok as-is then.

chrishtr (approved):

PR #171, opened 2021-07-28
New text on submitting forms

pavpanchekha:
This PR has *just the text*, so the tests fail, but we should ignore that and merge it anyway.

Mostly the changes here are small. First, I'm leveraging `tree_to_list` to make the code that finds all `inputs` inside a `form` clearer. I also implemented proper percent encoding because I realized that implementing "bad" percent encoding is a bad habit security wise. This will require proper decoding on the server, so it's at least two tricky code blocks, but I felt that cost was worth paying.

### DIFF
--- a/book/forms.md
+++ b/book/forms.md
@@ -261,119 +261,104 @@ slow in your browser, so typing into input forms is actually going to
 be quite painful. We'll return to this in [Chapter 10](reflow.md) and
 implement incremental layout to resolve this issue.
 
+
 Implementing forms
 ==================
 
-We're going to need to implement a few different things:
-
--   Buttons
--   Handling button clicks
--   Finding the form containing a button
--   Finding all the input areas in a form
--   Form-encoding their data
--   Making POST requests
-
-First, buttons. Buttons are a bit like `<input>` elements, in that
-they are little rectangles placed in lines. Unlike `<input>` elements,
-`<button>` elements aren't self-closing, and the contents of that
-`<button>` element is what text goes inside the button. Modify
-`InlineLayout` to create `InputLayout` objects for `<button>`
-elements; now, we need to make `InputLayout` do something different in
-its `draw` call.
-
-First, let's give buttons a different color:
+You submit a form by clicking on a `button`. So let's add another
+condition to the big `while` loop in `click`:
 
 ``` {.python}
-class InputLayout:
-    def paint(self, to):
-        # ...
-        bgcolor = \
-            "light gray" if self.node.tag == "input" else "yellow"
-        to.append(DrawRect(x1, y1, x2, y2, bgcolor))
-        # ...
+class Tab:
+    def click(self, x, y):
+        while elt:
+            # ...
+            elif elt.tag == "button":
+                # ...
+            # ...
 ```
 
-Then, buttons should get text from their contents instead of their
-attributes:
-
-``` {.python}
-class InputLayout:
-    def paint(self, to):
-        # ...
-        if self.node.tag == "input":
-            text = self.node.attributes.get("value", "")
-        else:
-            text = self.node.children[0].text
-        # ...
-```
-
-The real reason buttons surround their contents is because a button
-might contain an image, or styled text, or something like that---this
-code doesn't support that, which in real browsers relies on something
-called the `inline-block` display mode. You could implement that by
-having the `InputLayout` have a child `BlockLayout`, but I'm skipping
-it here for simplicity.
-
-Ok, next up, button clicks. We need to extend `handle_click` with
-button support. That requires modifying the condition in the big
-`while` loop and then adding a new case to the big `if` statement:
-
-``` {.python indent=16}
-# ...
-elif elt.tag == "button":
-    self.submit_form(elt)
-# ...
-```
-
-Third, we need to find the form containing our button. That can happen
-inside `submit_form`:[^3]
+Once we've found the button, we need to find the form that it's in, by
+walking up the HTML tree:[^3]
 
 [^3]: Fun fact: HTML standardizes the `form` attribute for _input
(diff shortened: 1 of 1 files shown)


book/forms.md
```diff
@@ -261,119 +261,104 @@ slow in your browser, so typing into input forms is actually going to
...
-while real form-encoding escapes characters like the equal sign, the
-ampersand, and so on. But our browser is a toy anyway, so for now
-let's just try to avoid typing equal signs, ampersands, and so on into
-forms.
+This "percent encoding" replaces all special characters with a percent
+sign followed by those characters' hex codes:
 
-Finally, we need to submit this form-encoded data in a POST request:
+``` {.python}
+def percent_encode(s):
```
chrishtr, line 345: I think urlllib.quote(...) does the same thing. https://stackoverflow.com/questions/1695183/how-to-percent-encode-url-parameters-in-python

Is there extra value to doing it in the text from scratch?

chrishtr: or urllib.urlencode(...)

book/forms.md
```diff
@@ -382,69 +367,56 @@ def submit_form(self, elt):
...
-    body += "Host: {}\r\n".format(host)
-    body += "\r\n" + (payload or "")
-    s.send(body.encode("utf8"))
+    if payload:
+        length = len(payload.encode("utf8"))
+        body += "Content-Length: {}\r\n".format(length)
+    # ...
 ```
 
-Also, when you send a payload like this in the request, you need to
```
chrishtr, line 415: It seems the diff is larger than this PR actually contains. The text here is not actually removed by your chain of edits, right?

book/forms.md
pavpanchekha: Well, this is a good discussion question. I generally have a preference in the book to do things from scratch. But at times I'll bend that rule, like using `makefile` for network reads to avoid looping over socket reads. Since we neither focus on percent encoding in the text nor treat it as a central concept, using urllib.quote or similar might be the right move.

Here's an alternative: what if we skip percent encoding and decoding entirely? Explain that the reader should be thinking about `&` and `=`, the answer is percent encoding, link to it, but don't implement it. It will make our web server act a little weird when used by a normal browser (spaces will become pluses?) but we can just note that in the text. The thinking here is, if we're going to do it in a library, and it's not strictly necessary, what if we don't do it?

The previous revision of this chapter had "crappy" percent encoding, and I thought that taught bad habits, but skipping it entirely seems kind of better—at least it's clear we're not even pretending and we can clearly note that this is an issue.

book/forms.md
pavpanchekha: Looking this over—I think this text is removed because the `Content-Length` header is described (and described as mandatory) in the first "How forms work" section now. You think we should review here? I could see that.

book/forms.md
chrishtr: I think we should keep it, since percent encoding/decoding is actually what happens.

Here are the reasons we should use urllib.quote:
* The code becomes a one-line and doesn't distract from the book contents
* It's not a good practice to implement this kind of thing yourself anyway, for security reasons (and JS has a method built-in that does it)

book/forms.md
chrishtr: Ok that part makes sense, but what about the part that used to be there about the "enter" key?

book/forms.md
pavpanchekha: Alright, I'll make a change to use a library for this.

book/forms.md
pavpanchekha: I removed the Enter key submit handling—it's now a little more involved than before thanks to the Browser/Tab split. Also, I think it would make a good exercise.

book/forms.md
pavpanchekha: The new text says,

> This "percent encoding" replaces all special characters with a percent
> sign followed by those characters' hex codes; for example, a space
> becomes `%20` and a period becomes `%2e`. Python provides a
> percent-encoding function as `quote` in the `urllib` module, or you
> can write your own. (You can even skip percent encoding, but then you
> browser won't handle requests with equal signs, percent signs, or
> ampersands correctly.)

book/forms.md
pavpanchekha: Added the exercise

book/forms.md
```diff
@@ -692,13 +653,22 @@ is better with friends!
...
 elements with the `type` attribute set to `checkbox`. The checkbox is
 checked if it has the `checked` attribute set, and unchecked
 otherwise. Submitting checkboxes in a form is a little tricky,
 though. A checkbox named `foo` only appears in the form encoding if
 it is checked. Its key is its `name` and its value is the empty string.
 
+*Blurring*: Right now, if you click inside a text entry, and then
+inside the address bar, two cursors will appear on the screen. To fix
+this, add a `blur` method to each `Tab` which unfocuses anything that
+is focused, and call it whenever focus leaves that tab.
```
chrishtr, line 670: .. or focus changes to a different input element.

chrishtr (approved):

book/forms.md
```diff
@@ -261,119 +261,93 @@ slow in your browser, so typing into input forms is actually going to
...
+```
+
+Now, any time you see something like this, you've got to ask: what if
+the name or the value has an equal sign or an ampersand in it? So form
+encoding has special handling for special characters:
+
+``` {.python indent=8}
+for input in inputs:
+    # ...
+    name = percent_encode(name)
```
chrishtr, line 336: Also, change this to the python method that does it?

pavpanchekha: Done

pavpanchekha:
@chrishtr is this good to merge?

chrishtr (approved):

PR #173, opened 2021-08-01
Re-use code from lab in lab5

chrishtr:
The duplicated code was print_tree, Element, Text and HTMLParser.

To support this in the compiler, and support for loading the outline of imported classes, so that methods on them
can be called.


pavpanchekha (approved):

pavpanchekha:
It's particularly nice that this shortens the hints files.

PR #178, opened 2021-08-07
Lab 8 fixes

chrishtr:
There were a number of typos / bugs in lab8 and server 8:

* Two typos in server8
* Not updating browser8.css to include new styles
* Some missing parameters to a helper function in lab8
* Failing to compute a font object
* Painting text children of buttons twice
* Failing to re-paint the display list after invalidating the value attribute of an input.


chrishtr:
I think an audit of the text in the chapter is also needed, to make sure it explains well enough how to write this code correctly and all the subtleties.

Also, we need unittests! (I'll do that next week.)

PR #184, opened 2021-08-18
More "Go Further" blocks

pavpanchekha:
I added "Go Further" blocks on the history of HTML forms, multipart encoding, and some server stuff.

### DIFF
--- a/book/forms.md
+++ b/book/forms.md
@@ -549,16 +549,21 @@ and asynchronous uses of forms are based on the same principles.
 [web20]: https://en.wikipedia.org/wiki/Web_2.0
 
 ::: {.further}
-There are more types of request than GET and POST. For example, there
-is [PUT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT),
-which is like POST but is idempotent---if you send the same request multiple
-times, the effect is the same as if it has been sent once. PUT models
-placing a new keyed entry in a database. Another example is
-[DELETE](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/DELETE),
-which "deletes the specified resource".
-
-Neither PUT nor DELETE are commonly used.
+There are request types besides GET and POST, like [PUT][put-req]
+(create if nonexistant) and [DELETE][del-req]. There are also more
+obscure methods like CONNECT and TRACE. New methods were supposed to
+be a standard extension mechanism for HTTP, and some protocols like
+[WebDav][webdav] were built this way (with methods like PROPFIND,
+MOVE, and LOCK), and in 2010 the [PATCH method][patch-req] was
+standardized in [RFC 5789]. Still, this did not become an enduring way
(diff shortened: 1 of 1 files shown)


book/forms.md
```diff
@@ -70,6 +70,18 @@ a great starting point for transforming our toy browser into an
 application platform---our goal for these next few chapters. Let's get
 started implementing all that!
 
+::: {.further}
+HTML forms first standardized in [HTML+][htmlplus], which also
```
chrishtr, line 74:
```suggestion
HTML forms were first standardized in [HTML+][htmlplus], which also
```

book/forms.md
```diff
@@ -70,6 +70,18 @@ a great starting point for transforming our toy browser into an
...
+proposed tables, mathematical equations, and text that wraps around
+images. Amazingly, all three of these technologies survive, but in
+totally different standards: tables in [RFC 1942][rfc1942], equations
+in [MathML][mathml], and floating images in [CSS 1.0][css1].
+:::
+
+[htmlplus]: https://www.w3.org/MarkUp/htmlplus_paper/htmlplus.html
+[rfc1942]: https://datatracker.ietf.org/doc/html/rfc1942
```
chrishtr, line 82: And https://drafts.csswg.org/css-tables-3/

book/forms.md
```diff
@@ -203,6 +215,8 @@ The reason buttons surround their contents but input areas don't is
 that a button can contain images, styled text, or other content. In a
 real browser, that relies on the [`inline-block`][inline-block]
 display mode: a way of putting a block element within an inline.
+There's also `<input type=button>`, an older button syntax which
+worked more like other `input` areas.
```
chrishtr, line 219:
```suggestion
works more like other `input` areas.
```

book/forms.md
```diff
@@ -827,6 +868,21 @@ Try it! You should be able to restart the server, open it in your
 browser, and update the guest book a few times. You should also be
 able to use the guest book from a real web browser.
 
+::: {.further}
```
chrishtr, line 871: Does it make sense to call out specific technologies? It feels a bit like advertising, since there are many other options besides these.

book/forms.md
pavpanchekha: I'm not sure HTML plus actually ever standardized, say, what tables were supposed to look like, so I don't know if CSS 3 is the right link here. (Plus I kind of like the asymmetry of tables, math, and figures all ending up in different WGs.)

book/forms.md
pavpanchekha: I removed the mentions of Flask and Django. It maybe feels appropriate to mention bottle, since we use it?

book/forms.md
chrishtr: ok

book/forms.md
chrishtr: sgtm

PR #189, opened 2021-08-22
Add some more flavor about JS, and a go-further block

chrishtr:


pavpanchekha:
Edits to script running look good and I like the new "Go Further". I generally put just one short summary paragraph before the first section (the "head paragraph") and would like to keep that style. Can we move the paragraph on progressive enhancement and the "Go Further" to the "running JavaScript" section?

chrishtr:
I couldn't make it fit well in the "running JavaScript" section, since that is entirely focused on the implementation and not concepts.

I put it in the Summary section instead.

PR #193, opened 2021-08-27
Update Chapter 9 to match earlier chapters' code

pavpanchekha:
This PR changes the Chapter 9 code and text to match previous chapters. Beyond that, the only significant change is using the `keydown` instead of the `change` event, both to match real browser behavior and to make it cancellable.

### DIFF
--- a/book/scripts.md
+++ b/book/scripts.md
@@ -232,7 +232,7 @@ can call that JavaScript code our "JavaScript runtime"; we run it
 before we run any user code, so let's stick it in a `runtime.js`
 file that's run in `setup_js`:
 
-``` {.python replace=runtime/rt9}
+``` {.python replace=runtime/runtime9}
 def setup_js(self):
     # ...
     with open("runtime.js") as f:
--- a/book/scripts.md
+++ b/book/scripts.md
@@ -812,7 +812,7 @@ from local variables in the `load` method to new fields on a `Tab`.
 Note that styling moved from `load` to `render`, but downloading the
 stylesheets didn't. That's because `innerHTML` created new elements
 that have to be styled, but we don't need to redownload the styles to
-do that; we just need to reapply the styles we already have.
+do that; we just need to re-apply the styles we already have.
 
 Now, whenever the page changes, we can lay it out again by calling
 `render`:
--- a/book/scripts.md
+++ b/book/scripts.md
@@ -610,7 +610,7 @@ it needs to run the handlers set up by `addEventListener`, so those
 need to be stored somewhere. Since those handlers are JavaScript
 functions, we need to keep that data on the JavaScript side, in an
 variable in the runtime. I'll call that variable `LISTENERS`; we'll
-use it to look up handles and event types, so let's make map handles
+use it to look up handles and event types, so let's make it map handles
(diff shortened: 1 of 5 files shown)


book/scripts.md
```diff
@@ -197,21 +195,21 @@ def setup_js(self):
...
+When this call happens, Dukpy converts the JavaScript string `"Hi from
+JS"` into a Python string,^[This conversion also works on numbers,
+string, and booleans, but I wouldn't try it with other objects.] and
+then passes that Python string to the `print` function we registered.
 
-We actually want a `console.log` function, not a `call_python`
-function, so we need to define a `console` object and then give it a
-`log` property. We do that *in JavaScript*, with code like this:
+Since we ultimately want JavaScript to call a `console.log` function,
+not a `call_python` function, so we need to define a `console` object
```
chrishtr, line 211:
```suggestion
not a `call_python` function, we need to define a `console` object
```

book/scripts.md
```diff
@@ -234,15 +232,16 @@ can call that JavaScript code our "JavaScript runtime"; we run it
 before we run any user code, so let's stick it in a `runtime.js`
 file that's run in `setup_js`:
 
-``` {.python expected=False}
+``` {.python replace=runtime/runtime9}
```
chrishtr, line 235: What is this alchemy??

book/scripts.md
```diff
@@ -600,50 +577,41 @@ method and call it whenever an event is generated. First, any time we
...
 it needs to run the handlers set up by `addEventListener`, so those
-need to be stored somewhere. Where? I propose we keep that data on the
-JavaScript side, in an variable in the runtime. I'll call that
-variable `LISTENERS`; we'll use it to look up handles and event types,
-so let's make map handles to a dictionary that maps event types to a
-list of handlers:
+need to be stored somewhere. Since those handlers are JavaScript
+functions, we need to keep that data on the JavaScript side, in an
+variable in the runtime. I'll call that variable `LISTENERS`; we'll
+use it to look up handles and event types, so let's make map handles
```
chrishtr, line 613:
```suggestion
use it to look up handles and event types, so let's make it map handles
```

book/scripts.md
```diff
@@ -809,68 +780,79 @@ def js_innerHTML(self, handle, s):
...
+        self.display_list = []
+        self.document.paint(self.display_list)
 ```
 
-and then move the call to `style` into `layout`:
+For this code to work, you'll also need to change `nodes` and `rules`
+from local variables in the `load` method to new fields on a `Tab`.
+Note that styling moved from `load` to `render`, but downloading the
+stylesheets didn't. That's because `innerHTML` created new elements
+that have to be styled, but we don't need to redownload the styles to
```
chrishtr, line 814:
```suggestion
that have to be styled, but we don't need to re-download the styles to
```

book/scripts.md
```diff
@@ -809,68 +780,79 @@ def js_innerHTML(self, handle, s):
...
+do that; we just need to reapply the styles we already have.
```
chrishtr, line 815:
```suggestion
do that; we just need to re-apply the styles we already have.
```

chrishtr (approved):

PR #198, opened 2021-08-28
Update lab 10 and server 10

chrishtr:
* to match code form lab9. I did this by copying lab9 to lab10, then manually adding in cookie support by comparison
* to fix various bugs (now works correctly for login, etc)

I manually verified that the server and lab work correctly together (cookie login etc)

### DIFF
--- a/book/security.md
+++ b/book/security.md
@@ -77,8 +77,10 @@ Then, when responses are processed:
 def load(self, url, body=None):
     # ...
     if "set-cookie" in headers:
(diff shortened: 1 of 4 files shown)


chrishtr:
One reason I made this is because chapter 10 is coming up soon, so this will of course need to be done.

Another is that I'm writing code for chapter 12, and I needed a baseline that was not out of date.

Chapter 11's code is still out of date, that'll need to be fixed later.

PR #201, opened 2021-08-31
Code for chapter 12

chrishtr:
* The code so far is just lab11.py, with Skia + SDL instead of Tkinter
* A first section is added to the chapter text explaining the code changes made for the above


chrishtr:
Going to commit this to keep going on chapter 12. Full review later.

pavpanchekha:
Quite right, don't wait on me.

PR #203, opened 2021-09-01
Add font caching to Chapter 3

pavpanchekha:
This PR takes the Chapter 13 code and text on font caching and ports it to Chapter 3. I think it's a better place since Ch3 already discusses text, while Ch13 is mostly about something else; plus, it means the browser won't be super-slow for most of the book.

I updated the code for Chapters 3-9 to account for this change, but didn't remove the relevant part of Chapter 13, since we might want further reorganization there.

### DIFF
--- a/book/text.md
+++ b/book/text.md
@@ -816,13 +816,12 @@ class Layout:
 ```
 
 ::: {.further}
-Fonts are surprisingly large, sometimes on the order of multiple
-megabytes for scripts like Chinese with a lot of different, complex
-characters. For this reason they are generally stored on disk and only
-loaded into memory on-demand, which is slow. Optimizing font loading
-(and the cost to shape them and lay them out, since many web pages
-have a *lot* of text) turns out to be one of the most important parts
-of speeding up text rendering.
+Fonts for scripts like Chinese can be megabytes in size, so they are
+generally stored on disk and only loaded into memory on-demand. That
+makes font loading slow. Browsers also have extensive caches for
+measuring, shaping, and rendering text. Because web pages have a lot
+of text, these caches turn out to be one of the most important parts
+of speeding up rendering.
 :::
 
 Summary
--- a/book/text.md
+++ b/book/text.md
@@ -495,7 +495,7 @@ def token(self, tok):
 In fact, the body of the `isinstance(tok, Text)` branch can be moved
 to its own method:
 
-``` {.python replace=16/self.size indent=4}
+``` {.python expected=False}
 def text(self, tok):
     font = tkinter.font.Font(
         size=16,
@@ -545,7 +545,7 @@ self.size = 16
 
 That variable is used to create the font object:
 
-``` {.python}
+``` {.python expected=False}
 font = tkinter.font.Font(
     size=self.size,
     weight=self.weight,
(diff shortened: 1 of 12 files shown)


book/text.md
```diff
@@ -748,6 +748,75 @@ styles. A particular challenge is [Mongolian script][mongolian].
...
+
+Now that you've implemented styled text, you've probably
+noticed---unless you're on macOS[^macos-cache]---that on a large web
+page like this chapter your browser has slowed significantly from the
+[last chapter](graphics.md). That's because text layout, and
+specifically the part where you measure each word, is quite slow.
+
+[^macos-cache]: The macOS text APIs do fairly complex system-wide font
+    caching that Linux and Windows do not do. The optimization
+    described in this section won't hurt any.
```
chrishtr, line 762:
```suggestion
    described in this section won't hurt any on that platform, though.
```

book/text.md
```diff
@@ -748,6 +748,75 @@ styles. A particular challenge is [Mongolian script][mongolian].
...
+measuring text can require pretty complex computations. But on a large
+web page, some words likely appear a lot---for example, this page
+includes the word "the" over two hundred times. Instead of measuring
+these words over and over again, we could measure them once, and then
+cache the results. On normal English text, this usually results in a
+substantial speedup.
+
+Caching is such a good idea that Tk already implements it internally.
+But there's a small catch: Tk stores the cache inside each `Font`
+object. But our `text` method creates a new `Font` object for each
```
chrishtr, line 775: Did you verify this?

chrishtr (approved):

chrishtr:
Actually - please don't merge until the one question I asked is answered.

book/text.md
pavpanchekha: Good point. Working on it.

pavpanchekha: Ok—here's what I learned. Creating a `tkinter.font.Font` object the way we do calls the `font create` command in Tk. That calls the `FONT_CREATE` branch in `Tk_FontObjCmd` method in `generic/tkFont.c`. The important thing is that in this process, `tkinter.font.Font` chooses a new name for the font, which causes `Tk_FontObjCmd` to create a new `Tk_Font` object, because the cache is keyed by name. Ok, so now we need to know what the created `Tk_Font` does. This is platform-specific, so we go to `unix/tkUnixFont.c` as an example, and if I follow the code correctly, the underlying X font is created in `GetScreenFont`, with no cache.

So I think what is ultimately happening here is that caching the `tkinter.font.Font` objects reuses the underlying system fonts—`XFontStruct`s and windows `HFONT`s, which I assume ultimately means that system-level caching kicks in. I'll revise the text to explain this better.

book/text.md
```diff
@@ -748,6 +748,83 @@ styles. A particular challenge is [Mongolian script][mongolian].
...
+        font = get_font(self.size, self.weight, self.style)
+        # ...
+```
+
+::: {.further}
+Fonts are surprisingly large, sometimes on the order of multiple
+megabytes for scripts like Chinese with a lot of different, complex
+characters. For this reason they are generally stored on disk and only
+loaded into memory on-demand, which is slow. Optimizing font loading
+(and the cost to shape them and lay them out, since many web pages
```
chrishtr, line 823: Maybe remove this go-further block? The last sentence is kind of redundant with what you've already said.

pavpanchekha:
Ok, I did a pass on the "Go Further" block. I think it's not redundant—but it would be nice to link to some description of fast text rendering or the caches involved. @chrishtr, do you know if there is, say, a good description of the caches in Harfbuzz? I know Patrick Walton has some great blog posts on Pathfinder text rendering, and I'd love to link to those but they are on a slightly different topic

pavpanchekha:
In either case, I think we should merge this, not least so we can unravel the discussion in Ch 13 that this is pulled from and also because my class is going to get here soon.

PR #205, opened 2021-09-09
Change the standard line spacing from 1.2 to 1.25

pavpanchekha:
The point of changing the standard line spacing is to avoid floating-point round-off error in our tests. (Yes it is dumb.) I also reordered the soft-hyphens and small-caps exercises because in retrospect small-caps is harder.

I'll work on getting all the tests fixed up before merging, but I'm deploying this to the draft env now for benefit of my students.

### DIFF
--- a/book/text.md
+++ b/book/text.md
@@ -863,10 +863,6 @@ should be smaller (perhaps half the normal text size) and be placed so
 that the top of a superscript lines up with the top of a normal
 letter.
 
-*Small caps:* Make the `<abbr>` element render text in small caps,
-<abbr>like this</abbr>. Upper-case letters should be in a normal font,
-while lower-case letters should be small, capitalized, and bold.
-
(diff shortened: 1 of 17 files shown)


book/text.md
```diff
@@ -669,15 +669,16 @@ max_ascent = max([metric["ascent"] for metric in metrics])
 The line is then `max_ascent` below `self.y`—or actually a little more
 to account for the leading:[^leading-half]
 
-[^leading-half]: Actually actually, 20% leading doesn't add 20% of the
-    ascender above the ascender and 20% of the descender below the
-    descender. Instead, it adds [10% of the line height in both
-    places][line-height-def].
+[^leading-half]: Actually actually, 25% leading doesn't add 25% of the
```
chrishtr, line 672:
```suggestion
[^leading-half]: Actually, 25% leading doesn't add 25% of the
```

chrishtr (approved):
LGTM with tests passing, and the one edit.

PR #206, opened 2021-09-12
Revisions to chapter 9

chrishtr:
* Some typo fixes
* Improvements to text descriptions
* Fixes a bug in comment9.js and server9.py: it was using an id selector, but the browser only supports tag selectors at present

### DIFF
--- a/book/scripts.md
+++ b/book/scripts.md
@@ -5,7 +5,7 @@ prev: forms
 next: security
 ...
 
-But form-based web applications, like [last chapter's guest
+Form-based web applications, like [last chapter's guest
 book](forms.md), reload the page every time the user does anything,
 and fell out of favor in the early 2000s. What took their place are
 JavaScript-enhanced web applications, which can respond to user input
@@ -199,14 +199,14 @@ that the other uses, say on a page like this:
 
 Suppose `a.js` is "`var x = 2;`" and `b.js` is "`console.log(x + x)`";
 the variable `x` is set in `a.js` but used in `b.js`. In real web
-browsers, that's important since one script might define library
+browsers, that's important, since one script might define library
 functions that another script wants to call.
 
 Let's start exporting functions. The JavaScript function `console.log`
 corresponds to the Python `print` function.[^5] We can leverage this
 correspondence using `export_function`:
 
-[^5]: If you're using Python 2, for some reason, you'll need to write
+[^5]: If you're using Python 2, you'll need to write
     a little wrapper function around `print` instead.
 
 ``` {.python replace=__init__(self)/__init__(self%2c%20tab)}
@@ -244,6 +244,7 @@ there.
 
 [^brush-up]: Now's a good time to [brush up][mdn-js]---this chapter
     has a ton of JavaScript!
+
 [mdn-js]: https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/A_first_splash
 
 Taking a step back, when we run JavaScript in our browser, we're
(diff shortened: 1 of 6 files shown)


book/scripts.md
```diff
@@ -878,7 +886,7 @@ do that; we just need to re-apply the styles we already have.[^update-styles]
     `link` nodes are created. This is tricky to get right, so I'm
     skipping it here.
 
-Now, whenever the page changes, we can lay it out again by calling
+Now, whenever the page changes, we can update its rendering again by calling
```
chrishtr, line 889: Doesn't it need to also call draw? I think the browser has a bug in that render happens but not draw in this case.

chrishtr: I answered my own question and added a footnote about it.

book/scripts.md
```diff
@@ -659,7 +665,6 @@ first, any time we click in the page:
 ``` {.python expected=False}
 class Tab:
     def click(self, x, y):
-        while elt:
```
pavpanchekha, line 662: Perhaps this is confusing, but I usually try to add "nesting" hints like this so readers can find the right line of code. Plus it justifies indenting things the right amount.

book/scripts.md
```diff
@@ -864,6 +872,13 @@ class Tab:
         self.document.paint(self.display_list)
 ```
 
+[^why-no-draw]: Why is `Browser.draw` not one of the phases that has to run?
+After all, running `paint` will have no visible effect until `draw` happens.
+It's technically not necessary right now because the only way to cause a script
+to run is loading or event handling, and you'll see that in  all such cases,
+such as `handle_key`, `draw` is called at the end. (All of this will be cleaned
+up and made more precise in [Chapter 13](scheduling-and-threading.md).)
```
pavpanchekha, line 880:
```suggestion
such as `handle_key`, `draw` is called at the end. (Subtleties like this will be
clarified and made more precise in [Chapter 13](scheduling-and-threading.md).)
```

src/server9.py
```diff
@@ -60,7 +60,7 @@ def show_comments():
     for entry in ENTRIES:
         out += "<p>" + entry + "</p>"
     out += "<link rel=stylesheet src=/comment.css>"
-    out += "<div id=errors></div>"
+    out += "<errors></errors>"
```
pavpanchekha, line 63: Why use a non-standard element here? Also doesn't this now not match the text?

pavpanchekha (changes requested):

book/scripts.md
chrishtr: I removed it because I found the "while elt" confusing when reading the chapter, as there is no indication of why it's there yet. I think it will be easy to find the code because it's in the same method.

src/server9.py
chrishtr: Well, we have to do it base don a tag, because that's the only selector we have. Which one do you want?

https://way2tutorial.com/html/tag/index.php

I could do `<span>` just because it doesn't exist elsewhere, or `<pre>`?

book/scripts.md
pavpanchekha: OK. Can we un-indent the later lines of code? (Is the differ unhappy with that?)

src/server9.py
pavpanchekha: Oh man, I'm a moron, I forgot about this. I think technically `<label>` is recommended, since it describes an input element.

chrishtr:
```suggestion
    out += "<label></label>"
```

src/comment9.js
```diff
@@ -1,5 +1,5 @@
 var allow_submit = true;
-var p_error = document.querySelectorAll("#errors")[0];
+var p_error = document.querySelectorAll("errors")[0];
```
chrishtr, line 2:
```suggestion
var p_error = document.querySelectorAll("label")[0];
```

pavpanchekha (approved):

PR #217, opened 2021-09-17
WIP: chapter 12 code updates

chrishtr:
This PR finishes rewriting all of the chapter to correspond to actual code in the browser (*), adds examples, and some
exercises.

From here on out, the remaining edits should be mostly editorial and pedagogical.

(*) Except for some Python code to implement the various compositing and blend operations.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,3 @@
+# These are supported funding model platforms
+
+patreon: browserengineering
--- a/.github/workflows/compare.yml
+++ b/.github/workflows/compare.yml
@@ -10,6 +10,8 @@ jobs:
       - uses: actions/setup-python@v2
         with:
           python-version: 3.9
+      - name: Install dependencies
+        run: python -m pip install dukpy
       - uses: r-lib/actions/setup-pandoc@v1
         with:
           pandoc-version: 2.11.4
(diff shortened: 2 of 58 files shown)


pavpanchekha:
Hi @chrishtr, I fixed some but not all of the issues with loading images. Among the changes made:

 - I changed the CSSParser to treat single quotes as part of words, so we can load `url('...')` images.
 - I switched the `request` function to decode to `utf8` bit by bit instead of forcing the whole socket to be decoded. This way avoids having to use `recv` or similar.
 - I used PIL's `convert` method to make sure I'm output RGBA pixels to Skia

This does show images, but seems to swap or mess up some color channels—it wasn't immediately obvious to me what the problem is and I'm hoping this one is easier for you to debug.

chrishtr:
> This does show images, but seems to swap or mess up some color channels—it wasn't immediately obvious to me what the problem is and I'm hoping this one is easier for you to debug.

It works for me on the image I tested with. Which image did you test with that was broken?

pavpanchekha:
Here's a bug I can reproduce:

[err.tar.gz](https://github.com/browserengineering/book/files/7254501/err.tar.gz)

image

Note that the image is colored super weird.

Also a few bug reports:

- [ ] Background images inside `style` attributes aren't downloaded.
- [ ] The full 194x194 image is drawn, the background isn't clipped to 100x100 as they should be
- [ ] The text should be inside the image, but is located half-outside.

pavpanchekha:
Now that I think about, it seems less like a channel swap and more like an inversion...

chrishtr:
Weird. That same image works for me. I wonder if there is something about
your pillow installation.

PR #220, opened 2021-09-22
Add support for from-imports to the outliner

pavpanchekha:
This PR allows the outliner to inline from-imported functions and classes. This means the outlines in each chapter are once again complete. I also added support for from-importing multiple things at a time, and fixed a few minor test bugs.

The order of the inlined elements depends on the order of the from-imports. Right now they're not ordered in the ideal way, but it's at least complete.

### DIFF
--- a/infra/compiler.md
+++ b/infra/compiler.md
@@ -281,6 +281,12 @@ deduplicate the code a bit:
     >>> Test.stmt("from lab1 import request")
     import { request } from "./lab1.js";
     >>> assert "request" in LAB_IMPORT_FNS
+    >>> Test.stmt("from lab2 import HSTEP")
+    import { HSTEP } from "./lab2.js";
+    >>> assert "HSTEP" in LAB_IMPORT_CONSTANTS
+    >>> Test.stmt("from lab4 import Element")
+    import { Element } from "./lab4.js";
+    >>> assert "HSTEP" in LAB_IMPORT_CLASSES
     >>> Test.stmt("from lab2 import WIDTH, HEIGHT, HSTEP, VSTEP")
     import { WIDTH, HEIGHT, HSTEP, VSTEP } from "./lab2.js";
     >>> assert "WIDTH" in LAB_IMPORT_CONSTANTS
--- a/infra/compiler.md
+++ b/infra/compiler.md
@@ -4,39 +4,38 @@ WBE's Python-to-JS compiler
 This file is mostly unit tests, but you could pretend it is
 documentation if you were so inclined.
 
-```
-This is some helper code for tests. First, import the compiler:
+First, some helper code for tests. We need import the compiler:
 
->>> import sys
->>> sys.path.append("infra")
->>> from compile import *
->>> test_mode()
+    >>> import sys
+    >>> sys.path.append("infra")
+    >>> from compile import *
+    >>> test_mode()
 
-Then, this pretends all variables are in scope:
+Then, this fake context pretends all variables are in scope, which is
+useful for short tests:
 
->>> class FakeCtx(dict):
-...    type = "module"
-...    def __getitem__(self, i): return True
-...    def __contains__(self, i): return True
-...    def is_global_constant(self, i): return False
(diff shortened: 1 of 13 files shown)


infra/compiler.md
```diff
@@ -276,6 +275,16 @@ compiler state:
     // Please configure the 'ssl' module
     >>> assert "ssl" in RT_IMPORTS
 
+However, `from ... import` statements *are* supported, and are used to
+deduplicate the code a bit:
+
+    >>> Test.stmt("from lab1 import request")
```
chrishtr, line 281: Also test importing a class name.

src/lab5.py
```diff
@@ -9,16 +9,9 @@
...
-from lab4 import HTMLParser
-from lab4 import Text            
-
-WIDTH, HEIGHT = 800, 600
-HSTEP, VSTEP = 13, 18
-
-SCROLL_STEP = 100
+from lab2 import WIDTH, HEIGHT, HSTEP, VSTEP, SCROLL_STEP
+from lab3 import FONTS, get_font
+from lab4 import Text, Element, print_tree, HTMLParser
```
chrishtr, line 14: order all these alphabetically

src/lab6.py
```diff
@@ -309,6 +276,51 @@ def __repr__(self):
         return "InlineLayout(x={}, y={}, width={}, height={})".format(
             self.x, self.y, self.width, self.height)
 
+class BlockLayout:
```
chrishtr, line 279: Did this class change in lab6 but we failed to notice it?

src/lab7-tests.md
```diff
@@ -37,7 +37,7 @@ Here is how the lines are represented in chapter 7:
      DocumentLayout()
        BlockLayout(x=13, y=18, width=774, height=45.0)
          BlockLayout(x=13, y=18, width=774, height=45.0)
-           LineLayout(x=13, y=18, width=774, height=45.0)
+           InlineLayout(x=13, y=18, width=774, height=45.0)
```
chrishtr, line 40: Is this because of the lab6 change to BlockLayout?

src/lab7-tests.md
```diff
@@ -60,9 +60,9 @@ has the same total height:
     >>> browser2.load(url)
     >>> lab6.print_tree(browser2.document)
      DocumentLayout()
-       BlockLayout(x=13, y=18, width=774, height=60.0)
-         BlockLayout(x=13, y=18, width=774, height=60.0)
-           InlineLayout(x=13, y=18, width=774, height=60.0)
+       BlockLayout(x=13, y=18, width=774, height=45.0)
```
chrishtr, line 63: Ditto.

src/lab6.py
pavpanchekha: Ah, no. This is subtle. Since `BlockLayout` creates `InlineLayout` objects, if you import `BlockLayout` from lab 5 it will use lab 5's `InlineLayout` even if in lab 6 we define a new one. (This is an argument, perhaps, for construing the layout tree separately from layout.)

src/lab7-tests.md
pavpanchekha: Nah, the `__repr__` for `InlineLayout` called itself `LineLayout`; I assume it's a copy-and-paste error.

src/lab7-tests.md
pavpanchekha: This one is due to adding `BlockLayout`.

src/lab5.py
pavpanchekha: The order they're listed is the order they end up in the outline, so I think keeping a "conceptual" order is actually important here.

src/lab6.py
chrishtr: Ok, you're fixing that bug then. Got it.

src/lab7-tests.md
chrishtr: ..and therefore fixing a bug in that InlineLayout was wrong.

chrishtr (approved):

PR #223, opened 2021-09-29
Rewrite Chapter 10 to use sessions from the start

pavpanchekha:
This PR rewrite the start of Chapter 10.

*Sessions*: The server now implements sessions using a token cookie from the very beginning. This avoids presenting readers with insecure code (which they might remember better than later fixes!) and also helps us avoid focusing too much on the server.

*One cookie*: The browser now implements a single cookie per origin. Crazy, but it simplifies the code quite a bit, doesn't hurt anything else in the chapter, and makes the lack of support for multiple `Set-Cookie` headers less bad.

### DIFF
--- a/book/security.md
+++ b/book/security.md
@@ -33,9 +33,9 @@ look just like those of a new visitor.
 
 The web fixes this problem with cookies. A cookie---the name is
 meaningless, ignore it---is a little bit of information stored by your
-browser on behalf of a web server. The cookie establishes that
-browser's identity and allows the server to distinguish one web
-request from another.
+browser on behalf of a web server. The cookie distinguishes your
+browser, and is sent with each web request so the server can
+distinguish its users.
 
 [^1]: I don't mean anonymous against malicious attackers, who might
     use *browser fingerprinting* or similar techniques to tell users
--- a/book/security.md
+++ b/book/security.md
@@ -23,13 +23,18 @@ you're writing security-sensitive code, this book is not enough.
 Cookies
 =======
 
-The most basic kind of user data on the web is the *cookie*. A
-cookie---the name is meaningless, ignore it---is a little bit of
-information stored by your browser on behalf of a web server. It's
-that information that allows the server to distinguish one web request
-from another. Bereft of cookies, your web browser is effectively
-anonymous:[^1] it isn't logged in anywhere so it can't do any useful
-personalization or show you your private data.
+With what we've implemented so far there's no way for a web server to
+tell whether two HTTP requests come from the same browsers, or
+different ones. Our web browser is effectively anonymous.[^1] That
+means it can't "log in" anywhere, because there's no way for the
+server to know which later requests come from the logged in browser
+and which come from some unrelated browser.
+
+The web fixes this problem with cookies. A cookie---the name is
+meaningless, ignore it---is a little bit of information stored by your
+browser on behalf of a web server. The cookie establishes that
+browser's identity and allows the server to distinguish one web
+request from another.
 
 [^1]: I don't mean anonymous against malicious attackers, who might
     use *browser fingerprinting* or similar techniques to tell users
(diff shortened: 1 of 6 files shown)


pavpanchekha:
This chapter is factually incorrect: for cookies, the "scheme" is not used in the same-origin policy. If we build a nice cookie-parameter system we can throw in the "secure" attribute.

pavpanchekha:
Actually this still isn't right, because for cookies SOP doesn't include ports if they're not explicitly stated—otherwise HTTP and HTTPS generally wouldn't match. Ugh.

src/server10.py
```diff
@@ -18,123 +21,104 @@ def handle_connection(conx):
...
     out += "<script src=/comment.js></script>"
     return out
 
-def check_nonce(params, username):
-    if 'nonce' not in params: return False
-    if username not in NONCES: return False
-    return params['nonce'] == NONCES[username]
+def login_form(session):
+    body = "<!doctype html>"
+    body += "<form action=/login method=post>"
```
chrishtr, line 95: Why should this be at "/"?  The login operation should end up at "/", so either that needs to be the POST destination or we need to support redirects.

book/security.md
```diff
@@ -38,418 +43,314 @@ personalization or show you your private data.
...
+Try it out in a normal web browser. You should be able to go to the
+main guest book page, click the link to log in, use one of the
+username/password pairs above, and post entries.^[The login flow slows
+down debugging. You might want to add the empty string as a
+username/password pair.] Of course, this login system has a whole slew
+of insecurities.[^7] But the focus of this book is the browser, not
+the server, so once you're sure it's all working, let's switch gears
+and implement cookies inside our own browser.
 
 [^7]: I should be hashing passwords! Using `bcrypt`! We should verify
```
chrishtr, line 279:
```suggestion
[^7]: The insecurities include not hashing passwords, not using `bcrypt`, not verifying
```

book/security.md
```diff
@@ -38,418 +43,314 @@ personalization or show you your private data.
...
-    email addresses! Over TLS! We should be comparing passwords in a
-    constant-time fashion!
-
-Changing your login
-===================
+    email addresses! Over TLS! We should run the server in a sandbox!
```
chrishtr, line 280:
```suggestion
    email addresses, not forcing TLS, and not running the server in a sandbox.
```

chrishtr (commented):
Is "Keeping Data Private" the best name for this chapter? It seems like the chapter is now just Cookies?

src/lab10.py
```diff
@@ -56,8 +58,9 @@ def request(url, headers={}, payload=None):
     method = "POST" if payload else "GET"
     body = "{} {} HTTP/1.0\r\n".format(method, path)
     body += "Host: {}\r\n".format(host)
-    for header, value in headers.items():
```
chrishtr, line 59: Why did you delete this part?

src/lab10.py
```diff
@@ -138,7 +149,8 @@ def innerHTML(self, handle, s):
         self.tab.render()
 
     def cookie_string(self):
-        return self.tab.cookie_string()
+        # origin ???
```
chrishtr, line 152: Do you intend to implement it before publishing this chapter?

pavpanchekha: Yes—it's implemented in the next PR where we actually start using this.

src/lab10.py
pavpanchekha: So, in earlier versions of this chapter we took a map of extra headers as an input to `request`. It was only used for cookies. Now I handle cookies directly in `request` and don't need that argument.

Passing in a headers dict was done incorrectly in edge cases (you could send the Host header twice, for example) and using it for cookies was also hard / incorrect (you need to handle `Cookie`/`Set-Cookie` at every request, which we weren't doing). It also didn't work well with a multi-tab design. Plus it shadowed an exercise in Ch1. I think handling cookies entirely in `request` is better design, cleaner code, and that there's now no reason to support a `headers` argument.

src/server10.py
pavpanchekha: Yeah, you're right, I'll change this.

book/security.md
```diff
@@ -23,13 +23,18 @@ you're writing security-sensitive code, this book is not enough.
...
+tell whether two HTTP requests come from the same browsers, or
+different ones. Our web browser is effectively anonymous.[^1] That
+means it can't "log in" anywhere, because there's no way for the
+server to know which later requests come from the logged in browser
+and which come from some unrelated browser.
+
+The web fixes this problem with cookies. A cookie---the name is
+meaningless, ignore it---is a little bit of information stored by your
+browser on behalf of a web server. The cookie establishes that
+browser's identity and allows the server to distinguish one web
```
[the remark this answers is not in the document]
pavpanchekha: Let me think more about this, this is a really important point.

book/security.md
```diff
@@ -23,433 +25,324 @@ you're writing security-sensitive code, this book is not enough.
...
-    if 'guest' in params and len(params['guest']) <= 100 and username:
-        ENTRIES.append((params['guest'], username))
+This `do_login` function checks passwords and logs people in by
+storing their user name in the session data:[^timing-attack]
+
+[^timing-attack]: Actually, using `==` to compare passwords like here
+    is a bad idea: Python's equality function for strings scans the
+    string from left to right, and exits as soon as it finds a
+    difference. So, *how long* it takes to check passwords gives you
+    clues about the password; this is called a "timing side channel".
```
chrishtr, line 252:
```suggestion
    clues about the password; this is called a [timing side channel attack](https://en.wikipedia.org/wiki/Timing_attack).
```

book/security.md
```diff
@@ -23,433 +25,324 @@ you're writing security-sensitive code, this book is not enough.
...
+username/password pair.] Of course, this login system has a whole slew
+of insecurities.[^7] But the focus of this book is the browser, not
+the server, so once you're sure it's all working, let's switch gears
+and implement cookies inside our own browser.
 
-Try it out! You should be able to go to the main guest book page,
-click the link to log in, use one of the username/password pairs
-above, and post entries.^[The login flow slows down debugging. You
-might want to add the empty string as a username/password pair.]
+[^7]: The insecurities include not hashing passwords, not using `bcrypt`, not verifying
```
chrishtr, line 278:
```suggestion
[^7]: The insecurities include not hashing passwords, not using [`bcrypt`](https://en.wikipedia.org/wiki/Bcrypt), not verifying
```

book/security.md
```diff
@@ -23,433 +25,324 @@ you're writing security-sensitive code, this book is not enough.
...
+different ones. Our web browser is effectively anonymous.[^1] That
+means it can't "log in" anywhere---after logging in, its requests will
+look just like those of a new visitor.
 
 [^1]: I don't mean anonymous against malicious attackers, who might
     use *browser fingerprinting* or similar techniques to tell users
     apart. I mean anonymous in the good-faith sense.
 
+The web fixes this problem with cookies. A cookie---the name is
+meaningless, ignore it---is a little bit of information stored by your
```
chrishtr, line 39:
```suggestion
meaningless, ignore it^[If you're interested, the history is [here](https://en.wikipedia.org/wiki/HTTP_cookie)]---is a little bit of information stored by your
```

chrishtr (approved):

PR #225, opened 2021-10-04
Rewrite Chapter 10 XSS section

pavpanchekha:
This PR rewrites the XSS portion of Chapter 10 to focus on `Content-Security-Policy` rather than escaping (though escaping is still mentioned!) It depends on #223, but I think the text of the XSS section can be reviewed separately. The CSP implementation is, luckily, pretty slim, because I focus only on the `default-src` policy and only full url sources.

### DIFF
--- a/book/forms.md
+++ b/book/forms.md
@@ -678,7 +678,7 @@ request line:
 
 ``` {.python file=server}
 def handle_connection(conx):
-    req = conx.makefile("rb")
+    req = conx.makefile("b")
     reqline = req.readline().decode('utf8')
     method, url, version = reqline.split(" ", 2)
     assert method in ["GET", "POST"]
--- a/book/forms.md
+++ b/book/forms.md
@@ -838,12 +838,14 @@ def form_decode(body):
     params = {}
     for field in body.split("&"):
         name, value = field.split("=", 1)
-        params[urllib.unquote(name)] = urllib.unquote(value)
+        params[urllib.unquote_plus(name)] = urllib.unquote_plus(value)
     return params
 ```
 
-The `add_entry` function then looks up the `guest` parameter and adds
-its content as a new guest book entry:
+Note that I use `unquote_plus` instead of `unquote`, because some
+browsers use a plus sign to encode a space. The `add_entry` function
+then looks up the `guest` parameter and adds its content as a new
+guest book entry:
 
 ``` {.python file=server}
 def add_entry(params):
--- a/book/security.md
+++ b/book/security.md
@@ -23,13 +23,18 @@ you're writing security-sensitive code, this book is not enough.
 Cookies
 =======
 
-The most basic kind of user data on the web is the *cookie*. A
-cookie---the name is meaningless, ignore it---is a little bit of
-information stored by your browser on behalf of a web server. It's
-that information that allows the server to distinguish one web request
-from another. Bereft of cookies, your web browser is effectively
-anonymous:[^1] it isn't logged in anywhere so it can't do any useful
-personalization or show you your private data.
+With what we've implemented so far there's no way for a web server to
+tell whether two HTTP requests come from the same browsers, or
+different ones. Our web browser is effectively anonymous.[^1] That
+means it can't "log in" anywhere, because there's no way for the
+server to know which later requests come from the logged in browser
+and which come from some unrelated browser.
+
+The web fixes this problem with cookies. A cookie---the name is
+meaningless, ignore it---is a little bit of information stored by your
+browser on behalf of a web server. The cookie establishes that
+browser's identity and allows the server to distinguish one web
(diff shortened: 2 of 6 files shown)


book/forms.md
```diff
@@ -838,12 +838,14 @@ def form_decode(body):
...
         name, value = field.split("=", 1)
-        params[urllib.unquote(name)] = urllib.unquote(value)
+        params[urllib.unquote_plus(name)] = urllib.unquote_plus(value)
     return params
 ```
 
-The `add_entry` function then looks up the `guest` parameter and adds
-its content as a new guest book entry:
+Note that I use `unquote_plus` instead of `unquote`, because some
+browsers use a plus sign to encode a space. The `add_entry` function
```
chrishtr, line 846: not all browsers use a plus sign? Really?

book/security.md
```diff
@@ -38,592 +43,532 @@ personalization or show you your private data.
...
-```
+    <p>Hi! <script src="http://my-server/evil.js"></script>
+    <i> by crashoverride</i></p>
 
-Note that `entry` can be anything, anything the user might stick into
-our comment form. "Anything" might even include a user-written
-`<script>` tag! Let's imagine that you're logged in as one user, maybe
-`crashoverride`, but would like to gain access to another username,
-maybe `nameless`. You could post the comment:
+So, our browser would download and run the `evil.js` script. So
```
chrishtr, line 435:
```suggestion
Our browser would download and run the `evil.js` script. Then
```

book/security.md
```diff
@@ -38,592 +43,532 @@ personalization or show you your private data.
...
+`evil.js` could access `document.cookie` and do something evil; in
+real browser, this might mean using `fetch` to secretly send that
```
chrishtr, line 437:
```suggestion
a real browser, this might mean using `fetch` to secretly send that
```

book/security.md
```diff
@@ -38,592 +43,532 @@ personalization or show you your private data.
...
+    <p>Hi! <script src="http://my-server/evil.js"></script>
+    <i> by crashoverride</i></p>
 
-Note that `entry` can be anything, anything the user might stick into
-our comment form. "Anything" might even include a user-written
-`<script>` tag! Let's imagine that you're logged in as one user, maybe
-`crashoverride`, but would like to gain access to another username,
-maybe `nameless`. You could post the comment:
+So, our browser would download and run the `evil.js` script. So
+`evil.js` could access `document.cookie` and do something evil; in
```
chrishtr, line 436:
```suggestion
`evil.js` could access `document.cookie` and do something evil[^Like impersonating you]; in
```

book/security.md
```diff
@@ -38,592 +43,532 @@ personalization or show you your private data.
...
+real browser, this might mean using `fetch` to secretly send that
+cookie to a server the attacker controls.[^no-fetch]
 
-``` {.example}
-Hi! <script src=http://my-server/evil.js></script>
-```
+[^no-fetch]: Our browser doesn't implement `fetch`, but that doesn't
+    mean this attack won't work. For example, the evil script can
+    replace the whole page with a link that directs their site and
```
chrishtr, line 442:
```suggestion
    replace the whole page with a link that points at their site and
```

book/security.md
```diff
@@ -38,592 +43,532 @@ personalization or show you your private data.
...
+                return True
+        return False
+```
 
-Anyway, whatever it's called, we need to fix it, and the way we do that
-is not sending `entry`, the user comment, directly to the browser.
-Instead, we'd do something to ensure the browser would not interpret it
-as code, something like:
+To test that this works, we can first make our server send a
+`Content-Security-Policy`:
```
chrishtr, line 547:
```suggestion
`Content-Security-Policy` header:
```

book/forms.md
pavpanchekha: I'm not sure. Definitely both plus and `%20` are legal, right?

pavpanchekha: In any case, let's delete `some` without being too specific.

chrishtr: SG

book/security.md
```diff
@@ -38,592 +43,532 @@ personalization or show you your private data.
...
-    return text.replace("&", "&amp;").replace("<", "&lt;")
+``` {.python file=server}
+def handle_connection(conx):
+    # ...
+    response += "Content-Security-Policy: default-src http://localhost:8000\r\n"
+    # ...
+```
 
-def show_comments(username):
+For testing, let's next modify the web pages the server generates to
```
chrishtr, line 556:
```suggestion
For testing, let's next modify the web pages the server to generate a
```

chrishtr (approved):

pavpanchekha:
I'll merge this once #223 is merged.

PR #226, opened 2021-10-04
Fixes and optimizations to compiler

pavpanchekha:
This PR makes the compiler work even if you import constants. This had been broken.

It also commits a major optimization for Font.metrics, at least in Chrome and likely in other browsers.


pavpanchekha:
I have to add: optimizing a web browser implementation that runs in a web browser is quite a fun trick!

chrishtr (approved):

PR #227, opened 2021-10-04
Rewrite Chapter 10 CSRF section

pavpanchekha:
This rewrites the portion of Chapter 10 focused on cross-site request forgery, and re-focuses it on the `SameSite` cookie attribute. The implementation is, as far as I can tell, compliant with the draft RFC and matches what browsers currently do. Note that this PR is dependent on #225 but I think the text can be reviewed independently.


book/security.md
```diff
@@ -38,732 +43,781 @@ personalization or show you your private data.
...
 
-Thanks to the same-origin policy and the mitigations for cross-site
-scripting, it's now hard for an evil-doer to steal the login token
-from our browser. But that doesn't leave us totally safe. Another
-concern is our browser misusing the token. One popular exploit of this
-type is called *cross-site request forgery*, often shortened to CSRF.
+Thanks to `Content-Security-Policy`, web servers can now secure
+themselves against cross-site scripting attacks. But that doesn't
+leave attackers totally out of luck. While they can no longer convince
+web servers to do the wrong things, perhaps they can convince
```
chrishtr, line 578:
```suggestion
web servers to do the wrong things, perhaps they can
```

book/security.md
```diff
@@ -38,732 +43,781 @@ personalization or show you your private data.
...
+
+
+How do we defend against this attack?
+
+Well, the usual advice is to make sure that every POST request to
+`/add` comes from a form on our website. The way to do that is to
+embed a secret value, called a *nonce*, into the form, and to reject
+form submissions that don't come with this secret value. Since you can
+only get a nonce from the server, and since the nonce is tied to your
+user account, the attacker could not embed it in their form and so
```
chrishtr, line 626: s/user account/cookie/ ?

book/security.md
```diff
@@ -38,732 +43,781 @@ personalization or show you your private data.
...
-        out +=   "<p><input name=nonce type=hidden value=" + nonce + "></p>"
-        # ...
+The `SameSite` attribute can take the value `Lax`, `Strict`, or
+`None`, and as I write this browsers have and plan different defaults.
+Our browser will default to `None` and implement `Lax` as an option.
+When `SameSite` is set to `Lax`, the cookie is not sent on cross-site
+`POST` requests, but is sent on same-site `POST` or cross-site `GET`
+requests.[^iow-links]
+
+[^iow-links]: Cross-site `GET` requests are also known as "clicking a
```
chrishtr, line 706: Are you sure this applies to MPA navigations and not just fetches or SPA navigations?

chrishtr: I think it's just that some sites need to support cross-origin GETs for legacy reasons.

book/security.md
```diff
@@ -38,732 +43,781 @@ personalization or show you your private data.
...
-    # ...
-    if username:
-        nonce = str(random.random())[2:]
-        out += "<form action=add method=post>"
-        out +=   "<p><input name=nonce type=hidden value=" + nonce + "></p>"
-        # ...
+The `SameSite` attribute can take the value `Lax`, `Strict`, or
+`None`, and as I write this browsers have and plan different defaults.
+Our browser will default to `None` and implement `Lax` as an option.
+When `SameSite` is set to `Lax`, the cookie is not sent on cross-site
```
chrishtr, line 702: Did you define same site?
https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-05#section-5.2

says:

A request is "same-site" if its target's URI's origin's registrable
   domain is an exact match for the request's client's "site for
   cookies", or if the request has no client.  The request is otherwise
   "cross-site".

We should also note that this behavior corresponds to how cookies have a different definition than the same-origin policy.

book/security.md
```diff
@@ -38,732 +43,781 @@ personalization or show you your private data.
...
-    then use in a CSRF attack for another user's submission.
+Now we can reference the `SameSite` parameter of a cookie. But to
+actually use it, we need to know which site an HTTP request is being
+made from. Let's add a new `top_level` parameter to `request` to track
+that:
 
-``` {.python expected=False}
-def show_comments(username):
+``` {.python}
+def request(url, top_level, payload=None):
```
chrishtr, line 744:
```suggestion
def request(url, top_level_url, payload=None):
```

book/security.md
```diff
@@ -38,732 +43,781 @@ personalization or show you your private data.
...
 
-``` {.python expected=False}
-def add_entry(params, username):
-    if not check_nonce(params, username):
-        return "Invalid nonce", {}
-    # ...
+``` {.python}
+class Tab:
+    def load(self, url, body=None):
+        headers, body = request(url, self.url, payload=body)
```
chrishtr, line 755: This behavior restricts navigations from one url to another to skip cookies...I don't think that is right, as I mentioned above.

src/lab10.py
```diff
@@ -28,8 +28,10 @@
 
 def url_origin(url):
     return "/".join(url.split("/")[:3])
+        
+COOKIE_JAR = {}
 
-def request(url, headers={}, payload=None):
+def request(url, top_level, payload=None):
```
chrishtr, line 34:
```suggestion
def request(url, top_level_url, payload=None):
```

book/security.md
```diff
@@ -38,732 +43,781 @@ personalization or show you your private data.
...
+The idea is that a server can opt into its cookies being `SameSite`,
+which means they're not sent on cross-site form
+submissions.[^in-progress]
 
-We want to make sure that every POST request to `/add` comes from a form
-on our website. The normal way to ensure that is to embed a secret
-value, called a *nonce*, into the form, and to reject form submissions
-that don't come with this secret value.
+[^in-progress]: At the time of this writing, `SameSite` cookies are
+    not standardized, and different browsers handle them differently.
```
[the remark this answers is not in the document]
pavpanchekha: That's true, it's a draft standard but there is a standard. I'll modify the text.

book/security.md
chrishtr: Ok it seems you're right.

pavpanchekha:
I'll merge this once #223 is merged.

PR #232, opened 2021-10-15
Various fixes to the chapter 10 text

chrishtr:


book/security.md
```diff
@@ -142,15 +142,14 @@ You'll also need to modify the argument lists for `add_entry` and
 each user accessing the guest book, we can now build a login system.
 
 ::: {.further}
-The [patent][patent] for cookies says there is "no compelling reason"
```
chrishtr, line 145: I think this might be the wrong patent. Also, I don't like software patents.

book/security.md
```diff
@@ -323,11 +322,8 @@ particular tab. That means that if you're logged in to a website and
 you open a second tab, you're logged in on that tab as well.
 
 When the browser visits a page, it needs to send the cookie for that
-site:[^multi-cookies]
+site:
 
-[^multi-cookies]: Actually, a site can store multiple cookies at once,
```
chrishtr, line 328: Multiple cookies was already mentioned in an earlier paragraph.

book/security.md
```diff
@@ -598,12 +594,6 @@ def url_origin(url):
 Now an attacker can't read the guest book web page. But can they write
 to it? Actually...
 
-::: {.further}
```
chrishtr, line 601: Not filled in, so dump it?

book/security.md
```diff
@@ -848,7 +838,8 @@ To help secure our guest book, we can now mark our guest book cookies
 ``` {.python file=server}
 def handle_connection(conx):
     if 'cookie' not in headers:
-        response += "Set-Cookie: token={}; SameSite=Lax\r\n".format(token)
+        response += \
```
chrishtr, line 841: SameSite appears not be implemented in server10.py. Why?

pavpanchekha:
I pushed all of my changes from a full pass. Besides many small changes I made one big one: I remove the implementation of `document.cookie` and moved it to an exercise. I did this for two reasons:

1. We don't actually use this any more. It used to be necessary to demonstrate an actual XSS attack, but the chapter no longer does that. (Nor should it, since the focus is now elsewhere.)
2. The implementation was incorrect: it didn't add parameters to the `document.cookie` string. This is fixable, but more complicated, and I'm not sure why it's important.

Instead `document.cookie` makes a great exercise, especially when paired with `HttpOnly` to continue our theme of bundling features with security policies.

book/security.md
pavpanchekha: Agreed, but it's still of historic interest, and the only available textual reference I could find on the name.

book/security.md
```diff
@@ -704,12 +700,29 @@ if you forget a form---and relying on every website out there to do it
...
 
+::: {.further}
+One unusual attack, similar in spirit to cross-site request forgery,
+is [click-jacking][clickjacking]. This attack involves including an
+external site in a transparent `iframe` positioned over a
+seemingly-innocent site. The user thinks they are clicking around the
+innocent site, but they actually take actions on the external site.
+Nowadays, sites can prevent this with the [`frame-ancestors`
+directive][csp-frame-ancestors] to `Content-Security-Policy` or the
+older [`X-Frame-Options` header][x-frame-options].
```
chrishtr, line 711: There is also an occlusion detection API as part of IntersectionObserver, targeted at this use case
specifically:

https://web.dev/intersectionobserver-v2/

book/security.md
```diff
@@ -875,48 +896,53 @@ out += "<p>" + entry + " <i>by " + who + "</i></p>"
...
-     these won't work, but the evil script can, for example, replace
-     the whole page with a link that goes to their site and includes
-     the token value in the URL. You've seen "please click to
-     continue" screens and have clicked the button unthinkingly; your
-     users will too.
+which can send[^document-cookie][^cross-domain][^how-send] the cookies
+to the attacker. The attacker could then impersonate other users,
+posting as them or misusing any other capabilities those users had.
+
+[^document-cookie]: A site's cookies and cookie parameters are
```
chrishtr, line 929: Isn't this already clear without a footnote, because the reader implements document.cookie?

book/security.md
```diff
@@ -875,48 +896,53 @@ out += "<p>" + entry + " <i>by " + who + "</i></p>"
...
-[^cross-domain-xhr]: See the exercise on `Access-Control-Allow-Origin`
-    for more details on how web servers can *opt in* to allowing
-    cross-origin requests.
+[mdn-doc-cookie]: https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie
+
+[^cross-domain]: See the exercise on `Access-Control-Allow-Origin` for
+    more details on how web servers can *opt in* to allowing
+    cross-origin requests. To steal cookies, it's the attacker's
+    server that would to opt in to receiving stolen cookies. Or, in a
+    real browser, `evil.js` add images or scripts to the page to
```
chrishtr, line 939:
```suggestion
    real browser, `evil.js` might add images or scripts to the page to
```

book/security.md
pavpanchekha: I guess this is like a more programmable `X-Frame-Options`, if you do in fact need to allow iframe embedded on unknown sites? Do you think we should write about this?

PR #235, opened 2021-10-20
Add an "about the authors" page

chrishtr:
Also: remove the feedback links from non-chapters.


chrishtr:
(this PR is in response to some questions on HN about who we are)

PR #268, opened 2021-11-17
Refactor chapter 11 to be better motivated

chrishtr:
Now the conversation goes like:

* Introduce color and pixels
* Introduce groups and blending trees
* How it's implemented in Skia with SaveLayer, Save, and Restore
* Size and transform
* Opacity
* Blend mode
* Clip-path and rounded corners
* Images
* Fun demo

which I think works out better than before, and has a more clear overall message around the core concept of groups, surfaces and blending.

I added a significant amount of new text (which was, in a different form, there in the original version I wrote last year), but that new text allows me to delete some text from the sections implementing stuff, since it can now just refer back to already-defined concepts.

Still TODO in subsequent PRs:
* Add the demo
* Delete position:relative code
* Simplify code in other ways as suggested in previous discussions (remove rasterizer, make image loading simpler)


pavpanchekha:
Ok, I've read through it and I like the changes. The introductory material on pixels and rasterization is very good. I think we can still make the core lessons (groups, surfaces, blending) stand out more, but I don't think that's a major issue right now with the chapter: the organization is good. Let's merge this.

In terms of next steps, I think the biggest issue with the chapter is to make the implementation parts clearer and shorter where possible. That'll help the explanatory content stand out. I can try to do a pass for that—there are also code changes we've discussed that will help (like loading images in `compute_style` or using compound objects in the display list). Possibly there will be more tweaks to the overall structure once we've done some lower-level fixes, possibly not, but we want to do them anyway.

pavpanchekha (approved):

PR #272, opened 2021-11-25
Remove support for position:relative

chrishtr:


chrishtr:
Just deleting code, so going to go ahead and land this. However, there was some error in lint I couldn't figure out, leading me to put expected=False in one case..any ideas?

PR #273, opened 2021-11-25
Remove the Rasterizer class

chrishtr:
... and inline the methods instead.


chrishtr:
Going to merge this as well, pretty mechanical. PTAL and I can fix any issues in another PR.

PR #274, opened 2021-11-25
Simplify image loading

chrishtr:
Now it just loads it after computing style for an element. Much simpler.


chrishtr:
This one is getting in the way of #276, so going ahead and landing it now.

PR #275, opened 2021-11-27
Add linting and tests for example code

chrishtr:
Chapter 11 has some example python code for the purpose of describing the compositing and blend mode operators.

This PR adds the actual code and tests it. In the process I found a few typos and fixed them.


pavpanchekha (approved):
Good idea

PR #278, opened 2021-11-29
Pass on first section of Chapter 11

pavpanchekha:
I'm doing a heavy pass on Chapter 11, and this PR covers the first section (on installing Skia and SDL, and porting the existing browser code over to it). Besides rephrasing a lot of things to try to make it shorter, I also made some major code changes. My main principles here are to avoid duplication and to keep things the same as last chapter when possible.

- I moved the `SDL_CreateWindow` call into `Browser`, so that it's more similar to last chapter. This requires moving the `SDL_DestroyWindow` call into `Browser` as well, in a new `handle_quit` method. I also split the discussion of the mainloop into multiple code blocks.
- Merge `to_sdl_canvas` with the `SDL_BlitSurface` call and related code. I called the combined method `commit_canvas` but perhaps we should call it `blit` or something else instead. (Chris, you will know the right terminology.) I also moved where in the text this is introduced.
- I made `draw_line`, `draw_text`, and `draw_rect` global helper functions. This allowed `DrawText` and `DrawRect` to call them directly, instead of duplicating their contents. This required passing a `surface` to each draw helper method. (See question 1 below) I also added some foreshadowing for the `Paint` object.
-  I also removed the polyline feature in `draw_polyline`, in favor of just inlining that code where we draw the back button. (See question 2 below.)
- I changed the `DrawRect` and `DrawText` commands not to store their bounds in a Skia `Rect`, just to keep things more similar to the last chapter.
- I split up the part where we port `Browser.draw` into several chunks
- To underscore that ascents are typically negative in Skia, I added a negation when describing how to access it.

Some simplifications I thought we should do, but I didn't quite know if it would work:

1. Instead of passing around the Skia `surface`, can we pass around the Skia `canvas` object? It would reduce the number of `with` blocks and related line noise, and also reduce the number of times we need to change argument names from `canvas` to `surface`. As far as I can see from the documentation and some experiments I ran, this should work fine: every time you get a `Canvas` from a `Surface` you get the same one.
2. Instead of drawing a triangle for a back button, what if we just draw a less than sign `<`? This would let us use the text helper method instead of manually drawing a polyline. We could do this in previous chapters too, which would simplify them too, but for now we can do just this one. We could also use the `DrawXXX` commands to draw the browser chrome; then we wouldn't need the helper methods at all.

Before this PR is ready I'll update the lab code and tests, but I've requested your review before I do that to parallelize.

### DIFF
--- a/book/visual-effects.md
+++ b/book/visual-effects.md
@@ -194,9 +194,26 @@ Now our browser is creating an SDL window can draw to it via Skia. But
 most of the browser codebase is still using Tkinter drawing commands,
 which we now need to replace. Skia is a bit more verbose than Tkinter,
 so let's abstract over some details with helper functions.[^skia-docs]
+First, a helper function to convert colors to Skia colors:
 
-[skia-docs]: Consult the [Skia][skia] and [Skia-Python][skia-python]
+[^skia-docs]: Consult the [Skia][skia] and [Skia-Python][skia-python]
 documentation for more on the Skia API.
+
+``` {.python}
+def color_to_sk_color(color):
+    if color == "white":
+        return skia.ColorWHITE
+    elif color == "lightblue":
+        return skia.ColorSetARGB(0xFF, 0xAD, 0xD8, 0xE6)
+    # ...
+    else:
+        return skia.ColorBLACK
+```
+
+You can add more "elif" blocks to support your favorite color names;
+modern browsers support [quite a lot][css-colors].
+
+[css-colors]: https://developer.mozilla.org/en-US/docs/Web/CSS/color_value
 
 To draw a line, you use Skia's `Path` object:
 
--- a/book/visual-effects.md
+++ b/book/visual-effects.md
@@ -44,7 +44,7 @@ Python bindings like so:
 As elsewhere in this book, you may need to use `pip`, `easy_install`,
 or `python3 -m pip` instead of `pip3` as your installer, or use your
 IDE's package installer. If you're on Linux, you'll need to install
-additional dependencies, like OpenGL, fontconfig. Also, you may not be
+additional dependencies, like OpenGL and fontconfig. Also, you may not be
 able to install `pysdl2-dll`; you'll need to find SDL in your system
 package manager instead. Consult the [`skia-python`][skia-python] and
 [`pysdl2`][sdl-python] web pages for more details.
@@ -77,7 +77,7 @@ if __name__ == "__main__":
     # ...
 ```
 
-Next, we need create an SDL window, instead of a Tkinter window,
+Next, we need to create an SDL window, instead of a Tkinter window,
 inside the Browser, and set up Skia to draw to it. Here's the SDL
 incantation to create a window:
 
@@ -98,7 +98,7 @@ colors). A surface may or may not be bound to the actual pixels on the
 screen via a window, and there can be many surfaces. A *canvas* is an
 API interface that allows you to draw into a surface with higher-level
 commands such as for rectangles or text. Our browser uses separate
-Skia and SDL surface for simplicity, but in a highly optimized
+Skia and SDL surfaces for simplicity, but in a highly optimized
 browser, minimizing the number of surfaces is important for good
 performance.
 
@@ -118,7 +118,7 @@ place to another:
 
 ``` {.python}
 class Browser:
-    def commit_canvas(self):
+    def draw_to_screen(self):
         # Raster the results and copy to the SDL surface:
         skia_image = self.skia_surface.makeImageSnapshot()
         skia_bytes = skia_image.tobytes()
@@ -339,7 +339,7 @@ Once this is done, we need to copy from the Skia surface to the SDL window:
 class Browser:
     def draw(self):
         # ...
-        self.commit_canvas()
+        self.draw_to_screen()
 ```
 
 We've only got a few minor changes left elsewhere in the browser.
@@ -385,7 +385,7 @@ behave just as it did in previous chapters, and it'll probably feel
 faster, because Skia and SDL are faster than Tkinter.
 
 ::: {.further}
-Implement high-quality raster libraries is very interesting in its own
+Implementing high-quality raster libraries is very interesting in its own
 right. These days, it's especially important to leverage GPUs when
 they're available, and browsers often push the envelope. Browser teams
 typically include raster library experts: Skia for Chromium and [Core
--- a/book/visual-effects.md
+++ b/book/visual-effects.md
@@ -89,8 +89,8 @@ class Browser:
             WIDTH, HEIGHT, sdl2.SDL_WINDOW_SHOWN)
 ```
 
-To set up Skia to draw to this window, we need to create two
-"surfaces":[^surface]
+To set up Skia to draw to this window, we also need create a
+"surface" for it:[^surface]
 
 [^surface]: In Skia and SDL, a *surface* is a representation of a
 graphics buffer into which you can draw "pixels" (bits representing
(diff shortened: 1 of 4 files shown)


pavpanchekha:
Another question I forgot:

*Question 3*: In the chapter you call `SDL_GetWindowSurface` in the constructor, and then use it in `commit_canvas`. Does the `SDL_GetWindowSurface` call take a while? If not, let's move it into `commit_canvas`—that way we have fewer fields on `Browser`.

book/visual-effects.md
```diff
@@ -5,312 +5,396 @@ prev: security
...
+    pip3 install skia-python pysdl2 pysdl2-dll
+
+[skia-python]: https://github.com/kyamagu/skia-python
+[sdl-python]: https://pypi.org/project/PySDL2/
+
+::: {.install}
+As elsewhere in this book, you may need to use `pip`, `easy_install`,
+or `python3 -m pip` instead of `pip3` as your installer, or use your
+IDE's package installer. If you're on Linux, you'll need to install
+additional dependencies, like OpenGL, fontconfig. Also, you may not be
```
chrishtr, line 47:
```suggestion
additional dependencies, like OpenGL and fontconfig. Also, you may not be
```

I guess you needed to do this when you tried it?

book/visual-effects.md
```diff
@@ -5,312 +5,396 @@ prev: security
...
+    browser.load(sys.argv[1])
+    # ...
+```
 
-So we need a new library that can implement visual effects. We'll use
-[Skia](https://skia.org), the library that Chromium uses. However, Skia is just
-a library for rastering content on the CPU and GPU, so we'll also use
-[SDL](https://www.libsdl.org/) to provide windows, input events, and OS-level
-integration.
+Next, we need create an SDL window, instead of a Tkinter window,
```
chrishtr, line 80:
```suggestion
Next, we need to create an SDL window, instead of a Tkinter window,
```

book/visual-effects.md
```diff
@@ -5,312 +5,396 @@ prev: security
...
-might need to install additional dependencies. Instructions
-[here](https://pypi.org/project/PySDL2/) explain how to install SDL on your
-computer (short version: `pip3 install PySDL2` and `pip3 install pysdl2-dll`).
+[^surface]: In Skia and SDL, a *surface* is a representation of a
+graphics buffer into which you can draw "pixels" (bits representing
+colors). A surface may or may not be bound to the actual pixels on the
+screen via a window, and there can be many surfaces. A *canvas* is an
+API interface that allows you to draw into a surface with higher-level
+commands such as for rectangles or text. Our browser uses separate
+Skia and SDL surface for simplicity, but in a highly optimized
```
chrishtr, line 101:
```suggestion
Skia and SDL surfaces for simplicity, but in a highly optimized
```

book/visual-effects.md
```diff
@@ -5,312 +5,396 @@ prev: security
...
-types.
+Typically, we'lll draw to the Skia surface, and then once we're done
+with it we'll copy it to the SDL surface to display on the screen.
+This looks a little hairy but really it's just copying pixels from one
+place to another:
 
-The main loop of the browser first needs some boilerplate to get SDL started:
+``` {.python}
+class Browser:
+    def commit_canvas(self):
```
chrishtr, line 121:
```suggestion
    def draw_to_screen(self):
```

"draw" is what Chromium calls the process of taking the rastered pixels and putting them onto the screen
(usually with non-trivial GL programs to do it). But the Skia canvas API has various methods called "draw",
so I suggest "draw_to_screen" as a way to make it clear.

book/visual-effects.md
```diff
@@ -5,312 +5,396 @@ prev: security
...
-        red_mask = 0x00ff0000
-        green_mask = 0x0000ff00
-        blue_mask = 0x000000ff
-        return sdl2.SDL_CreateRGBSurfaceFrom(
-            skia_bytes, WIDTH, HEIGHT, depth, pitch,
-            red_mask, green_mask, blue_mask, alpha_mask)
+class Browser:
+    def draw(self):
+        # ...
+        self.commit_canvas()
```
chrishtr, line 342:
```suggestion
        self.draw_to_screen()
```

src/lab11.py
```diff
@@ -1156,52 +1157,66 @@ def draw(self):
...
         # Draw the back button:
         self.draw_rect(skia.Rect.MakeLTRB(10, 50, 35, 90))
-        draw_polyline(
-            self.skia_surface, 15, 70, 30, 55, 30, 85, True)
+        path = skia.Path().moveTo(15, 70).lineTo(30, 55).lineTo(30, 85)
+        paint = skia.Paint(Color=skia.ColorBlack, Style=skia.Paint.kFill_Style)
+        with surface as canvas:
+            canvas.drawPath(path, paint)
+
+        self.commit_canvas()
```
chrishtr, line 1172:
```suggestion
        self.draw_to_screen()
```

src/lab11.py
```diff
@@ -1156,52 +1157,66 @@ def draw(self):
...
 
+    def commit_canvas(self):
```
chrishtr, line 1174:
```suggestion
    def draw_to_screen(self):
```

book/visual-effects.md
```diff
@@ -5,312 +5,396 @@ prev: security
...
-            paint.setStyle(skia.Paint.kStroke_Style)
-            paint.setStrokeWidth(1);
-            paint.setColor(skia.ColorBLACK)
-        with self.skia_surface as canvas:
-            canvas.drawRect(rect, paint)
+To draw text, you use `drawString`:
+
+``` {.python}
+def draw_text(surface, x, y, text, font, color=None):
+    sk_color = color_to_sk_color(color)
```
chrishtr, line 219: Looks like I forgot to show the code for color_to_sk_color...

book/visual-effects.md
```diff
@@ -5,312 +5,396 @@ prev: security
...
+        is self.focus:
+            # ...
             x = obj.x + obj.font.measureText(text)
+            y = obj.y - self.scroll + CHROME_PX
+            draw_line(surface, x, y, x, y + obj.height)
 ```
 
-Update all the other places that `measure` was called to use the Skia method
-(and also create Skia fonts instead of Tkinter ones, of course).
+Note that I've renamed `draw`'s argument from `surface` to `canvas`
```
chrishtr, line 359: You didn't do the rename....

In any case, to answer question 1, yes you can pass the canvas around. I think it's not a good idea
to keep it bound in a variable, but keeping it around for the duration of Browser.draw is a good
simplification.

book/visual-effects.md
```diff
@@ -5,312 +5,396 @@ prev: security
...
+
+You should now be able to run the browser again. It should look and
+behave just as it did in previous chapters, and it'll probably feel
+faster, because Skia and SDL are faster than Tkinter.
 
-Now you should be able to run the browser just as it did in previous chapters,
-and have all of the same visuals. It'll probably also feel faster, because
-Skia and SDL are highly optimized libraries written in C & C++.
+::: {.further}
+Implement high-quality raster libraries is very interesting in its own
```
chrishtr, line 388:
```suggestion
Implementing high-quality raster libraries is very interesting in its own
```

chrishtr (commented):
Overall: I like these changes and they are a good simplification.

Answer to question 1 inline, but yes I agree it's better to do a single with statement in Browser.draw and pass around the canvas.

Answer to question 2: it only saves 2 or 3 lines of code, and might interest the readers
who want to get hints at the fancier features of SKia...so I'd just leave it.

chrishtr:
Answer to question 3: looks like you're right that SDL_GetWindowSurface is fast. The surface is only invalidated if the window changes size. So it can be inlined.

https://wiki.libsdl.org/SDL_GetWindowSurface

book/visual-effects.md
pavpanchekha: Good point, I'll add it.

book/visual-effects.md
pavpanchekha: Ok, will do

pavpanchekha:
Ok, I think I've addressed all the feedback so far.

chrishtr (approved):

chrishtr:
Approved

PR #279, opened 2021-11-30
Fix event handling map key in chapter 10

chrishtr:
The key needs to be consistently using id() or not. I'm not sure why lab10 uses id() but lab9 and earlier
do not though. Do you remember?

This CL also fixes the innerHTML method to be called innerHTML_set, like runtime10.js requires.

### DIFF
--- a/src/lab10.py
+++ b/src/lab10.py
@@ -127,7 +127,7 @@ class JSContext:
         self.interp.evaljs(code)
 
     def dispatch_event(self, type, elt):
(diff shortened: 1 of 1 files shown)


pavpanchekha:
This looks like a bug of some kind. At first I used id() because I thought there was some issue with using the element directly as a key, but there wasn't. I think I had a bad merge or something? (That is—I started work on lab10 before we had fixed all the bugs in lab9???)

pavpanchekha (approved):

PR #229, opened 2021-10-08
Add code to correct for browser text alignment differences

pavpanchekha:
This PR fixes the annoying fact that `create_text` draws the text with different offsets on different browsers. The fix pre-draws a capital F on a canvas and measures its offset.

PR #282, opened 2021-12-02
Use a web font to align font metrics between browsers

chrishtr:
Using "serif" is not specific enough to ensure two browsers choose the same font, even on the same OS. Likewise, different OSes have different built-in fonts. The solution is to use a web font.

This problem was apparently the error in this blog post.
https://pqina.nl/blog/cross-browser-alignment-of-the-canvas-filltext-draw-call/

This PR also gets rid of a Chrome hack that makes it render the font differently than Firefox. I'm not sure, but in Chrome
96 there was a rewrite of SVG text that shipped, and it may have rendered this hack obsolete. (I have a question out to the team about this.)

This PR makes #229 unnecessary.

### DIFF
--- a/www/widgets/lab3-baselines.html
+++ b/www/widgets/lab3-baselines.html
@@ -2,7 +2,11 @@
 <meta charset="utf-8" />
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
 <link rel="stylesheet" href="widget.css" />
-
+<link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Merriweather:wght@300&display=swap">
+<script>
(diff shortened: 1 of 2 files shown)


pavpanchekha:
Nice, I approve this fix. Can we use this font in the UI chrome in `lab7.js`? In that lab we might want to adjust some fudge factors (like in #229) to make the plus sign in the UI aligned. (I'm starting to think that we should have just used two lines instead of a plus sign—less trouble!)

pavpanchekha (approved):

pavpanchekha:
Honestly don't know why I didn't think about "serif" being resolved differently by different browsers. Of course!

chrishtr:
Re lab7: sure..

PR #284, opened 2021-12-04
Remove non-essential features

chrishtr:
This PR moves transform and size to exercises. They increase the length of the chapter, and don't add enough value to the core concepts.

It also removes background images, since they are unconnected to the core concepts. I think it would make sense to put it in a future chapter on replaced elements: images, video and iframes.

### DIFF
--- a/book/visual-effects.md
+++ b/book/visual-effects.md
@@ -575,8 +575,10 @@ complicated to handle in real browsers.
 
 [containing-block]: https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block
 
-Surfaces and canvases
+Surfaces and Canvases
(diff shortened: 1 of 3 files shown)


book/visual-effects.md
```diff
@@ -578,6 +578,8 @@ complicated to handle in real browsers.
 Surfaces and canvases
 =====================
 
+<a name="surfaces-canvases"></a>
```
pavpanchekha, line 581: Every heading in pandoc has an auto-generated name. This one is `#surfaces-and-canvases`. So we don't need to add these anchors to the book.

pavpanchekha (approved):

PR #286, opened 2021-12-06
Browser compositor

chrishtr:
This PR adds:

* A new section about browser compositing, with how to use it to separate browser chrome from web contents
* Accompanying implementation that uses three surfaces, and stops baking scroll offsets and browser chrome offsets into the display lists


chrishtr:
Ready for review.

pavpanchekha (approved):
This is great. I think after I finish my pass over the other sections, we will do a little editing so that we don't have to do multiple renames in the same chapter (like for draw_to_screen vs draw).

Let's finish editing everything (blocked on me :( ) and then see how the length is. If we feel like it's too long we don't need to do this, but you could imagine adding animations + rotate or something like that, which would require a proper tree of surfaces.

PR #288, opened 2021-12-08
Big pass on opacity, blending, and clipping

pavpanchekha:
This PR mostly edits three sections of Chapter 11, on opacity, blending, and clipping. There are some big changes implemented:

- I inlined `paint_background` and removed the `paint_x` and `paint_y` variables, which were important abstractions when we had positioning and background images but aren't any more.
- All the examples are updated to use parent-child relationships to create overlap instead of transforms or positioning.
- The `SaveLayer` drawing command now forms an explicit tree, instead of having a paired `Restore` drawing command. This let me remove the restore count, and also unify `paint_visual_effects` with `paint_clip_path`, since the new `SaveLayer` drawing command lets us add Skia commands both before or after the element contents.
- I moved rounded rectangles to their own section called "Optimizing surface use". To make that section more interesting, I also purposefully implemented opacity and blend modes to each add their own surface, which can then be removed by future text in this section.
- I switched circular clip-paths to use pixel radii; without `width` and `height` it's hard to create squares, so percentage clip paths look weird. The new `SaveLayer` structure also let me do circular clips in terms of an explicit `DrawCircle` command, which I think is clearer.
- The theoretical discussion of alpha is moved right next to its implementation, which I think is more natural and avoids defining special-purpose anchors.
- I redid the pedagogical code snippets for blend modes in terms of an imaginary `Pixel` class.

Besides this PR, I think the chapter still needs 1) a pass over the new scrolling section; 2) a pass over the new "optimizing surface use" section; and 3) a pass over the "surfaces and canvases" section, and possibly its merger into the next "compositing with SaveLayer" section. Besides those, there are some points of general cleanup (for example, we need to explain that we are now assigning a `rect` field to each drawing command, and also we need to explain how we are parsing colors).

### DIFF
--- a/book/visual-effects.md
+++ b/book/visual-effects.md
@@ -466,7 +466,7 @@ Note that Skia supports `RRect`s, or rounded rectangles, natively, so
 we can just draw one right to a canvas. Now we can draw these rounded
 rectangles for the background:
 
-``` {.python}
+``` {.python replace=display_list./cmds.}
 class BlockLayout:
     def paint(self, display_list):
         if bgcolor != "transparent":
@@ -1089,8 +1089,10 @@ large enough that the border radius is obvious.
 </div>
 
 Look at how the letters near the corner are cut off to maintain a
-sharp rounded edge. That's clipping; without the `overflow: clip`
-property these letters would instead be fully drawn.
+sharp rounded edge. (Uhh... actually, at the time of this writing,
+Safari does not support `overflow: clip`, so if you're using Safari
+you won't see this effect.) That's clipping; without the `overflow:
+clip` property these letters would instead be fully drawn.
 
 Counterintuitively, we'll implement clipping using blending modes.
 We'll make a new surface (the mask), draw a rounded rectangle into it,
--- a/book/visual-effects.md
+++ b/book/visual-effects.md
@@ -63,8 +63,8 @@ If any of these imports fail, you probably need to check that Skia and
 SDL were installed correctly. Note that the `ctypes` module comes
 standard in Python; it is used to convert between Python and C types.
 
-SDL replaces Tkinter
-====================
+SDL creates the Window
+======================
 
 The main loop of the browser first needs some boilerplate to get SDL
 started:
@@ -193,8 +193,8 @@ constructor; this main loop replaces them. Also note that I've changed
 the signatures of the various `handle_xxx` methods; you'll need to
 make analogous changes in `Browser` where they are defined.
 
-Skia replaces Tkinter
-=====================
+Skia is the canvas
+==================
 
 Now our browser is creating an SDL window can draw to it via Skia. But
 most of the browser codebase is still using Tkinter drawing commands,
@@ -381,6 +381,34 @@ class Tab:
             draw_line(canvas, x, y, x, y + obj.height)
 ```
 
+This takes care of just about all of our graphics commands, with just
+one small thing left: text handling.
+
+::: {.further}
+Implementing high-quality raster libraries is very interesting in its own
+right. These days, it's especially important to leverage GPUs when
+they're available, and browsers often push the envelope. Browser teams
+typically include raster library experts: Skia for Chromium and [Core
+Graphics][core-graphics] for Webkit, for example. Both of these
+libraries are used outside of the browser, too: Core Graphics in iOS
+and macOS, and Skia in Android.
+
+If you're looking for another book on how these libraries are implemented, check
+out [Real-Time Rendering][rtr-book]. There is also [Computer Graphics:
+Principles and Practice][classic], which incidentally I remember buying back
+the days of my youth (1992 or so). At the time I didn't get much further than
+rastering lines and a few polygons (and in assembly language!). These days you
+can do the same and more with Skia and a few lines of Python.
+:::
+
+[core-graphics]: https://developer.apple.com/documentation/coregraphics
+[rtr-book]: https://www.realtimerendering.com/
+[classic]: https://en.wikipedia.org/wiki/Computer_Graphics:_Principles_and_Practice
+
+
+Skia is also the font library
+=============================
+
 Note that I've also replaced Tkinter's `measure` call with Skia's
 `measureText` equivalent. We'll also need to create Skia fonts instead
 of Tkinter ones. Update all the other places that `measure` was called
@@ -406,26 +434,49 @@ You should now be able to run the browser again. It should look and
 behave just as it did in previous chapters, and it'll probably feel
 faster, because Skia and SDL are faster than Tkinter.
 
-::: {.further}
-Implementing high-quality raster libraries is very interesting in its own
-right. These days, it's especially important to leverage GPUs when
-they're available, and browsers often push the envelope. Browser teams
-typically include raster library experts: Skia for Chromium and [Core
-Graphics][core-graphics] for Webkit, for example. Both of these
-libraries are used outside of the browser, too: Core Graphics in iOS
-and macOS, and Skia in Android.
+Let's reward ourselves for the big refactor with a simple feature that
+Skia enables: rounded corners via the `border-radius` CSS property,
+like this:
 
-If you're looking for another book on how these libraries are implemented, check
-out [Real-Time Rendering][rtr-book]. There is also [Computer Graphics:
-Principles and Practice][classic], which incidentally I remember buying back
-the days of my youth (1992 or so). At the time I didn't get much further than
-rastering lines and a few polygons (and in assembly language!). These days you
-can do the same and more with Skia and a few lines of Python.
-:::
+    <div style="border-radius: 10px">
+        This is some example text.
+    </div>
 
-[core-graphics]: https://developer.apple.com/documentation/coregraphics
-[rtr-book]: https://www.realtimerendering.com/
-[classic]: https://en.wikipedia.org/wiki/Computer_Graphics:_Principles_and_Practice
+Which looks like this:
+
+<div style="border-radius: 10px">
+This is some example text.
+</div>
+
+Implementing requires drawing a rounded rectangle, so let's add a new
+`DrawRRect` command:
+
+``` {.python}
+class DrawRRect:
+    def __init__(self, rect, radius, color):
+        self.rect = rect
+        self.rrect = skia.RRect.MakeRectXY(rect, self.radius, self.radius)
+        self.color = color
+
+    def execute(self, canvas):
+        canvas.drawRRect(self.rrect, paint=skia.Paint(Color=self.color))
+```
+
+Note that Skia supports `RRect`s, or rounded rectangles, natively, so
+we can just draw one right to a canvas. Now we can draw these rounded
+rectangles for the background:
+
+``` {.python}
+class BlockLayout:
+    def paint(self, display_list):
+        if bgcolor != "transparent":
+            radius = float(self.node.style.get("border-radius", "0px")[:-2])
+            display_list.append(DrawRRect(rect, radius, bgcolor))
+```
+
+In this way, one advantage of using Skia is that, since it is also
+used in the Chrome browser, we know it has fast, built-in support for
+all of the shapes we might need.
 
 Pixels, Color, Raster
 =====================
@@ -646,7 +697,7 @@ way each time.
 ``` {.html.example}
 <div style="background-color:orange">
     Parent
-    <div style="background-color:white">Child</div>
+    <div style="background-color:white;border-radius:5px">Child</div>
     Parent
 </div>
 ```
@@ -655,7 +706,7 @@ It looks like this:
 
 <div style="background-color:orange">
 Parent
-<div style="background-color:white">Child</div>
+<div style="background-color:white;border-radius:5px">Child</div>
 Parent
 </div>
 
@@ -670,7 +721,7 @@ the white child element, it looks like this:
 
 <div style="background-color:orange">
 Parent
-<div style="background-color:white;opacity:0.5">Child</div>
+<div style="background-color:white;border-radius:5px;opacity:0.5">Child</div>
 Parent
 </div>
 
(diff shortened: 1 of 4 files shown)


pavpanchekha:
Like with the last PR, I'll work on passing the lints and tests but wanted to get the text up as soon as I could for you to read and give feedback on.

book/visual-effects.md
```diff
@@ -648,338 +631,305 @@ of thumb is: if you don't need a non-default blend mode, then you can use
...
-Right now, the blue rectangle completely obscures part of the orange one, and
-the orange one does the same to the light blue one. In order to explore
-what happens with blending further, let's make the elements
-semi-transparent.
+Right now, the white rectangle completely obscures part of the orange
+one; The two colors don't really need to "mix", and in fact it kind of
+looks like two orange rectangles instead of an orange rectangle with a
+white one on top. Now let's make the white child element
+semi-transparent, so the colors have to mix. In CSS, that requires
+adding an `opacity` property with a value somewhere in between 0
```
chrishtr, line 667:
```suggestion
adding an `opacity` property with a value somewhere between 0
```

book/visual-effects.md
```diff
@@ -648,338 +631,305 @@ of thumb is: if you don't need a non-default blend mode, then you can use
...
+(completely transparent) to 1 (totally opaque). With 50% opacity on
```
chrishtr, line 668:
```suggestion
(completely transparent) and 1 (totally opaque). With 50% opacity on
```

book/visual-effects.md
```diff
@@ -648,338 +631,305 @@ of thumb is: if you don't need a non-default blend mode, then you can use
...
+==================
 
-Next, compositing: let's also assume that the coordinate spaces and pixel
-densities of the two surfaces are the same, and therefore their pixels overlap
-and have a 1:1 relationship. Therefore we can "composite" each pixel in the
-popped surface on top of (into, really) its corresponding pixel in the restored
-surface. Let's call the popped surface the *source surface* and the restored
-surface the *backdrop surface*. Likewise each pixel in the popped surface is
-a *source pixel* and each pixel in the restored surface is a *backdrop pixel*.
+Let's pause and explore how opacity actually works under the hood.
```
chrishtr, line 770:
```suggestion
Now let's pause and explore how opacity actually works under the hood.
```

book/visual-effects.md
```diff
@@ -648,338 +631,305 @@ of thumb is: if you don't need a non-default blend mode, then you can use
...
-def restore(source_surface, backdrop_surface, width, height, opacity):
-    for x in range(0, width):
-        for y in range(0, height):
-            backdrop_surface.setPixel(
-                x, y,
-                composite(
-                    apply_opacity(source_surface.getPixel(x, y)),
-                    backdrop_surface.getPixel(x, y)))
-```
+That `Alphaf` operation applies to pixel in one surface. But with
```
chrishtr, line 819:
```suggestion
That `Alphaf` operation applies to pixels in one surface. But with
```

book/visual-effects.md
```diff
@@ -648,338 +631,305 @@ of thumb is: if you don't need a non-default blend mode, then you can use
...
+`SaveLayer` we will end up with two surfaces, with all of their pixels
+aligned, and therefore we will need to combined, or *blend* pairs of
```
chrishtr, line 821:
```suggestion
aligned, and therefore we will need to combine, or *blend* pairs of
```

book/visual-effects.md
```diff
@@ -648,338 +631,305 @@ of thumb is: if you don't need a non-default blend mode, then you can use
...
+corresponding pixels.
+
+Here the terminology can get confusing: we imagine that the pixels "on
+top" are blending into the pixels "below", so we call the top surface
+the *source surface*, with source pixels, and the bottom surface the
+*destination surface*, with destination pixels. When we combine them,
```
chrishtr, line 827: I called it backdrop to conform to the spec. If destination is more clear, we can use that. But backdrop
does have an intuitive meaning maybe?

book/visual-effects.md
```diff
@@ -648,338 +631,305 @@ of thumb is: if you don't need a non-default blend mode, then you can use
...
 
-[^real-life-reading-pixels]: In real browsers it's a bad idea to read canvas
-pixels into memory and manipulate them like this, because it would be very
-slow. Instead, it should be done on the GPU. So libraries such as Skia don't
-make it convenient to do so. (Skia canvases do have `peekPixels` and
-`readPixels` methods that are sometimes used, but not for this use case).
+[^simple-alpha]: The formula for this code can be found
+[here](https://www.w3.org/TR/SVG11/masking.html#SimpleAlphaBlending).
+Note that that page refers to premultiplied alpha colors, but Skia
+does not use premultiplied color representations internally.
```
chrishtr, line 835:
```suggestion
does not use premultiplied color representations internally, and the code below doesn't either.
```

book/visual-effects.md
```diff
@@ -648,338 +631,305 @@ of thumb is: if you don't need a non-default blend mode, then you can use
...
+        self.b = self.b * (1 - source.a) * self.a + source.b * source.a
+        self.a = 1 - (1 - source.a) * (1 - self.a))
+```
 
-[^vs-blending-2]: Again, see [here](#compositing-blending) for compositing vs
-blending.
+Here the destination pixel `self` is modified to blend in the source
+pixel `source`. The mathematical expressions for the red, green, and
+blue color channels are identical, and basically average the source
+and destination colors, weighted by alpha.[^source-over-example] You
```
chrishtr, line 850: Footnote missing?

book/visual-effects.md
```diff
@@ -990,315 +940,181 @@ def parse_blend_mode(blend_mode_str):
...
-over the first one, and then using scissors to cut along the edge of the
-mask.
+The "multiply" and "difference" blend modes can seem kind of obscure,
+but blend modes are a flexible way to implement per-pixel operations.
+One common use case is clipping---erasing the pixels in a surface that
+don't intersect with a given shape. It's called clipping because it's
+like putting a second piece of paper (called a *mask*) over the first
+one, and then using scissors to cut along the mask's edge.
+
+Clipping can be used via the CSS `clip-path` property. The [full
```
chrishtr, line 989:
```suggestion
One way of expressing clipping on the web is via the CSS `clip-path` property. The [full
```

book/visual-effects.md
```diff
@@ -990,315 +940,181 @@ def parse_blend_mode(blend_mode_str):
...
-Let's apply a circular mask:
+<div style="clip-path:circle(20px);background-color:lightblue">
+This test text exists here to ensure that the "div" element is
+large enough that a reasonably large circle is drawn to your screen.
+</div>
 
-    <div style="width:256px; height:256px;
-        clip-path:circle(50%);background-color:lightblue">
-    </div>
+Here, the percentage defines the radius of the circle, and is
```
chrishtr, line 1008: Remove this paragraph.

book/visual-effects.md
```diff
@@ -1503,6 +1319,138 @@ are almost endless.
...
+While the `mask` CSS property is relatively uncommon used (as is `clip-path`
+actually), there is a special kind of mask that is very common: rounded
+corners. Now that we know how to implement masks, this one is also easy to
+add to our browser (draw a mask image with the `drawRRect` Skia canvas method).
+But because it's so common, Skia provides a special-purpose
+method to clip to rounded corners: `clipRRect`.
+
+Rounded corners are specified in CSS via `border-radius`. Here's an example:
+
+    <div style="width:256px; height:256px;
```
chrishtr, line 1334: Change to not use width or height, but instead text, to make it work in our browser?

src/lab11.py
```diff
@@ -428,54 +408,38 @@ def style_length(node, style_name, default_value):
...
-            display_list.append(CircleMask(
-                center_x, center_y, radius, rect))
-
-def paint_visual_effects(node, display_list, rect):
-    restore_count = 0
-    
+def paint_visual_effects(node, cmds, rect):
+    opacity = float(node.style.get("opacity", "1.0"))
+    paint = skia.Paint(Alphaf=opacity)
+    cmds = [SaveLayer(paint, cmds)]
```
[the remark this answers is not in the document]
pavpanchekha: They probably do fail. I'm not sure I understand the bug, so let's discuss Friday so I understand it correctly.

(I initially had this code with opacity + blending working identically to `master`, but redid it with multiple `SaveLayer` calls to then have a discussion later about removing them. If it's incorrect and unsalvagable I'll unwind this change, it's not that important.)

src/lab11.py
```diff
@@ -428,54 +408,38 @@ def style_length(node, style_name, default_value):
...
+    if circle_radius:
+        width = rect.right() - rect.left()
+        height = rect.bottom() - rect.top()
+        center_x = rect.left() + width / 2
+        center_y = rect.top() + height / 2
+
+        mask_cmds = [DrawCircle(center_x, center_y, circle_radius, skia.ColorWHITE)]
+        paint = skia.Paint(BlendMode=skia.kDstIn)
+        cmds.append(SaveLayer(paint, mask_cmds))
+        cmds = [SaveLayer(skia.Paint(), cmds)]
```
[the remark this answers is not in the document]
pavpanchekha: The point of line 433 is to isolate the element content from the rest of the page so only the element content is clipped. So for example if we have a paragraph with a background that is clipped into a circle, we want its drawing commands to be:

    SaveLayer([DrawRect, DrawText, DrawText, SaveLayer(DstIn, [DrawCircle])])

I might have this wrong, though—I was attempting to do a pure refactor here of your code and probably got it wrong. Definitely we want to fix it before merging, let's discuss Friday.

src/lab11.py
```diff
@@ -428,54 +408,38 @@ def style_length(node, style_name, default_value):
...
-                    height * height) / math.sqrt(2)
-            radius = reference_val * percent / 100
-            (center_x, center_y) = center_point(rect)
-            display_list.append(CircleMask(
-                center_x, center_y, radius, rect))
-
-def paint_visual_effects(node, display_list, rect):
-    restore_count = 0
-    
+def paint_visual_effects(node, cmds, rect):
```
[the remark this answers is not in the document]
pavpanchekha: Sounds good, let's discuss.

book/visual-effects.md
pavpanchekha: Yeah I could go either way here. "Destination" is nice because then "destination-in" makes a lot of sense. But "backdrop" is nice because it reminds you which one is the backdrop. We could also introduce both terms, but that itself can be confusing.

book/visual-effects.md
pavpanchekha: Yes. I haven't passed over this section yet but definitely switch to text.

pavpanchekha:
I believe I now have the lints passing, and the only failing tests are failing because my new code executes a different series of `saveLayer`/`restore` calls—which, let's discuss whether this PR has bugs before we update the tests :)

src/lab11.py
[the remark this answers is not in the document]
chrishtr: When you do a dst-in blend, you have to first paint the backdrop, then paint the stuff that has dst-in on it. This is because when you call restore(), that means "immediately blend with the existing stuff", but not stuff that comes after.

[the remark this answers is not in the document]
pavpanchekha: I think this code does that. It creates a new layer for the element contents, draws the element contents, then creates a new layer for the circle, draws and restores it (thus clipping the element contents) and finally draws the clipped element contents into the page. I've confirmed that if I undo changes to the handling of opacity / blending (discussion above) this generates the same Skia commands as the old code.

pavpanchekha:
In keeping with "first do it correctly, then optimize" maybe instead implementing `clip-path: circle()` we should implement rounded rectangles with a mask?

pavpanchekha:
Actually, I played a bit with `border-radius` and it looks like it is only supposed to clip the background. MDN says:

> The radius applies to the whole background, even if the element has no border; the exact position of the clipping is defined by the background-clip property.

The spec (https://drafts.csswg.org/css-backgrounds/#corner-clipping) is not super clear to me, but it does seem that all browsers behave identically here. I think this is a problem with the existing implementation of border radius as well.

We could just note the difference and plead simplicity, or do something else—what do you think?

chrishtr:
Hmm. Yes it's true that just border-radius will only apply to self and not to descendants. I forgot about that. You need `overflow:clip` also to get the desired behavior. How about we implement both of them, except skip the "only border radius" case on purpose?

if (rounded && clip) {
 clipRRect
} else if (clip) {
  clipRect
} else {
 // do nothing
}

chrishtr:
And leave the "do nothing" case to an exercise, as a way of exploring effects only applying to some commands for self.

pavpanchekha:
Interesting idea, this also works well to demonstrate that clipping is a generic primitive.

chrishtr:
However, note that overflow:clip only applies to descendants, not self... so to make it simpler we could only implement the combo?

pavpanchekha:
Ok, I'll write it as focusing just on the combination of overflow: clip with border-radius. I'm going for a walk but will try to finish this tonight.

book/visual-effects.md
```diff
@@ -648,338 +631,309 @@ of thumb is: if you don't need a non-default blend mode, then you can use
...
-Next, compositing: let's also assume that the coordinate spaces and pixel
-densities of the two surfaces are the same, and therefore their pixels overlap
-and have a 1:1 relationship. Therefore we can "composite" each pixel in the
-popped surface on top of (into, really) its corresponding pixel in the restored
-surface. Let's call the popped surface the *source surface* and the restored
-surface the *backdrop surface*. Likewise each pixel in the popped surface is
-a *source pixel* and each pixel in the restored surface is a *backdrop pixel*.
+Note that `paint_visual_effects` receives a list of commands and
+returns another list of commands. It's just that the output list is
+always a single `SaveLayer` command.
```
chrishtr, line 817:
```suggestion
always a single `SaveLayer` command that wraps the original content---which makes sense, because
first we need to draw the commands to a surface, and *then* apply transparency to it when blending into
the parent.
```

book/visual-effects.md
chrishtr: ok let's go with destination because of dst-in.

book/visual-effects.md
```diff
@@ -989,316 +943,152 @@ def parse_blend_mode(blend_mode_str):
...
+_then_ blend the result into the rest of the page. If we switched the
+two `SaveLayer` calls, so that we first applied blending, there
+wouldn't be anything to blend it into!
+
 [mbm-mult]: https://drafts.fxtf.org/compositing-1/#blendingmultiply
 [mbm-diff]: https://drafts.fxtf.org/compositing-1/#blendingdifference
 
-[mix-blend-mode-def]: https://drafts.fxtf.org/compositing-1/#propdef-mix-blend-mode
+::: {.further}
+In reality, reading individual pixels into memory and manipulate them
```
chrishtr, line 1023:
```suggestion
In reality, reading individual pixels into memory to manipulate them
```

book/visual-effects.md
```diff
@@ -989,316 +943,152 @@ def parse_blend_mode(blend_mode_str):
...
+def paint_visual_effects(node, cmds, rect):
     # ...
-
-    clip_path = node.style.get("clip-path")
-    if clip_path:
-        display_list.append(SaveLayer(skia.Paint(), rect))
-        restore_count = restore_count + 1
+    border_radius = float(node.style.get("border-radius", "0px")[:-2])
+
+    return [
```
chrishtr, line 1133: \o/

book/visual-effects.md
```diff
@@ -989,316 +996,169 @@ def parse_blend_mode(blend_mode_str):
...
-            height = rect.bottom() - rect.top()
-            reference_val = \
-                math.sqrt(width * width +
-                    height * height) / math.sqrt(2)
-            radius = reference_val * percent / 100
-            (center_x, center_y) = center_point(rect)
-            display_list.append(CircleMask(
-                center_x, center_y, radius, rect))
-```
+<div style="border-radius:30px;background-color:lightblue;overflow:clip">
```
chrishtr, line 1088: We can add overflow:hidden as a hack. Also, I happen to know that the implementation is complete and it'll ship early next year...

book/visual-effects.md
```diff
@@ -406,26 +434,49 @@ You should now be able to run the browser again. It should look and
...
-[core-graphics]: https://developer.apple.com/documentation/coregraphics
-[rtr-book]: https://www.realtimerendering.com/
-[classic]: https://en.wikipedia.org/wiki/Computer_Graphics:_Principles_and_Practice
+Which looks like this:
+
+<div style="border-radius: 10px">
+This is some example text.
+</div>
+
+Implementing requires drawing a rounded rectangle, so let's add a new
```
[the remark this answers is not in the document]
chrishtr: I'll resolve this later.

book/visual-effects.md
chrishtr: I'll resolve this later.

PR #292, opened 2021-12-11
Fix lint and tests

chrishtr:
* Fixes some bugs
* Implements optimizations to avoid surfaces and drawRRect unless necessary

The optimizations are embedded within the SaveLayer class, to make the code still easy to read.


chrishtr:
Going to go ahead and commit, please review async.

PR #2, opened 2019-12-16
Add link to this repo

BO41:
This makes it easier for people to find this repo to submit typo fixes

PR #293, opened 2021-12-12
Add section on optimizing surfaces

chrishtr:
Added text:

* How to avoid surfaces with some simple boolean logic
* Why and how to replace destination-in compositing with clipRRect
* Newlines to avoid horizontal scrolling
* Some miscellaneous editorial/typo fixes

New code:

* clipRRect


pavpanchekha:
Nice.

A couple of thoughts—let's resolve these here in the comments to move faster toward publication:

1. What do you think of putting this before the scrolling section? Then we introduce drawing parameters / `clipRRect` here and then when we use `clipRect` later the reader will know that as an optimization and can focus on scrolling & multiple surfaces.
2. I don't think this does the maximum level of optimization. If you have both opacity and mix-blend-mode, we will create two layers instead of one. This could be a good place to reprise the "order matters" point from earlier.
3. Part of the above—adding booleans to skip layer creation is nice but keeps the tree structure the same. I'm not sure we can do optimizations like #2 with that constraint. The alternative is to have a bunch of control flow when creating the tree. Downside is the code is no longer as readable at the end, but we do keep the one big tree in the beginning.
4. How cheap is `Save`? I assume basically free? Why not move the `save` inside the `ClipRRect` command? That would make the display "list" (actually a tree now) basically stateless.

pavpanchekha (commented):
See comment in convo

pavpanchekha:
For #3 above, by the way, my proposal is to make the code look like your original `paint_visual_effects`, which had a bunch of control flow before creating the layers.

chrishtr:
> * What do you think of putting this before the scrolling section? Then we introduce drawing parameters / `clipRRect` here and then when we use `clipRect` later the reader will know that as an optimization and can focus on scrolling & multiple surfaces.

Good idea. DOne.

> * I don't think this does the maximum level of optimization. If you have both opacity and mix-blend-mode, we will create two layers instead of one. This could be a good place to reprise the "order matters" point from earlier.

Yes, but those are necessary in general, unless you want to mix the math up? Are you also suggesting I point out how order matters?

> * Part of the above—adding booleans to skip layer creation is nice but keeps the tree structure the same. I'm not sure we can do optimizations like [Add link to this repo #2](https://github.com/browserengineering/book/pull/2) with that constraint.

I think your hyperlink didn't work out...

> The alternative is to have a bunch of control flow when creating the tree. Downside is the code is no longer as readable at the end, but we do keep the one big tree in the beginning.

I like that it's minimal code differences in the current code, but gets across the point that you can skip surfaces, and in the end the logic is pretty simple. We're not implementing a real browser after all, where even a "new" operator is to be avoided.

> * How cheap is `Save`? I assume basically free?

Maybe, but easy to avoid in a similar pattern, so ok?

Why not move the `save` inside the `ClipRRect` command? That would make the display "list" (actually a tree now) basically stateless.

I think whatever is easier for students to understand and debug, while also being correct, is best.

I left it separate by analogy with SaveLayer. I could do a SaveAndClip display item.
I"m not sure what you mean by making it stateless...can you explain?

pavpanchekha:
> Yes, but those are necessary in general, unless you want to mix the math up? Are you also suggesting I point out how order matters?

I would say best to point this out; otherwise readers might notice the missing opportunity and not know why it's there. We could say something like "But this only works because opacity, clipping, and blending don't interact! If, we were applying a blur instead of opacity, then the order would matter: blur-then-clip results in a sharp outline, while clip-then blur gives you blurry edges. Carefully reasoning through situations like this is important to make a browser as fast as possible."

> I think your hyperlink didn't work out...

Lol, I just meant to refer to number 3 on my list, but Github assumed it was an issue. Woops!

> I left it separate by analogy with SaveLayer. I could do a SaveAndClip display item.

Yes. I'd just call it `ClipRRect`. Here's why. The actual `clipRRect` call doesn't do anything, but it changes what later commands do. So, it's stateful (like modifying a variable). But a `ClipRRect(cmds ...)` that saves, clips, runs the `cmds`, and restores doesn't affect any visible state, and its results are only affected by its "inputs", namely `cmds`. I think this model will be easier for students to understand as a result.

pavpanchekha:
> I like that it's minimal code differences in the current code, but gets across the point that you can skip surfaces, and in the end the logic is pretty simple. We're not implementing a real browser after all, where even a "new" operator is to be avoided.

Yeah, that's true, I like that aspect of it. I think it comes down to whether we want to combine alpha with blend mode or not. If we do, we can't keep the simple structure anyway. If we don't, the simple structure is best.  As for whether to combine alpha with blend mode, I definitely think we should *talk through* the decision in the text with the reader. The crux is: you _can_ combine alpha with blend mode, but you _can't_ combine most filters with blend modes, alpha + blending is a special case. Then in the text we either say "to emphasize that it's a special case, we don't do that optimization in our browser", or we go ahead and do it. I'd write it both ways and see whether one reads kind of silly or not in keeping with the rest of the book. It's not clear to me which is best.

chrishtr:
I ended up making two more of the changes you suggested:
* putting save/restore inside ClipRRect
* combining blend mode and opacity

I think I was wrong about the difficulties for filters, so no need to mention that.

Please take another look.

PR #294, opened 2021-12-13
Fix Skia-to-SDL pipeline for any endianness

pavpanchekha:
This PR adds specifies the pixel order when creating a Surface, so that the Skia colors are correct on all platforms. This has been tested on:

- [x] Pavel's x86 macOS
- [ ] Pavel's x86 Linux
- [ ] Chris's ARM (?) chromeOS
- [ ] Chris's (?) macOS


pavpanchekha:
Nope, this code fails on my Linux machine. It seems to have byte order BGRA, while my macOS has byte order RGBA.

pavpanchekha:
Ok, @chrishtr, I rewrote the test script to actually compute the correct mask values, and describe it that way in the text.

I've also completed experiments on my Linux machine the issue appears to be in Skia (I see different byte values on the two OSs, despite both running x86) not in SDL. More precisely, take the `generate_masks.py` file in this PR and add this text on line 44:

    print(r_bytes, g_bytes, b_bytes)

The code does not even call any SDL methods up to this point, so any differences are due to Skia only. I get a different output on Arch Linux (where I see order BGRA) and on macOS (where I see RGBA).

chrishtr:
I asked the Skia eng lead and he said it's a compile-time flag whether to use RGBA or BGRA.

"You can check whether kN32_SkColorType == kBGRA_8888_SkColorType or kRGBA_8888_SkColorType"

Is this exposed in the python bindings?

chrishtr:
They said that by default Skia uses BGRA on Windows and RGBA on all other platforms. Perhaps the dll backing skia python is compiled with a particular setting.

chrishtr:
Smoking gun:

https://source.chromium.org/chromium/chromium/src/+/master:third_party/skia/BUILD.gn;l=36

chrishtr:
He says  you can specify the color type when constructing the surface: kRGBA_8888_ColorType

pavpanchekha:
Ok, interesting, thanks for getting to the bottom of this. I've implemented this and it works. You gotta do this to create the surface:

    image_info = skia.ImageInfo.Make(300, 100, ct=skia.kBGRA_8888_ColorType, at=skia.kUnpremul_AlphaType)
    skia_surface = skia.Surface.MakeRaster(image_info)

Working on incorporating this into the text.

pavpanchekha:
Ready for re-review

book/visual-effects.md
```diff
@@ -122,20 +130,42 @@ class Browser:
         # doesn't actually copy anything yet.
         skia_image = self.root_surface.makeImageSnapshot()
         skia_bytes = skia_image.tobytes()
-        rect = sdl2.SDL_Rect(0, 0, WIDTH, HEIGHT)
+```
+
+Next, we need to copy the data to an SDL surface. This requires
+telling SDL what order the pixels are stored in and on your computer's
```
chrishtr, line 136:
```suggestion
telling SDL what order the pixels are stored in (that's why we specified RGBA_888 above).
```

book/visual-effects.md
```diff
@@ -122,20 +130,42 @@ class Browser:
...
+[endianness][wiki-endianness]:
```
chrishtr, line 137:
```suggestion
```

book/visual-effects.md
```diff
@@ -122,20 +130,42 @@ class Browser:
...
+
+[wiki-endianness]: https://en.wikipedia.org/wiki/Endianness
```
chrishtr, line 139:
```suggestion
```

chrishtr (approved):
Approved with two comments.

pavpanchekha:
Why remove the comments on endianness? Sadly it is still relevant, because SDL expects integer masks and the integer endianness the affects byte order.

chrishtr:
Ok yeah, I guess it does. Nevermind then, go ahead and commit.

PR #301, opened 2021-12-16
Pass over "Summary", "Outline", and "Exercises" in Chapter 11

pavpanchekha:
The PR elaborates the summary, adds an outline, and heavily modifies the exercises. In the exercises:

- I clarified the "transforms", "filters", and "overflow: scroll" exercises.
- I removed the "width/height" exercise since it duplicates an exercise from a prior chapter
- I removed the "z-index" exercises, which was quite sad, because without implementing `position` there's no way for `z-index` to have an effect. I considered adding that to the exercise (or making it depend on the transforms exercise—but I'm not sure that's enough) and that seemed too complex but I want to get a second opinion on that.
- I added an exercise on "interest regions". I'm not sure what's the best way to write this exercise, so would appreciate your take
- I added an exercise on hit testing, which also corrects a subtle bug in our rounded corners implementation, which is nice.

### DIFF
--- a/book/visual-effects.md
+++ b/book/visual-effects.md
@@ -1676,14 +1676,18 @@ are almost endless.
 Summary
 =======
 
-So there you have it. Now we don't have just a boring browser that can only
-draw simple input boxes plus text. It now supports:
+So there you have it: our browser can draw not only boring
+black-on-white text but also:
 
-* Opacity
-* Blending
-* Rounded-corner clips via destination-in blending or direct clipping
-* Surfaces for scrolling and animations
-* Optimizations to avoid surfaces
+- Partial transparency via an alpha channel
+- User-configurable blending modes via `mix-blend-mode`
+- Rounded rectangle clipping via destination-in blending or direct clipping
+- Optimizations to avoid surfaces when possible
+- Surfaces for scrolling and animations
+
+Besides the new features, we've upgraded from Tkinter to SDL and Skia,
+which makes our browser faster and more responsive and sets a
+foundation for more work on browser performance to come.
 
 ::: {.further}
 [This blog post](https://ciechanow.ski/alpha-compositing/) gives a really nice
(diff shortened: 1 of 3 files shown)


book/visual-effects.md
```diff
@@ -1676,14 +1676,18 @@ are almost endless.
 Summary
 =======
 
-So there you have it. Now we don't have just a boring browser that can only
-draw simple input boxes plus text. It now supports:
+So there you have it: our browser can draw not only boring
+black-on-white text but also:
```
chrishtr, line 1680:
```suggestion
text and boxes but also:
```

chrishtr: (it's not just "black on white"

book/visual-effects.md
```diff
@@ -1676,14 +1676,18 @@ are almost endless.
...
-* Surfaces for scrolling and animations
-* Optimizations to avoid surfaces
+- Partial transparency via an alpha channel
+- User-configurable blending modes via `mix-blend-mode`
+- Rounded rectangle clipping via destination-in blending or direct clipping
+- Optimizations to avoid surfaces when possible
+- Surfaces for scrolling and animations
+
+Besides the new features, we've upgraded from Tkinter to SDL and Skia,
+which makes our browser faster and more responsive and sets a
```
chrishtr, line 1689:
```suggestion
which makes our browser faster and more responsive, and sets a
```

book/visual-effects.md
```diff
@@ -1695,73 +1699,64 @@ highly recommend it.
 :::
 
 
-Exercises
-=========
-
-*z-index*: Right now, the order of paint is a depth-first traversal of the
```
chrishtr, line 1701: How about putting it back and just say that it also requires position:relative, which should be implemented,
except that we won't ask them to support top and left?

pavpanchekha: The issue is that even we add `position:relative` we can't create any overlaps between elements with the minimal CSS support we have. `top`/`left`, `transform`, `width`/`height`, and many other options are enough. I'll write it in a more general sense and add it back.

pavpanchekha: I added it back, take a look

PR #310, opened 2021-12-21
Fix lab10+ requests to allow omitting the trailing slash in a URL

chrishtr:


src/lab10.py
```diff
@@ -39,7 +39,10 @@ def request(url, top_level_url, payload=None):
     assert scheme in ["http", "https"], \
         "Unknown scheme {}".format(scheme)
 
+    if url.find("/") < 0:
```
pavpanchekha, line 42: I would use `"/" not in url`

pavpanchekha (commented):
I'm concerned that this will not work when combined with the `resolve_url` function. Does this work when `resolve_url` receives a current URL without a terminating slash? If yes, good to merge (though see Python nitpick below, which appears in two places.)

src/lab10.py
pavpanchekha: This appears in `lab6.py` as well.

chrishtr: Fixed

chrishtr:
> Does this work when `resolve_url` receives a current URL without a terminating slash? 

Yes. That's why I had to edit resolve_url in lab6. Without that fix it wasn't working.

chrishtr:
(There is also a test in this PR for that situation).

PR #321, opened 2021-12-29
Add a url parameter to specify single-threaded mode

chrishtr:
This is useful for comparing performance with threaded mode.

### DIFF
--- a/src/lab12.py
+++ b/src/lab12.py
@@ -1364,12 +1364,10 @@ if __name__ == "__main__":
     import argparse
 
     parser = argparse.ArgumentParser(description='Chapter 12 code')
-    parser.add_argument("--url", default=2, type=str, required=True,
-        help="URL to load")
-    parser.add_argument('--single_threaded', type=bool, default=False,
+    parser.add_argument("url", type=str, help="URL to load")
+    parser.add_argument('single_threaded', action="store_true", default=False,
         help='Whether to run the browser without a browser thread')
     args = parser.parse_args()
-    
 
     USE_BROWSER_THREAD = not args.single_threaded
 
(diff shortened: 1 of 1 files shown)


src/lab12.py
```diff
@@ -1361,10 +1361,21 @@ def handle_quit(self):
 
 if __name__ == "__main__":
     import sys
+    import argparse
+
+    parser = argparse.ArgumentParser(description='Chapter 12 code')
+    parser.add_argument("--url", default=2, type=str, required=True,
```
pavpanchekha, line 1367:
```suggestion
    parser.add_argument("url", type=str, required=True,
```

The name should not have dashes if it's a positional (not named) argument, and `default` describes a default value, not the position.

src/lab12.py
```diff
@@ -1361,10 +1361,21 @@ def handle_quit(self):
...
+        help="URL to load")
+    parser.add_argument('--single_threaded', type=bool, default=False,
+        help='Whether to run the browser without a browser thread')
```
pavpanchekha, line 1370: I think instead of `type=bool` you want `action="store_true"`. Argparse is pretty annoying.

pavpanchekha (commented):
I approve the idea of making a parameter here (we can describe it in the text if we want, or not if it's just for testing) but `argparse` is pretty screwy and I left some comments on what command line API I think you wanted. You can merge once you're happy with how the arguments are passed.

src/lab12.py
chrishtr: Apparently the -- is needed...seems to work as desired now.

Examples here use it: https://docs.python.org/3/library/argparse.html

PR #323, opened 2021-12-30
Remove dead code - cookie_string

chrishtr:


pavpanchekha:
Good catch, this is from back in the day when Ch10 allowed multiple cookies per domain. Huge win to remove it.

PR #327, opened 2022-01-14
Add author email to homepage

chrishtr:


pavpanchekha (approved):
Let's put it on the same line as the Patreon thing, either way let's merge and push.

PR #329, opened 2022-01-15
Add an implementation of setTimeout to chapter 12

chrishtr:
This accomplishes the following:
* Justifies the addition of Task and TaskQueue via an example
* Starts to flesh out what will become the MainThreadEventLoop
* Introduces threading a little bit earlier in the chapter
* Gives a concrete example of why scheduling and event loops are needed

### DIFF
--- a/book/scheduling-and-threading.md
+++ b/book/scheduling-and-threading.md
@@ -169,15 +169,15 @@ But now it's pretty clear that we have a lot more control over when to run
 scripts. For example, it's easy to make a change to `TaskRunner` to only run
 one script per second, or to not run them at all during page load, or when a
 tab is not the active tab. This flexibilty is quite powerful, and we can use it
-without having to dive into the guts of a `Tab` or how it loads web pages at
-all---all we'd have to do is implement a new `TaskRunner` heuristic.
+without having to dive into the guts of a `Tab` or how it loads web pages---all
+we'd have to do is implement a new `TaskRunner` heuristic.
 
 Now for the fun part I promised. Let's implement the
 [`setTimeout`][settimeout] JavaScript API, which provides a way to run
 JavaScript a given number of milliseconds from now. In terms of the JavaScript
 and Python communication, it'll use an approach with handles, similar to the
 `addEventListener` code we added in [Chapter 9](scripts.md#event-handling).
-The new part will be our first use of Python threads.
+The *new* part will be our first use of Python threads.
 
 [settimeout]: https://developer.mozilla.org/en-US/docs/Web/API/setTimeout
 
--- a/book/scheduling-and-threading.md
+++ b/book/scheduling-and-threading.md
@@ -34,7 +34,7 @@ feature of modern browsers.
 Task queues
 ===========
 
-At the moment, our browser has a lot of work tangled up, and it will take some
+At the moment, our browser has a lot of entangled code, and it will take
 substantial work to put everything in tasks on event loops. Let's start by
 defining the infrastructure of event loops, and then move JavaScript
 tasks to the new infrastructure. Then we can reward ourselves for this
@@ -59,12 +59,9 @@ it.[^event-loop] The job of the event loop is to schedule tasks
 [chapter 11](visual-effects.md#sdl-creates-the-window) (before that, we used
 `tkinter.mainloop`).
 
-Let's implement the `Task` and `TaskQueue` classes. Then we can move all of the
-event loop tasks into task queues later in the chapter.
-
-A `Task` encapsulates some code to run in the form of a function, plus arguments
-to that function.[^task-notes] A `TaskQueue` is simply a
-first-in-first-out list of `Task`s.
+Implement the `Task` and `TaskQueue` classes. A `Task` encapsulates some code to
+run in the form of a function, plus arguments to that function.[^task-notes] A
+`TaskQueue` is simply a first-in-first-out list of `Task`s.
 
 [^task-notes]: In `Task`, we're using the varargs syntax for Python, allowing
 any number of arguments to be passed to the task. We're also using Python's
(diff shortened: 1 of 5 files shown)


book/scheduling-and-threading.md
```diff
@@ -98,6 +102,167 @@ class TaskQueue:
...
+[Python thread][python-thread] via the `threading.Timer` class, which takes
+two parameters: a time delta in seconds from now, and a function to call when
+that time expires.
+
+[python-thread]: https://docs.python.org/3/library/threading.html
+
+``` {.python}
+def set_timeout(func, sec):     
+    t = threading.Timer(sec, func)
+    t.start()
```
pavpanchekha, line 204: I'm not totally clear why it's worth making a separate function for this—`threading.Timer(sec, func).start()` is no longer than calling `set_timeout`...

book/scheduling-and-threading.md
```diff
@@ -98,6 +102,167 @@ class TaskQueue:
...
+    delete SET_TIMEOUT_REQUESTS[handle];
+    callback();
+}
+```
+
+On the Python side, add a binding for `setTimeout` and an implementation that
+calls `set_timeout`. However, we have to be careful here, since in the code
+below, `run_callback` will run *on a different Python thread than the current
+one*. So we can't just call `evaljs` directly, or we'll end up with JavaScript
+running on two Python threads at the same time, which is not ok.[^js-thread]
```
pavpanchekha, line 233: This is really great.

book/scheduling-and-threading.md
```diff
@@ -98,6 +102,167 @@ class TaskQueue:
...
+
+Not all is lost though! This problem can be solved elegantly with the
+`TaskRunner`: instead of running the script right away, schedule a task to
+do it later, when the other thread is free.[^locking] Here's the code:
+
+[^locking]: This code has a bug: it accesses data structures shared
+between two threads without a thread lock. We'll see soon how to fix it.
```
pavpanchekha, line 240: ATM this chapter moves to async rendering first, and introduces all the necessary locks as it goes. I'm wondering whether, with `setTimeout`, there's a chance to introduce all the locks and discuss cross-thread communication first, before splitting off rendering. That would hopefully be a bit simpler?

book/scheduling-and-threading.md
```diff
@@ -305,27 +470,14 @@ class Browser:
...
-        header, body = request(script_url, url)
-        set_timeout(0, Task(self.js.run, script, body))
-```
-
-As you probably know, scripts are not just for running straight through in one
-task, or responding to input events. They can also schedule more events to be
-put on the rendering event loop and run later. There are multiple JavaScript
-APIs in browsers to do this, but let's focus on the one most related to
-rendering: `requestAnimationFrame`. It's used like this:
+Scripts are not just just for tasks unrelated to rendering. In fact, many
```
pavpanchekha, line 476: "just just" is a typo

book/scheduling-and-threading.md
```diff
@@ -305,27 +470,14 @@ class Browser:
...
+scripts tasks are there to update rendering state in the DOM after a state
```
pavpanchekha, line 477: "script tasks"

book/scheduling-and-threading.md
```diff
@@ -336,12 +488,13 @@ requestAnimationFrame(callback);
...
 
 This code will do two things: request an *animation frame* (rendering) task to
-be run on the event loop,[^animation-frame] and call `callback` at the
-beginning of that rendering task. This is super useful to web page authors, as
-it allows them to do any setup work related to rendering just before it occurs.
-The implementation of this JavaScript API is straightforward: add a new dirty
-bit to `Tab` and code to call the JavaScript callbacks during the next
-animation frame.
+be scheduled on the event loop,
+[^animation-frame] and call `callback` *at the beginning* of that rendering
```
pavpanchekha, line 492: Footnotes like this should go right after punctuation, without a space.

book/scheduling-and-threading.md
```diff
@@ -713,21 +866,23 @@ thread, and all code in `Tab` and `JSContext` will run on the main thread.
...
-rendering pipeline dirty bit.
+the main thread. Let's add more code to the `TaskRunner` class to make it
+into a complete event loop, and then rename it to `MainThreadEventLoop`.
+
+The two threads will communicate by reading and writing shared data structures,
+and use `threading.Lock` objects to prevent race conditions.
+`MainThreadEventLoop` will be the only class allowed to call methods on `Tab`
+or `JSContext`.
+
+`MainThreadEventLoop` will add a lock and a thread object. Calling `start` will
```
pavpanchekha, line 877: Since `MainThreadEventLoop` holds the lock and thread, it's honestly a little confusing to me why `TaskQueue` is even a separate class. It seems like it is just a list...

book/scheduling-and-threading.md
```diff
@@ -713,21 +866,23 @@ thread, and all code in `Tab` and `JSContext` will run on the main thread.
...
+begin the thread. This will excute the `run` method on that thread; `run`
```
pavpanchekha, line 878: "excute" -> "execute"

book/scheduling-and-threading.md
```diff
@@ -713,21 +866,23 @@ thread, and all code in `Tab` and `JSContext` will run on the main thread.
...
+(instead of `run_once`) will execute forever (or until the program quits, which
+is indicated by the `needs_quit` dirty bit) and is where we'll put the main
+thread event loop. There will also be a task queue for browser-generated tasks
+and scripts, and a rendering pipeline dirty bit.
```
pavpanchekha, line 882: I think we add the condition variable later, but it seems like here is actually a good place to describe it...

book/scheduling-and-threading.md
chrishtr: The aim was:

* Introduce Python threading in baby steps, without having to get into how it's used
* Hide that behind a method that may be more familiar once you know how setTimeout is supposed to be used in JS

Perhaps that's not a good enough set of reasons..

book/scheduling-and-threading.md
chrishtr: Fixed in direct commit to main.

book/scheduling-and-threading.md
chrishtr: Fixed in direct commit to main.

book/scheduling-and-threading.md
chrishtr: My aim was to represent the concept name introduced in the chapter by a class..

book/scheduling-and-threading.md
chrishtr: Fixed in direct commit to main.

book/scheduling-and-threading.md
chrishtr: Good idea. I will put up a PR that does better, by delaying introduction of the condition variable.

book/scheduling-and-threading.md
chrishtr: Done in PR 340.

PR #330, opened 2022-01-16
Make scrolling a main-thread task at first

chrishtr:
..and only in the threaded scrolling section do the work to make it threaded.

### DIFF
--- a/book/scheduling-and-threading.md
+++ b/book/scheduling-and-threading.md
@@ -770,7 +770,7 @@ In `run`, implement a simple event loop scheduling strategy that runs the
 rendering pipeline if needed, and also one browser method and one script task,
 if there are any on those queues.
 
-The part at thet end is the trickiest. First, check if there are more tasks
+The part at the end is the trickiest. First, check if there are more tasks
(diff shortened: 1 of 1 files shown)


book/scheduling-and-threading.md
```diff
@@ -1059,16 +1056,18 @@ Skia is thread-safe, but SDL may not be.
...
 ==================
 
-Recall how we've added some scroll-related code, but we didn't really get into
-how it works (I also omitted some of the code). Let's now carefully examine
-how to implement threaded scrolling.
+Two sections back we talked about how important it is to run scrolling on the 
+browser thread, to avoid slow scripts. But right now, even though there
+is a browser thread, scrolling still happens in a `Task` on the main thread.
+Let's now make scrolling truly *threaded*, meaning it runs on the browser
+thread in parallel with scripts and other main thread work.
```
pavpanchekha, line 1063: Honestly I would move the discussion of blocked scrolling to here, but right now this is hard to do because there are multiple dependencies on that example. We can discuss it next time. In any case, I think the scrolling discussion is much better now, since it now reads as an interesting complication on top of an existing, understood threading architecture.

PR #332, opened 2022-01-17
Fix incorrect wording and add a warning footnote about GC

chrishtr:
Previously, the text claimed that SDL_CreateRGBSurfaceFrom copied the bytes.
Instead, it wraps the bytes and [does not copy them](https://wiki.libsdl.org/SDL_CreateRGBSurfaceFrom).

This PR fixes that and adds a warning footnote about the possibility of bugs involving
mixed Python & C++.

[This discussion](https://github.com/browserengineering/book/discussions/328) brought the bug to my attention.

### DIFF
--- a/book/visual-effects.md
+++ b/book/visual-effects.md
@@ -162,17 +162,17 @@ class Browser:
 ```
 
 The `CreateRGBSurfaceFrom` method then wraps the data in an SDL surface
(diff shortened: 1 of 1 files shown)


pavpanchekha:
Retroactive :+1:

PR #333, opened 2022-01-18
Fix a race condition with display_scheduled

chrishtr:
There was a bug in which the display_scheduled variable on a Tab could be accessed from
multiple threads (via a threading.Timer thread). I fixed it and moved it to main_thread_runnner,
which is also simpler.

I also added text explaining that the threading locks really are useful, and why.


chrishtr:
Turns out this fixed the same race condition I fixed in december, but that PR languished so long I forgot about it... :(

PR #338, opened 2022-01-22
Improve async rendering section of chapter 12

chrishtr:
This PR does the following:

* Improves description of the what and why of the code
* Breaks up the code more with explanatory text
* Fixes a bug where `run_rendering_pipeline` was called on the timer thread, and replaces it with a call on the event loop.

### DIFF
--- a/book/scheduling-and-threading.md
+++ b/book/scheduling-and-threading.md
@@ -174,7 +174,7 @@ Asynchronous rendering
 ======================
 
 In order to separate rendering from other tasks and make it into a proper
-pipeline, let make it asynchronous. Instead of updating rendering right away,
+pipeline, let's make it asynchronous. Instead of updating rendering right away,
 the browser should set *dirty bits* indicating that a particular part of the
 pipeline needs updating. Then when the pipeline is run, it will run the parts
 indicated by the dirty bits. We'll also need some way of
@@ -193,6 +193,10 @@ function completes, the timer thread is automatically ended.
 execute at about the time desired---even if the current thread is busy at that
 time---but, as we'll see, leads to some complications if you really want the
 function to be called on the same thread as the one calling `set_timeout`.
+But in any case, it wouldn't make any sense to run the task on the "current"
+thread, because the only time the current thread is free to run such a task
+is when the program has quit. Unlike JavaScript programming on a website,
+Python has no built-in event loop.
 
 [python-thread]: https://docs.python.org/3/library/threading.html
 
@@ -338,7 +342,7 @@ class Browser:
         self.needs_draw = False
 ```
 
-In each case where raster or draw was previously called synchronously, set the
+In each case where raster or draw was previously synchronous, set the
 dirty bits instead. Here's the change to `handle_click`:
 
 ``` {.python expected=False}
--- a/book/scheduling-and-threading.md
+++ b/book/scheduling-and-threading.md
@@ -195,8 +195,8 @@ time---but, as we'll see, leads to some complications if you really want the
 function to be called on the same thread as the one calling `set_timeout`.
 But in any case, it wouldn't make any sense to run the task on the "current"
 thread, because the only time the current thread is free to run such a task
-is when the program has quit. Unlike JavaScript programming on a website,
-Python has no built-in event loop.
+is when the program has already quit! Unlike JavaScript programming on a
+website, Python has no built-in event loop.
 
 [python-thread]: https://docs.python.org/3/library/threading.html
 
@@ -209,9 +209,9 @@ def set_timeout(func, sec):
 Next, add three dirty bits to `Tab`:
 
 * `needs_pipeline_update`, indicating that
-the pipeline needs to be re-run
+the pipeline needs to be re-run.
 * `display_scheduled`, indicating that `set_timeout` was already called (this
-avoids double-running `set_timeout` unnecessarily)
+avoids double-running `set_timeout` unnecessarily).
 * `run_pipeline_now`, indicating that the event loop should
 run the pipeline.
 
--- a/book/scheduling-and-threading.md
+++ b/book/scheduling-and-threading.md
@@ -174,18 +174,25 @@ Asynchronous rendering
 ======================
 
 In order to separate rendering from other tasks and make it into a proper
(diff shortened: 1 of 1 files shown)


book/scheduling-and-threading.md
```diff
@@ -174,18 +174,29 @@ Asynchronous rendering
...
+class, which takes two parameters: a time delta from now, and a function to
+call when that time expires. It will start a new thread and call that
+function *on that thread*[^thread-timer] at the desired time. When the
+function completes, the timer thread is automatically ended.
+
+[^thread-timer]: This is convenient because it enables the timer to
+execute at about the time desired---even if the current thread is busy at that
+time---but, as we'll see, leads to some complications if you really want the
+function to be called on the same thread as the one calling `set_timeout`.
+But in any case, it wouldn't make any sense to run the task on the "current"
```
pavpanchekha, line 196: I found this footnote really confusing. The actual point it's making is pretty clear, so we should frame it positively (here's what's true) instead of negatively (here's why that other belief is false).

book/scheduling-and-threading.md
```diff
@@ -195,34 +206,53 @@ def set_timeout(func, sec):
...
-Also, rename `render` to `run_rendering_pipeline`, and add a new
-`run_animation_frame` method that runs the pipeline and calls the other
-rendering pipeline stages on `Browser` via a new `raster_and_draw`
-method (which we'll implement shortly).
+Next, add three dirty bits to `Tab`:
+
+* `needs_pipeline_update`, indicating that
+the pipeline needs to be re-run.
+* `display_scheduled`, indicating that `set_timeout` was already called (this
+avoids double-running `set_timeout` unnecessarily).
```
pavpanchekha, line 214: I'm a little confused:

1) Why it's called `display_scheduled`, if it's about `set_timeout`
2) What does this mean by double-running `set_timeout`?
3) How these flags relate to each other.

pavpanchekha: More precisely, I think this goes back to my concern that there are a lot of flags and we haven't given the reader any way of thinking about them. Why so many flags? How do you remember how they relate? How do you choose to have more or fewer? I'm not actually sure if we want these dirty flags to be a major player in this chapter, but if we do we should talk more about managing them.

book/scheduling-and-threading.md
```diff
@@ -195,34 +206,53 @@ def set_timeout(func, sec):
...
+* `run_pipeline_now`, indicating that the event loop should
+run the pipeline.
+
+Add methods to set the dirty bits and call `set_timeout` as
+needed.[^race-condition]
+
+[^race-condition]: This code has a race condition because it's setting variables
+on one thread from another without a thread lock. Let's ignore that for now and
+fix it later in the chapter.
```
pavpanchekha, line 223: Potentially we should move discussion of locks way up to where we first introduce threads (setTimeout) so that we don't have to continuously apologize for it throughout the book.

book/scheduling-and-threading.md
```diff
@@ -195,34 +206,53 @@ def set_timeout(func, sec):
...
 
         if not self.display_scheduled:
             set_timeout(callback, REFRESH_RATE_SEC)
             self.display_scheduled = True
+```
+ 
+Also, rename `render` to `run_rendering_pipeline`, and add a new
+`run_animation_frame` method that runs the pipeline and calls the other
+rendering pipeline stages on `Browser` via a new `raster_and_draw`
+method.
```
pavpanchekha, line 250: This is much better.

book/scheduling-and-threading.md
```diff
@@ -237,10 +267,33 @@ class Tab:
...
+
+``` {.python expected=False}
+    while True:
+        if sdl2.SDL_PollEvent(ctypes.byref(event)) != 0:
+            # ...
+        active_tab = browser.tabs[browser.active_tab]
+        if active_tab.run_pipeline_now:
+            active_tab.run_pipeline_now = False
+            active_tab.run_animation_frame()
+```
```
pavpanchekha, line 291: Is here the right place to explain why we must set `run_pipeline_now` before running the animation frame? It is related somehow to the idea that you need to be able to schedule a rAF inside a rAF callback maybe...

pavpanchekha (approved):

pavpanchekha:
I think my main concern with this PR and also the previous one on setTimeout is that we now have a lot of flags. We should simplify, organize, or explain. I don't think that should block merging, however.

book/scheduling-and-threading.md
```diff
@@ -174,18 +174,29 @@ Asynchronous rendering
...
-`threading.Timer` class, which takes two parameters: a time delta from now, and
-a function to call when that time expires.
+This function will run a callback at a specified time in the future. You can do
+that by starting a new [Python thread][python-thread] via the `threading.Timer`
+class, which takes two parameters: a time delta from now, and a function to
+call when that time expires. It will start a new thread and call that
+function *on that thread*[^thread-timer] at the desired time. When the
+function completes, the timer thread is automatically ended.
+
+[^thread-timer]: This is convenient because it enables the timer to
```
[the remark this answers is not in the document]
chrishtr: Yes-- see line 189

book/scheduling-and-threading.md
chrishtr: Fixed in another PR.

book/scheduling-and-threading.md
chrishtr: display_scheduled avoids running stuff twice, but doesn't mean the event loop should run just now. needs_animation_frame means "please run it at the next opportunity"

display_scheduled avoids double running because if you call set_timeout twice, before the callback has run, the second-time will be a no-op.

Unfortunately I do think we need at least some of these dirty flags or the browser will not function correctly (e.g. the ones above). Other flags are optional and for performance reasons. We could push them to the end of the chapter.

book/scheduling-and-threading.md
chrishtr: ok, good point

chrishtr:
TL;DR I agree that we should attempt to simplify and keep improving. Will keep at it.

book/scheduling-and-threading.md
pavpanchekha: Agreed, sounds good.

PR #341, opened 2022-01-26
Record raster time instead of raster-and-draw

chrishtr:


pavpanchekha (approved):

pavpanchekha:
One thing we might do is set up an explicit easier goal than 60fps. "Since our browser is much simpler, we're not going to hit 60 frames per second. But we can hit 20—which requires us to draw each frame in roughly 50 milliseconds." That way the reader feels accomplishment at the end of the chapter instead of failure.

chrishtr:
Good point. I also hope to do some more profiling to see if there is any obvious thing to make faster.

pavpanchekha:
Right. Or if, for example, one slow thing is blitting twice (from the tab / chrome surface to the browser surface, and then from there to SDL) we can at least say that explicitly ("this is slow but we're not changing it for simplicity's sake") and perhaps add an exercise to improve it.

PR #342, opened 2022-01-27
Improvements to async rendering section

chrishtr:
* Use dirty bits and implementation that takes advantage of the TaskRunner class defined in a previous section
* Use locks properly


pavpanchekha (approved):
Good, this definitely simplifies things a bit

PR #343, opened 2022-01-28
More cleanups in chapter 12

chrishtr:
This fixes and cleans up the descriptions in the sections introducing raF and the browser thread.

There were various stale references there and prose that could be optimized.


pavpanchekha (approved):
Looks good!

PR #344, opened 2022-01-29
Simplify browser->main thread communication

chrishtr:
Instead of special scroll and animation frame bits, just use a task.

There turns out to already be an exercise about a scheduling strategy that prioritizes rendering.


chrishtr:
Going to go ahead and commit this, because it is just simplifying refactoring along the lines of what we discussed yesterday, and
I want to fast-follow with deleting more dirty bits and the TabWrapper class.

PR #345, opened 2022-01-29
Consolidate the three Browser dirty bits

chrishtr:
This PR removes needs_chrome_raster, needs_tab_raster and needs_draw, and replaces them with a single dirty bit needs_raster_and_draw. Aims accomplished:

* Simplifies the code substantially (and found one bug in lab12.py as a result -- raster timing was wrong)
* Provides room for an exercise on this topic

I found that needs_raster_and_draw is necessary and can't be removed. The reason is that the browser thread event loop is unfortunately continuously polling and can't be made event-driven (because SDL doesn't seem to support it). So without this loop, the browser thread would have erratic behavior when trying to run animations, because it'd waste a lot of time doing useless rasters and delay responding to the main thread.


chrishtr:
Going to merge this one too, and the removal of TabWrapper after (PR 346). Then you can do a full review on Monday..

PR #346, opened 2022-01-29
Remove the TabWrapper class

chrishtr:
Instead, let Tab and Browser just call certain methods on each other.


chrishtr:
Merging this one now too. I'm sure there are some dangling references in the text that I missed, let's fix those in other passes.

I was very careful to do this CL in small steps and test along the way that all the browser features work, so I'm pretty sure I did it correctly.

PR #348, opened 2022-01-31
Remove the needs_commit variable

chrishtr:
It was just optimizing whether to do more work in the pipeline.


chrishtr:
This is a simple cleanup. It also fixes a typo "bug".

PR #352, opened 2022-02-01
Rework the rendering and task sections in chapter 12

chrishtr:
* Simplifications to the code and simpler names
* Moving XHR to the front to introduce threads in a simpler setting
* Making rendering a task, *then* talking about its candence, with a clear connection to rAF's relevance

### DIFF
--- a/book/scheduling-and-threading.md
+++ b/book/scheduling-and-threading.md
@@ -373,20 +373,26 @@ die.^[Note that for async requests, the return statement is meaningless;
 it's only there for the sync version.]
 
 ``` {.python}
+XHR_ONLOAD_CODE = "__runXHROnload(dukpy.out, dukpy.handle)"
+
 class JSContext:
+    def xhr_onload(self, out, handle):
+        do_default = self.interp.evaljs(
+            XHR_ONLOAD_CODE, out=out, handle=handle)
+
     def XMLHttpRequest_send(
         self, method, url, body, is_async, handle):
         full_url = resolve_url(url, self.tab.url)
         if not self.tab.allowed_request(full_url):
             raise Exception("Cross-origin XHR blocked by CSP")
+        if url_origin(full_url) != url_origin(self.tab.url):
+            raise Exception(
+                "Cross-origin XHR request not allowed")
 
         def run_load():
             headers, out = request(
                 full_url, self.tab.url, payload=body)
             handle_local = handle
-            if url_origin(full_url) != url_origin(self.tab.url):
-                raise Exception(
-                    "Cross-origin XHR request not allowed")
             self.tab.task_runner.schedule_task(
                 Task(self.xhr_onload, out, handle_local))
             return out
@@ -529,7 +535,14 @@ class Tab:
 ```
 
 All the places that call `raster_chrome`, `raster_tab` or `draw` directly will
-also need to call `set_needs_render` instead. Here's `handle_down`:
+also need to call `set_needs_render` instead.[^render-instead] Here's
+`handle_down`:
+
+[^render-instead]: Technically, it's not necessary to do so, but thinking of all
+of rendering (including raster and draw) as one pipeline that's either run or
+not run is a good way to think about what is going on. Later we'll add ways to
+get back equivalent performance to rastering directly without resorting to
+a short-circuit of the rendering pipeline.
 
 ``` {.python expected=False}
 class Browser:
@@ -545,7 +558,7 @@ loop!
 Unfortunately, we also regressed the overall performance of the browser by
 quite a lot in some cases. For example, scrolling down will now cause
 the entire rendering pipeline (style, layout, etc.) to run, instead of
-just `draw`. Let's see how to fix tha.
+just `draw`. Let's see how to fix that.
 
 Animating frames
 ================
@@ -725,7 +738,7 @@ To not go faster than the refresh rate, use a `Timer`:
 
 ``` {.python expected=False}
 class Tab:
-    def schedule_animation_frame(self):
+    def schedule_animation_fraggme(self):
         # ...
         def callback():
             self.task_runner.schedule_task( \
--- a/book/scheduling-and-threading.md
+++ b/book/scheduling-and-threading.md
@@ -56,9 +56,7 @@ Task queues
 
 At the moment, our browser has a lot of entangled code, and it will take
 substantial work to put everything in tasks on queues. Let's start by defining
-the infrastructure. This will be a bit of work, so once it's done we'll reward
-ourselves for this work by adding a fun new feature built on top of this
-tech--the `setTimeout` API.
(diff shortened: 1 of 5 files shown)


book/scheduling-and-threading.md
```diff
@@ -312,256 +327,241 @@ situation---some devices have more [CPU cores][cores] than others, or are more
...
+        if not self.tab.allowed_request(full_url):
+            raise Exception("Cross-origin XHR blocked by CSP")
+
+        def run_load():
+            headers, out = request(
+                full_url, self.tab.url, payload=body)
+            handle_local = handle
+            if url_origin(full_url) != url_origin(self.tab.url):
+                raise Exception(
+                    "Cross-origin XHR request not allowed")
```
pavpanchekha, line 389: I think we can do this before starting the thread, right?

book/scheduling-and-threading.md
```diff
@@ -312,256 +327,241 @@ situation---some devices have more [CPU cores][cores] than others, or are more
...
+    def XMLHttpRequest_send(
+        self, method, url, body, is_async, handle):
+        full_url = resolve_url(url, self.tab.url)
+        if not self.tab.allowed_request(full_url):
+            raise Exception("Cross-origin XHR blocked by CSP")
+
+        def run_load():
+            headers, out = request(
+                full_url, self.tab.url, payload=body)
+            handle_local = handle
```
pavpanchekha, line 386: Do we need this? (I think the scoping works correctly in this case.)

book/scheduling-and-threading.md
```diff
@@ -312,256 +327,241 @@ situation---some devices have more [CPU cores][cores] than others, or are more
...
+            if url_origin(full_url) != url_origin(self.tab.url):
+                raise Exception(
+                    "Cross-origin XHR request not allowed")
+            self.tab.task_runner.schedule_task(
+                Task(self.xhr_onload, out, handle_local))
```
pavpanchekha, line 391: Is xhr_onload included in the book?

book/scheduling-and-threading.md
```diff
@@ -312,256 +327,241 @@ situation---some devices have more [CPU cores][cores] than others, or are more
...
+On thing that is special about rendering is that it's a "singleton" task. There
+is only one rendering task, and either it's been scheduled or not, but it
+doesn't make sense to schedule it twice at the same time. On the other hand,
+it's totally fine and natural to have many `setTimeout` or `XMLHttpRequest`
+callbacks pending at the same time. The singleton nature of the rendering task
+has to do with there being only one DOM to render.
 
-And two to `TaskRunner`:
+Because it's a singleton, we'll need an additional boolean variable on a `Tab`
+to avoid having two rendering tasks, called `needs_animation_frame`. This
```
pavpanchekha, line 468: This is a great explanation, but should we call it `has_rendering_task` or something like that instead of the current `needs_animation_frame`? Or we should drop in a footnote saying that the name will make a bit more sense later on.

book/scheduling-and-threading.md
```diff
@@ -312,256 +327,241 @@ situation---some devices have more [CPU cores][cores] than others, or are more
...
+render, and a new method `run_animation_frame` as the task callback. Note
+how we avoid two scheduled frames with an if statement.
 
-* `display_scheduled`, indicating that a rendering task was already scheduled
-via a `threading.Timer` (this avoids double-running a timer unnecessarily).
-* `needs_animation_frame`, indicating that the event loop should
-run the pipeline ASAP.
+[^animation-frame]: It's called an "animation frame" because sequential
+rendering of different pixels is an animation, and each time you render it's
+one "frame"---like a drawing in a picture frame.
```
pavpanchekha, line 476: Ah, the footnote is here. Let's just move this up to where we first use "animation frame"

book/scheduling-and-threading.md
```diff
@@ -312,256 +327,241 @@ situation---some devices have more [CPU cores][cores] than others, or are more
...
-``` {.python}
-class Browser:
-    def handle_down(self):
-        # ...
-        self.set_needs_raster_and_draw()
-```
+Unfortunately, we also regressed the overall performance of the browser by
+quite a lot in some cases. For example, scrolling down will now cause
+the entire rendering pipeline (style, layout, etc.) to run, instead of
+just `draw`. Let's see how to fix tha.
```
pavpanchekha, line 548: "tha" to "that"

book/scheduling-and-threading.md
```diff
@@ -312,256 +327,241 @@ situation---some devices have more [CPU cores][cores] than others, or are more
...
-            return
-        # ...
-        self.needs_raster_and_draw = False
-
-```
-
-Here's the change to `handle_click` to set the dirty bit instead of calling
-`raster_chrome` or `raster_tab` directly:
+All the places that call `raster_chrome`, `raster_tab` or `draw` directly will
+also need to call `set_needs_render` instead. Here's `handle_down`:
```
pavpanchekha, line 532: Is this actually necessary? I guess so for "raster_tab" but not for "raster_chrome" or "draw", right? I'm not too sure.

book/scheduling-and-threading.md
```diff
@@ -723,15 +692,141 @@ def show_count():
...
 ```
 
+Load this up and observe an animation from 0 to 100.
+
+However, there is a pretty big flaw in our implementation that might be apparent
+when you try this demo. Depending on your computer's speed, it might run really
+fast or really slow. In fact, the animation will take a total amount of time
+about equal to the time it takes to run 100 rendering tasks. This is
+bad---animations should have a smooth and consistent frame rate, or *cadence*,
+regardless of the computer hardware.
```
pavpanchekha, line 702: This was missing from the chapter before and is a key addition!

pavpanchekha (approved):
I think this is a great improvement to the chapter!

book/scheduling-and-threading.md
chrishtr: Done

book/scheduling-and-threading.md
chrishtr: Fixed.

book/scheduling-and-threading.md
chrishtr: Isn't this the paragraph that first does so?

book/scheduling-and-threading.md
chrishtr: Fixed

book/scheduling-and-threading.md
chrishtr: It's not "necessary", but I put it in because they should think of the entire pipeline as one thing. I added a footnote.

book/scheduling-and-threading.md
chrishtr: Cool

PR #356, opened 2022-02-03
Improve the parallel loading section

chrishtr:
* Better explanation of key benefit
* Example of using join()
* Breaks up the code examples
* Upgrades a footnote to a learn-more block


pavpanchekha (approved):

pavpanchekha:
I did not review this deeply. Instead I will merge and do a full read.

PR #357, opened 2022-02-04
Chapter 12 wordsmithing and polishing pass

pavpanchekha:
This PR contains a pass over roughly the first half of Chapter 12. It does minor reorganizations, wordsmithing, and polish, plus a couple of major changes:

- [x] Merge the Tasks and Task Queues sections. With the various changes throughout the chapter over the last month, there's much less need for an organizational section at the top, and we can use it to introduce the `Task` and `TaskRunner` instead.
- [x] Replace `Task.__call__` with `Task.run`. The use of the `__call__` convention goes back to initial experiments with the Tkinter task queue.
- [x] Rephrased the discussion of multiple threads and locks to be positive ("we have to do XXX") instead of negative ("we're done! but we have a bug")
- [x] Renamed `xhr_onload` to `dispatch_xhr_onload` to parallel `dispatch_event`

### DIFF
--- a/book/scheduling-and-threading.md
+++ b/book/scheduling-and-threading.md
@@ -1232,12 +1232,16 @@ class Tab:
         self.render()
         # ...
         self.browser.commit(
-            self.url, self.scroll,
+            self, self.url, self.scroll,
             document_height,
             self.display_list)
 ```
 
-In `Browser`:
+In `Browser`, commit will copy across the url, scroll offset and display list,
+and call `set_needs_raster_and_draw` as needed. Since each `Tab` has its own
+thread that is always running, the `tab` parameter is
+compared with the active tab to avoid committing display lists for invisible
+tabs.
 
 ``` {.python}
 class Browser:
@@ -1245,8 +1249,12 @@ class Browser:
         self.url = None
         self.scroll = 0
 
-    def commit(self, url, scroll, tab_height, display_list):
+    def commit(self, tab, url, scroll, tab_height, display_list):
         self.lock.acquire(blocking=True)
+        self.display_scheduled = False
+        if tab != self.tabs[self.active_tab]:
+            self.lock.release()
+            return
         if url != self.url or scroll != self.scroll:
             self.set_needs_raster_and_draw()
         self.url = url
@@ -1274,13 +1282,32 @@ Why the browser thread and not the main thread? The reason
 is simple: there is no point to rendering display lists
 faster than they can be drawn to the screen.
 
-Implement this by adding a `needs_animation_frame` dirty bit on `Browser`.
+Implement this by adding a `needs_animation_frame` dirty bit on `Browser`. Tabs
+call a special version of this method that uses a lock and disallows setting
+the bit for a non-active tab.^[The `Browser` use-cases that set this dirty bit
+also need a lock, but all of the calling functions already hold the lock. If
+they tried to call the version that locks then a [deadlock] would occur.]
+
+Setting the bit only for active tabs prevents others from setting a
+dirty bit they don't need (because there is nothing to display for a non-active
+tab), and elegantly prevents any `requestAnimationFrame`
+callbacks from running. Try making a second tab while the counter demo is
+running, then go back to the demo tab. Notice that it stopped counting up
+while the other tab was visible, and resumes when it is made visible again!
+
+[deadlock]: https://en.wikipedia.org/wiki/Deadlock
 
 ``` {.python}
 class Browser:
     def __init__(self):
         # ...
         self.needs_animation_frame = False
+
+    def set_tab_needs_animation_frame(self, tab):
+        self.lock.acquire(blocking=True)
+        if tab == self.tabs[self.active_tab]:
+            self.needs_animation_frame = True
+        self.lock.release()
 
     def set_needs_animation_frame(self):
         self.needs_animation_frame = True
@@ -1297,9 +1324,7 @@ if __name__ == "__main__":
     while True:
         # ...
         browser.raster_and_draw()
-        if browser.needs_animation_frame:
-            browser.schedule_animation_frame()
-        browser.needs_animation_frame = False
+        browser.schedule_animation_frame()
 ```
 
 And `schedule_animation_frame` on `Browser` works just like the version
@@ -1314,30 +1339,30 @@ class Browser:
     def schedule_animation_frame(self):
         def callback():
             self.lock.acquire(blocking=True)
-            self.display_scheduled = False
             active_tab = self.tabs[self.active_tab]
-            self.lock.release()
             active_tab.task_runner.schedule_task(
                 Task(active_tab.run_animation_frame))
+            self.lock.release()
         self.lock.acquire(blocking=True)
-        if not self.display_scheduled:
+        if not self.display_scheduled and self.needs_animation_frame:
             threading.Timer(REFRESH_RATE_SEC, callback).start()
             self.display_scheduled = True
+            self.needs_animation_frame = False
         self.lock.release()
 ```
 
-To make use of this sytem, call `set_needs_animation_frame` from
-`set_needs_render` and `request_animation_frame_callback`:
+To make use of this sytem, call `set_tab_needs_animation_frame` from
+`set_needs_render`, and also from `request_animation_frame_callback`:
 
 ``` {.python}
 class Tab:
     def set_needs_render(self):
         self.needs_render = True
-        self.browser.set_needs_animation_frame()
+        self.browser.set_tab_needs_animation_frame(self)
 
     def request_animation_frame_callback(self):
         self.needs_raf_callbacks = True
-        self.browser.set_needs_animation_frame()
+        self.browser.set_tab_needs_animation_frame(self)
 ```
 
 And also in `Browser`:
@@ -1460,9 +1485,9 @@ class Browser:
             self.lock.acquire(blocking=True)
             scroll = self.scroll
             active_tab = self.tabs[self.active_tab]
-            self.lock.release()
             active_tab.task_runner.schedule_task(
                 Task(active_tab.run_animation_frame, scroll))
+            self.lock.release()
         # ...
 ```
 
@@ -1500,7 +1525,7 @@ class Tab:
     def run_animation_frame(self, scroll):
         # ...
         self.browser.commit(
-            self.url,
+            self, self.url,
             clamped_scroll if self.scroll_changed_in_tab \
                 else None, 
             document_height, self.display_list)
--- a/book/scheduling-and-threading.md
+++ b/book/scheduling-and-threading.md
@@ -415,66 +415,50 @@ not directly related to running JavaScript.
 Rendering pipeline tasks
 ========================
 
+So far we've focused on creating tasks that run JavaScript code. But
+the results of that JavaScript code---and also the results of
+interactions like loading new pages, scrolling, clicking, and
+typing---are only available to the user after the browser renders the
+page. In this sensem, the most important task in a browser is running
+the [rendering pipeline][graphics-pipline]: styling the HTML elements,
+constructing the layout tree, computing sizes and positions, painting
+layout objects to a display list, rastering the result into surfaces,
+and drawing those surfaces to the screen.
 
+Right now, the browser executes these rendering steps eagerly: as soon
+as the user scrolls or clicks, or as soon as JavaScript modifies the
+document. But we want to make these interactions faster and smoother,
+and the very first step is to make rendering a schedulable task, so we
+can decide when it occurs.
 
-One of the most expensive tasks a browser does is render a web
-page---style the HTML elements, construct a layout tree, compute sizes
-and positions, paint it to a display list, raster the result into
-surfaces, and draw tha surfaces to the screen. These rendering steps
-make up the [*rendering pipeline*][graphics-pipeline]
-for the browser, and can be expensive and slow. For this reason,
-modern browsers split the rendering pipeline across threads and make
-sure to run the rendering pipeline only when necessary.
-Those task queues also allow multiple
-threads to communicate by assigning each other additional tasks.
+At a high level, that requires code like this:
 
+``` {.python expected=False}
+self.task_runner.schedule_task(Task(self.render))
+```
 
-[graphics-pipeline]: https://en.wikipedia.org/wiki/Graphics_pipeline
-
-Refactoring our browser to think in terms of tasks will require
-significant changes throughout---concurrent programming is
-never easy! But these architectural changes are a key optimization
-behind modern browsers, and enable many advanced features discussed in
-this and later chapters.
-
-Everything in a browser can be considered a task, including rendering. In fact,
-the most important task in a browser is the rendering pipeline---but not just
-for the obvious reason that it's impossible to see web pages that aren't
-rendered. Most of the time spent doing work in a browser is in *rendering
-interactions* with the browser, such as loading, scrolling, clicking and
-typing. All of these interactions require rendering. If you want to make those
-interactions faster and smoother, the very first think you have to do is
-schedule the rendering pipeline, and to achieve that it'll need to be a
-schedulable task.
-
-On thing that is special about rendering is that it's a "singleton" task. There
-is only one rendering task, and either it's been scheduled or not, but it
-doesn't make sense to schedule it twice at the same time. On the other hand,
-it's totally fine and natural to have many `setTimeout` or `XMLHttpRequest`
-callbacks pending at the same time. The singleton nature of the rendering task
-has to do with there being only one DOM to render.
-
-Because it's a singleton, we'll need an additional boolean variable on a `Tab`
-to avoid having two rendering tasks, called `needs_animation_frame`. This
-variable means "a rendering task was already scheduled". Also add a
-`schedule_animation_frame`[^animation-frame] method that adds a new task to
-render, and a new method `run_animation_frame` as the task callback. Note
-how we avoid two scheduled frames with an if statement.
-
-[^animation-frame]: It's called an "animation frame" because sequential
-rendering of different pixels is an animation, and each time you render it's
-one "frame"---like a drawing in a picture frame.
-
+However, rendering is special in that it never makes sense to do
+scheduling twice in a row, since the page wouldn't have changed in
+between. To avoid having two rendering tasks we'll add a boolean
+called named `needs_animation_frame` to each `Tab` which indicates
+whether a rendering task is scheduled:
 
 ``` {.python expected=False}
 class Tab:
     def __init__(self, browser):
         # ...
         self.needs_animation_frame = False
+```
 
-    def set_needs_render(self):
-        self.schedule_animation_frame()
+A new `schedule_animation_frame`[^animation-frame] method will check
+the flag before scheduling a new rendering task:
 
+[^animation-frame]: It's called an "animation frame" because
+sequential rendering of different pixels is an animation, and each
+time you render it's one "frame"---like a drawing in a picture frame.
+
+``` {.python expected=False}
+class Tab:
     def schedule_animation_frame(self):
         if self.needs_animation_frame:
             return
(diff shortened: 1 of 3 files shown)


pavpanchekha:
Please review the text changes and I'll fix the lint failures. I'll do the other sections of Chapter 12 in a different PR.

book/scheduling-and-threading.md
```diff
@@ -93,137 +44,130 @@ class Task:
...
+        self.task_runner = TaskRunner()
 ```
 
-Of course, this is all pointless at the moment, since there aren't any tasks
-yet. Let's try out this infrastructure on a simple example: evaluating scripts
-after load. Currently the browser just evaluates them right away, but instead
-let's make script evaluationt a task.
+This is a pretty simplistic of choosing which task to run next, and
+real browsers have sophisticated *schedulers* which consider things
+like XXX.
```
chrishtr, line 89: ?

book/scheduling-and-threading.md
```diff
@@ -93,137 +44,130 @@ class Task:
...
+The point of a task is that it can be created at one point, and then
+run at some later point by a task runner of some kind.[^event-loop] In
+our browser, the task runner will store tasks in a first-in first-out
+queue:
 
-``` {.python expected=False}
+[^event-loop]: Event loops, which we discussed in [Chapter
+2](graphics.md#eventloop) and [Chapter
+11](visual-effects.md#sdl-creates-the-window) are one kind of task
+runner, where the tasks to run are provided by the operating system.
```
chrishtr, line 61: What other kind of task runners than event loops did you have in mind?

book/scheduling-and-threading.md
```diff
@@ -93,137 +44,130 @@ class Task:
...
+class Tab:
+    def __init__(self):
+        self.task_runner = TaskRunner()
 ```
 
-Of course, this is all pointless at the moment, since there aren't any tasks
-yet. Let's try out this infrastructure on a simple example: evaluating scripts
-after load. Currently the browser just evaluates them right away, but instead
-let's make script evaluationt a task.
+This is a pretty simplistic of choosing which task to run next, and
```
chrishtr, line 87:
```suggestion
First-in-first-out is a simplistic way of choosing which task to run next, and
```

book/scheduling-and-threading.md
```diff
@@ -93,137 +44,130 @@ class Task:
...
             header, body = request(script_url, url)
             self.task_runner.schedule_task(
                 Task(self.run_script, script_url, body))
 ```
 
-That's it! Now our browser will not run scripts until after `load` has completed
-and the event loop comes around again. But before continuing, let's consider
-why this change is interesting---is it anything other than just some
-refactoring?
+To run those tasks, we need to call the `run` method on our
```
chrishtr, line 113: I think this makes more sense above the example of loading tasks, since without it it's not clear when tasks would ever run.

book/scheduling-and-threading.md
```diff
@@ -93,137 +44,130 @@ class Task:
...
-when to run scripts. For example, it's easy to make a change to `TaskRunner` to
-only run one script per second, or to not run them at all during page load, or
-when a tab is not the active tab. This flexibility is quite powerful, and we
-can use it without having to dive into the guts of a `Tab` or how it loads web
-pages---all we'd have to do is implement a new `TaskRunner` heuristic.
+This change is nice---pages will load a bit faster---but there's more
+to it than that. Before this change, we no choice but to run scripts
+right away just as they were loaded. But now that running scripts is a
+`Task`, the task runner controls when it runs. It could run only one
+script per second, or only when the page is active, or only if there
```
chrishtr, line 131:
```suggestion
script per second, or at different rates for active and inactive pages, or only if there
```

chrishtr: (this is to distinguish from the previous paragraph, which already demonstrates only-active)

book/scheduling-and-threading.md
```diff
@@ -93,137 +44,130 @@ class Task:
...
 
 Timers and setTimeout
-====================
+=====================
 
-It's often the case that when adding a task, it's not to be run right away, but
-instead at some point in the future. One example is the
-[`setTimeout`][settimeout] JavaScript API, which provides a way to run a
-function a given number of milliseconds from now. For example,
+Tasks are a natural way to support several JavaScript APIs that ask
```
chrishtr, line 138:
```suggestion
Tasks are *also* a natural way to support several JavaScript APIs that ask
```

book/scheduling-and-threading.md
```diff
@@ -93,137 +44,130 @@ class Task:
...
 ``` {.javascript expected=false}
-function callback() {
-    console.log('Callback')
-}
+function callback() { console.log('Callback'); }
 setTimeout(callback, 1000);
 ```
-will print "Callback" to the console log one second from now.
 
-This API *could* be implemented by recording a time associated with each new
```
chrishtr, line 192: I guess you didn't like this point? The aim was to motivate why we should use these fancy thread things instead of polling, which is probably what a programmer unfamiliar with threads would have attempted.

book/scheduling-and-threading.md
```diff
@@ -93,137 +44,130 @@ class Task:
...
 
-The `Timer` class lets you run a callback at a specified time in the future. It
-takes two parameters: a time delta in seconds from now, and a function to call
-when that time expires. The following code will run `callback` 10 seconds in
-the future on a new Python thread:
-
-``` {.python expected=False}
-threading.Timer(10, callback).start()
-```
+This runs `callback` in 1 seconds on a new Python thread. Now, it's
```
chrishtr, line 163:
```suggestion
This runs `callback` 1 second from now on a new Python thread. Now, it's
```

book/scheduling-and-threading.md
```diff
@@ -93,137 +44,130 @@ class Task:
...
+going to be a little tricky to use `Timer` to implement `setTimeout`
+due to the fact that multiple threads will be involved.
```
chrishtr, line 165:
```suggestion
due to the fact that multiple threads will be involved, but it's worth it.
```

chrishtr: This again is related to the polling sentence you deleted and my comment on it..

book/scheduling-and-threading.md
```diff
@@ -233,24 +177,26 @@ function setTimeout(callback, time_delta) {
     SET_TIMEOUT_REQUESTS[handle] = callback;
     call_python("setTimeout", handle, time_delta)
 }
+```
+
+The exported `setTimeout` function will create a timer, wait for the
+requested time period, and then call back into JavaScript to run the
+callback. The call back can call this `__runSetTimeout` function:
```
chrishtr, line 184:
```suggestion
callback. This happens via the `__runSetTimeout` function:
```

book/scheduling-and-threading.md
```diff
@@ -233,24 +177,26 @@ function setTimeout(callback, time_delta) {
     SET_TIMEOUT_REQUESTS[handle] = callback;
     call_python("setTimeout", handle, time_delta)
 }
+```
+
+The exported `setTimeout` function will create a timer, wait for the
+requested time period, and then call back into JavaScript to run the
```
chrishtr, line 183:
```suggestion
requested time period, and inform JavaScript to run the
```

book/scheduling-and-threading.md
```diff
@@ -233,24 +177,26 @@ function setTimeout(callback, time_delta) {
...
-
-This is easy to fix by using the `TaskRunner` you already implemented: instead
-of running the script right away, schedule a task to do it later, when the
-other thread is free. Here's the code:
+The Python side, however, is quite a bit more complex, because
+`threading.Timer` executes its callback *on a new Python thread*. That
+thread can't just call `evaljs` directly: we'll end up with JavaScript
+running on two Python threads at the same time, which is not
+ok.[^js-thread] Instead, the timer will have to merely add a new
+`Task`, which the main thread will execute, to call the callback:
```
chrishtr, line 199:
```suggestion
`Task`, which the primary thread will execute, to call the callback:
```

chrishtr: (can't say "main" or it would be inconsistent with later in the chapter)

book/scheduling-and-threading.md
```diff
@@ -274,25 +220,20 @@ class JSContext:
...
-from or writing to a data structure shared across threads, acquire the lock;
-after you're done, release it.^[The `blocking` parameter to `acquire` indicates
-whether the thread should wait for the lock to be available before continuing;
-in this chapter you'll always set it to true. (When the thread is waiting, it's
-said to be *blocked*.)] The code changes in `TaskRunner` are pretty easy---just
-be careful to not forget to release the lock, and hold it for the minimum time
-possible, so as to maximize thread parallelism. That's why the code releases
-the lock before calling `task`: after the task has been removed from the queue,
-it can't be accessed by another thread.
+Now only the main thread will call `evaljs`, which is better. But now
```
chrishtr, line 223:
```suggestion
Now only the proimary thread will call `evaljs`,. But now
```

book/scheduling-and-threading.md
```diff
@@ -93,137 +44,130 @@ class Task:
...
+to it than that. Before this change, we no choice but to run scripts
+right away just as they were loaded. But now that running scripts is a
+`Task`, the task runner controls when it runs. It could run only one
+script per second, or only when the page is active, or only if there
+isn't a higher-priority user action to respond to. A browser could
+even have multiple task runners, optimized for different use cases.
 
 Timers and setTimeout
-====================
+=====================
```
chrishtr, line 136: Regarding whether we should keep setTimeout in the book: I've concluded that we should, for two reasons:
* the one I mentioned yesterday about introducing timers in a less complicated setting
* as a more gentle way to introduce asyc javascript tasks and prepare for rAF

book/scheduling-and-threading.md
chrishtr: I do like how you tried to use fewer words and more of a summary explanation of the runtime binding code though, to minimize the overhead.

book/scheduling-and-threading.md
```diff
@@ -327,116 +275,129 @@ situation---some devices have more [CPU cores][cores] than others, or are more
...
-die.
-
-In this example:
-``` {.python expected=False}
-def run:
-    count = 0
-    while count < 100:
-        print("Thread")
-        count += 1
+Threads can also be used to have the browser multitask. For example,
```
chrishtr, line 283:
```suggestion
Threads can also be used to add browser multitasking. For example,
```

book/scheduling-and-threading.md
```diff
@@ -327,116 +275,129 @@ situation---some devices have more [CPU cores][cores] than others, or are more
...
+in [Chapter 10](security.md#cross-site-requests) we implemented the
+`XMLHttpRequest` class, which lets scripts make requests to other
+websites. But in our implementation, the whole browser would seize up
+while waiting for the request to finish. That's obviously bad.^[For
+this reason, the synchronous version of the API that we implemented in
+Chapter 10 is basically useless and a huge performance footgun. Some
+browsers are now moving to deprecate this API.]
```
chrishtr, line 290:
```suggestion
browsers are now moving to deprecate the sync version of this API.]
```

book/scheduling-and-threading.md
```diff
@@ -327,116 +275,129 @@ situation---some devices have more [CPU cores][cores] than others, or are more
...
-Our approach will be to start a thread, run some code on it that does the
-request and gets a response, then schedule a `Task` to send the response back
-to the script. Starting a thread is easy: define a function that is the "main"
-function of the thread, then start the thread with the function passed as the
-`target` argument. When the main function exits, the thread will automatically
-die.
-
-In this example:
-``` {.python expected=False}
-def run:
```
chrishtr, line 352: This example is too complicated, you think?

book/scheduling-and-threading.md
```diff
@@ -327,116 +275,129 @@ situation---some devices have more [CPU cores][cores] than others, or are more
...
+``` {.python}
+class JSContext:
+    def XMLHttpRequest_send(self, method, url, body, is_async, handle):
+        # ...
         def run_load():
-            headers, out = request(
+            headers, body = request(
                 full_url, self.tab.url, payload=body)
-            handle_local = handle
             self.tab.task_runner.schedule_task(
```
chrishtr, line 362: I think there is my bug in the code here - should only schedule a task if is_async is true?

book/scheduling-and-threading.md
```diff
@@ -327,116 +275,129 @@ situation---some devices have more [CPU cores][cores] than others, or are more
...
-
-* Allow `is_async` to be `true` in the constructor of `XMLHttpRequest` in the
-runtime.
-* Store a unique handle for each `XMLHttpRequest` object, analogous to how we
-  used handles for `Node`s.
-* Store a map from request handle to object.
+Note that in the async case, the `XMLHttpRequest_send` method starts a
+thread and then immediately returns. That thread will run in parallel
+to the browser's main work until it adds a new task for running
+`dispatch_xhr_online` on the main thread. This method runs the
```
chrishtr, line 385:
```suggestion
`dispatch_xhr_onload` on the main thread. This method runs the
```

book/scheduling-and-threading.md
```diff
@@ -446,51 +407,58 @@ function __runXHROnload(body, handle) {
...
-
-[^animation-frame]: It's called an "animation frame" because sequential
-rendering of different pixels is an animation, and each time you render it's
-one "frame"---like a drawing in a picture frame.
+So far we've focused on creating tasks that run JavaScript code. But
+the results of that JavaScript code---and also the results of
+interactions like loading new pages, scrolling, clicking, and
+typing---are only available to the user after the browser renders the
+page. In this sensem, the most important task in a browser is running
+the [rendering pipeline][graphics-pipline]: styling the HTML elements,
```
chrishtr, line 423:
```suggestion
the [rendering pipeline][graphics-pipleine]: styling the HTML elements,
```

book/scheduling-and-threading.md
```diff
@@ -446,51 +407,58 @@ function __runXHROnload(body, handle) {
...
+constructing the layout tree, computing sizes and positions, painting
+layout objects to a display list, rastering the result into surfaces,
+and drawing those surfaces to the screen.
+
+Right now, the browser executes these rendering steps eagerly: as soon
+as the user scrolls or clicks, or as soon as JavaScript modifies the
+document. But we want to make these interactions faster and smoother,
+and the very first step is to make rendering a schedulable task, so we
```
chrishtr, line 431:
```suggestion
and the very first step in doing so is to make rendering a schedulable task, so we
```

book/scheduling-and-threading.md
```diff
@@ -446,51 +407,58 @@ function __runXHROnload(body, handle) {
...
+At a high level, that requires code like this:
+
+``` {.python expected=False}
+self.task_runner.schedule_task(Task(self.render))
+```
 
+However, rendering is special in that it never makes sense to do
+scheduling twice in a row, since the page wouldn't have changed in
+between. To avoid having two rendering tasks we'll add a boolean
+called named `needs_animation_frame` to each `Tab` which indicates
```
chrishtr, line 443:
```suggestion
called `needs_animation_frame` to each `Tab` which indicates
```

book/scheduling-and-threading.md
```diff
@@ -502,16 +470,53 @@ class Tab:
...
+As can `innerHTML_set`:
+
+``` {.python}
+class JSContext:
+    def innerHTML_set(self, handle, s):
+        # ...
+        self.tab.set_needs_render()
+```
+
+There are more calls to `render`; you should find and fix all of them.
```
chrishtr, line 505: This repeats the footnote on line 487.

book/scheduling-and-threading.md
```diff
@@ -502,16 +470,53 @@ class Tab:
...
+
+So this handles the front half of the rendering pipeline: style,
+layout, and paint. The back half of the rendering pipeline (raster and
+draw) is handled by `Browser`, so the `Tab` needs to tell the
+`Browser` when to run it. I'll add a new `raster_and_draw` method for
```
chrishtr, line 510:
```suggestion
`Browser` to run it. I'll add a new `raster_and_draw` method for
```

book/scheduling-and-threading.md
pavpanchekha: Fixed, I meant to refer to the two specific event loops we've seen. Changed to:

> The event loops we discussed in Chapter 2 and Chapter 11 are task runners, where the tasks to run are provided by the operating system.

book/scheduling-and-threading.md
pavpanchekha: Done

book/scheduling-and-threading.md
pavpanchekha: Ah, yes, I was hoping you could tell me what kinds of things the Chrome scheduler, for example, considers.

book/scheduling-and-threading.md
pavpanchekha: Oops, I meant to move this to a footnote and it got lost somewhere. I've resurrected the footnote. The point is good, but I think the first description should be a "positive" one.

book/scheduling-and-threading.md
pavpanchekha: This is good whether or not polling is presented first!

book/scheduling-and-threading.md
pavpanchekha: Great catch!

book/scheduling-and-threading.md
pavpanchekha: Ok

book/scheduling-and-threading.md
pavpanchekha: Yeah. I think there's also a bit of a question here of defining our audience. If they haven't seen threads at all before, will they be able to follow the description here? It also introduces tricky questions like whether the console out is thread-safe, which we don't actually need to raise.

book/scheduling-and-threading.md
pavpanchekha: You're right. Fixed. Do you think we should add some code to default `is_async` to `true` on the JavaScript side? Otherwise if you don't pass the value we will assume it is sync. Argument against is just that the code is super ugly since DukPy doesn't support ES6. Maybe we footnote it?

book/scheduling-and-threading.md
pavpanchekha: Ah, woops, I meant to move this footnote to the main text.

pavpanchekha:
Uhh, quick question as I think about the is_async bug: if you create a synchronous XHR, does it run onload()? When? I looked at the standard and the plain reading seems to be that you must run onload, synchronously, before returning, but the standard also doesn't specify that the body is returned, so you get the sense that the synchronous case wasn't worked out in too much detail (no surprise!). And I can see an alternative reading of the standard where it's valid to fire onload() asynchronously, though I think this reading is more tenuous.

book/scheduling-and-threading.md
```diff
@@ -93,137 +43,139 @@ class Task:
...
+class Tab:
+    def __init__(self):
+        self.task_runner = TaskRunner()
 ```
 
-Finally we just need to modify the main event loop to run the task runner each
-time, in case there is a script task to execute:
+First-in-first-out is a simplistic way to choose which task to run
+next, and real browsers have sophisticated *schedulers* which consider
+things like XXX.
```
chrishtr, line 88:
```suggestion
many different factors.
```

chrishtr: There is also this link, but it's kind of old and weak sauce:

chrishtr: https://blog.chromium.org/2015/04/scheduling-tasks-intelligently-for_30.html

book/scheduling-and-threading.md
chrishtr: Let's go with a footnote.

chrishtr:
> Uhh, quick question as I think about the is_async bug: if you create a synchronous XHR, does it run onload()? When? I looked at the standard and the plain reading seems to be that you must run onload, synchronously, before returning, but the standard also doesn't specify that the body is returned, so you get the sense that the synchronous case wasn't worked out in too much detail (no surprise!). And I can see an alternative reading of the standard where it's valid to fire onload() asynchronously, though I think this reading is more tenuous.

Let's just go with not calling onload? Or I guess we can leave the code as-is and technically it's more correct w.r.t. the spec.

chrishtr (approved):

book/scheduling-and-threading.md
pavpanchekha: I'll include the link, better than nothing.

PR #358, opened 2022-02-06
Fix chapter 12 concurrency bugs

chrishtr:
* Don't commit display lists for non-active tabs
* Make sure locks surround all accesses to shared data structures
* Don't set display_scheduled to false until an actual commit. (Otherwise multiple animation frames could be scheduled at once.)


chrishtr:
Going to merge this now since it blocks another PR I'm working on to revise the profiling section.

PR #362, opened 2022-02-09
Pass over threads section of Chapter 12

pavpanchekha:
This PR makes a couple of big changes to Chapter 12, from the profiling section through threading and threaded scrolling.

- Does all the profiling before introducing the idea of threaded rendering.
- Moves "slow scripts" section to be right before threaded scrolling, which it motivates.
- Makes `set_needs_animation_frame` the one that checks tab identity, and makes calls from `Browser` directly modify the variable.
- Replaces the many arguments to `commit` with a new `CommitForRaster` object, and add some text about ownership.

### DIFF
--- a/book/scheduling-and-threading.md
+++ b/book/scheduling-and-threading.md
@@ -831,27 +831,17 @@ behavior when animating. Once you've read the whole chapter and implemented
 that thread, try removing this dirty bit and see for yourself!
 :::
 
-Parallel rendering
-==================
+Profiling rendering
+===================
 
-What happens if rendering takes much more than 16ms to finish and we're out of
-ideas for how to make it run more efficiently? If it's a rendering task that's
-slow, such as font loading (see [chapter 3][faster-text-loading]), if we're
-lucky we can make it faster. But sometimes it's not possible to make the code a
-lot faster, it just has a lot to do. In rendering, this could be because the
-web page is very large or complex.
+We now have a system for scheduling a rendering task every 16ms. But
+what if rendering takes longer than 16ms to finish? Before we answer
+this question, let's instrument the browser and measure how much time
+is really being spent rendering. It's important to always measure
+before optimizing, because the result often surprising.
 
-[faster-text-loading]: text.md#faster-text-layout
-
-What if we ran raster and draw *in parallel* with the main thread, by using CPU
-parallelism? After all, they take as input only the display list and not the
-DOM.
-
-That sounds fun to try, but before adding such complexity, let's
-instrument the browser and measure how much time is really being spent
-in raster and draw.^[Pro tip: always measure before optimizing. You'll
-often be surprised at where the bottlenecks are.] We'll want to
-average across multiple raster-and-draw cycles:
+Let's implement some simple instrumentation to measure time. We'll
+want to average across multiple raster-and-draw cycles:
 
 ``` {.python}
 class MeasureTime:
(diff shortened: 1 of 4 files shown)


book/scheduling-and-threading.md
```diff
@@ -1199,13 +1112,34 @@ class Browser:
...
+            active_tab.task_runner.schedule_task(task)
         self.lock.release()
 ```
 
-Now let's go the other direction, and implement the `commit` method that
-copies the display list, url and scroll offset to the browser thread.
+Do the same with any other calls from the `Browser` to the `Tab`.
+
+Communication in the other direction is a little subtler. We already
+have the `set_needs_animation_frame` variable, but now we need the
```
chrishtr, line 1123:
```suggestion
have the `set_needs_animation_frame` method, but now we need the
```

book/scheduling-and-threading.md
```diff
@@ -1199,13 +1112,34 @@ class Browser:
...
+`Tab` to call `commit` when it's finished a display list.
```
chrishtr, line 1124:
```suggestion
`Tab` to call `commit` when it's finished creating a display list.
```

book/scheduling-and-threading.md
```diff
@@ -1215,91 +1149,119 @@ class Tab:
 
     def run_animation_frame(self):
         self.render()
+        commit = CommitForRaster(
```
chrishtr, line 1152:
```suggestion
        commit_data = CommitForRaster(
```

book/scheduling-and-threading.md
```diff
@@ -1215,91 +1149,119 @@ class Tab:
...
+            url=self.url,
+            scroll=self.scroll,
+            height=document_height,
+            display_list = display_list
```
chrishtr, line 1156: display_list.copy() for obvious thread safety?

book/scheduling-and-threading.md
```diff
@@ -1336,93 +1298,73 @@ class Browser:
...
+``` {.javascript file=eventloop}
+function callback() {
+    for (var i = 0; i < 1e9; i++);
+    // ...
+}
+```
+
+Now, every tick of the counter has an artificial pause during which
+the main thread is stuck running JavaScript. This means it can't
+respond to any events; for example, if you hold down the down array,
```
chrishtr, line 1348:
```suggestion
respond to any events; for example, if you hold down the down key,
```

book/scheduling-and-threading.md
```diff
@@ -1336,93 +1298,73 @@ class Browser:
...
+the scrolling will be janky and annoying. To fix this, we need to move
+scrolling from the main thread to the browser thread.
```
chrishtr, line 1350: "I encourage you to try it for yourself and witness the annoyance."

This point is important IMO.

book/scheduling-and-threading.md
```diff
@@ -1437,25 +1379,27 @@ class Browser:
...
-The `set_active_tab` method is an interesting case, because it causes
-scroll to be set back to 0, but we need the main thread to run in order
-to commit a new display list for the other tab. That's why this method doesn't
-call `set_needs_raster_and_draw`.
+This code sets `needs_raster_and_draw` to apply the new scroll offset.
+
+The scroll offset also needs to change when the user switches tabs,
+but in this case we don't know the right scroll offset yet. We need
+the main thread to run in order to commit a new display list for the
+other tab, and at that point we will have a new scroll offset as well.
```
chrishtr, line 1387: Good point, I missed this.

chrishtr (commented):
I like these improvements. The chapter keeps getting tighter with each iteration.

book/scheduling-and-threading.md
```diff
@@ -928,181 +918,112 @@ class Browser:
...
-browser thread will be for:
+Two threads
+===========
 
-[^also-compositor]: Th browser thread is similar to what modern browsers often
-call the [*compositor thread*][cc].
+Running rendering in parallel with raster and draw would allow us to
+produce a new frame every 66ms, instead of every 88ms. But more
+importantly, since there's no point to running render more often than
+raster-and-draw, the render thread would have 46ms left over for run
```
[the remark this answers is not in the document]
pavpanchekha: Hmm, I think this is maybe too confusing. I meant that for each animation frame it would have 20ms of rendering time and 46ms of JS time. I'll rewrite.

book/scheduling-and-threading.md
pavpanchekha: Woops!

book/scheduling-and-threading.md
pavpanchekha: I thought about this and I don't think it's right. The display list, after all, is now a tree structure, so while `.copy()` looks like it might be thread safe it's actually not. Moreover, we actually don't expect the main thread to ever access the display list after making it (in a real browser that might not be true—perhaps this needs a footnote) so better to explain that logic directly. (A better implementation would avoid assigning the display list to a variable at all, but I think that would require a refactor which doesn't seem worth it.)

chrishtr: Good point!

chrishtr (approved):

PR #368, opened 2022-02-11
Invert the order of rendering sections

pavpanchekha:
This PR changes around the order of the rendering sections. It now introduces the steady rendering cadence _first_, then introduces the dirty bits, and sets up `needs_animation_frame` once we start talking about `requestAnimationFrame`.

The other big change is to set up `schedule_animation_frame` on the `Browser` from the beginning, which means we need to change less in the two-threads section. That makes that section a little easier.

### DIFF
--- a/book/scheduling.md
+++ b/book/scheduling.md
@@ -1275,7 +1275,7 @@ do that inside `commit`.
 On the `Browser` side, the new `commit` method needs to read out all
 of the data it was sent and call `set_needs_raster_and_draw` as
 needed. Because this call will come from another thread, we'll need to
-take a lock:
+acquire a lock:
 
 ``` {.python}
 class Browser:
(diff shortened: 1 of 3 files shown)


book/scheduling.md
```diff
@@ -1207,25 +1196,41 @@ class Tab:
...
+```
+
+Also, remove the `set_needs_raster_and_draw` call from `render`. We'll
+do that inside `commit`.
 
 On the `Browser` side, the new `commit` method needs to read out all
 of the data it was sent and call `set_needs_raster_and_draw` as
-needed.
+needed. Because this call will come from another thread, we'll need to
+take a lock:
```
chrishtr, line 1219:
```suggestion
acquire the lock:
```

chrishtr (approved):

PR #369, opened 2022-02-11
Minor changes in Chapter 12

pavpanchekha:
These are mostly small wording or code style things, but it also fixes an issue that could occur if you had multiple `setTimeout` requests that fired in a different order from how they were created.

### DIFF
--- a/book/scheduling.md
+++ b/book/scheduling.md
@@ -169,9 +169,9 @@ Timers and setTimeout
 
 Tasks are *also* a natural way to support several JavaScript APIs that
 ask for a function to be run at some point in the future. For example,
-the [`setTimeout`][settimeout] JavaScript API lets you run a function
-some number of milliseconds from now. This code prints "Callback" to
-the console one minute in the future, for example:
+[`setTimeout`][settimeout] lets you run a JavaScript function some
+number of milliseconds from now. This code prints "Callback" to the
+console one second frow now:
 
 [settimeout]: https://developer.mozilla.org/en-US/docs/Web/API/setTimeout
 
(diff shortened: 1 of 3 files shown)


book/scheduling.md
```diff
@@ -383,7 +384,7 @@ do security checks:
 
 ``` {.python}
 class JSContext:
-    def XMLHttpRequest_send(self, method, url, body, is_async, handle):
+    def XMLHttpRequest_send(self, method, url, body, isasync, handle):
```
chrishtr, line 394: Why this change?

chrishtr:
Which line of code fixes the bug you mentioned?

pavpanchekha:
The line that fixes the bug is in the JavaScript, where the line `delete XHR_REQUESTS[handle]` would remove a key, which would potentially make the handle assigned to the next XHR be a handle that had already been assigned. By removing the `delete` line I've introduced a potential memory leak, but we already had memory leaks (and they're called out in the JS chapter) so I think it's OK. (Alternatively, we could fix it by introducing a real counter... I'll at least throw up a footnote with a warning.)

pavpanchekha:
Added a footnote earlier, when discussing `setTimeout`, since it has a similar issue.

chrishtr:
> The line that fixes the bug is in the JavaScript, where the line `delete XHR_REQUESTS[handle]` would remove a key, which would potentially make the handle assigned to the next XHR be a handle that had already been assigned. By removing the `delete` line I've introduced a potential memory leak, but we already had memory leaks (and they're called out in the JS chapter) so I think it's OK. (Alternatively, we could fix it by introducing a real counter... I'll at least throw up a footnote with a warning.)

Does the javascript runtime need to be in this PR then?

pavpanchekha:
Please see latest commit message :)

chrishtr (approved):

book/scheduling.md
pavpanchekha: lol, these code blocks were one character too wide, and I didn't want to spill the arguments over two lines.

PR #372, opened 2022-02-13
Chapter 12 final pass

pavpanchekha:
This final pass does a lot of small wording changes, and moves around a bunch of go-furthers so they are 1:1 with sections. It also:

- Adds a go-further on refresh rates
- Merges two go-furthers on allocating event loops to cores
- Adds a section, "Committing display lists", to split the mega-section into two

### DIFF
--- a/book/scheduling.md
+++ b/book/scheduling.md
@@ -295,25 +295,18 @@ held while the task is running.
 
 ::: {.further}
 Unfortunately, Python currently has a [global interpreter lock][gil],
-so Python threads don't truly run in parallel.[^why-gil] Threading in
-Python therefore doesn't increase *throughput*, though it can increase
-*responsiveness*. This is an unfortunate limitation of Python that
-doesn't affect real browsers, so in this chapter just try to pretend
-the GIL isn't there.[^why-locks]
+so Python threads don't truly run in parallel. This is an unfortunate
+limitation of Python that doesn't affect real browsers, so in this
+chapter just try to pretend the GIL isn't there. : Despite the global
+interpreter lock, we still need locks. Each Python thread can yield
+between bytecode operations, so you can still get concurrent accesses
+to shared variables, and race conditions are still possible. And in
+fact, while debugging the code for this chapter, I often encountered
+this kind of race condition when I forgot to add a lock; try removing
+some of the locks from your browser to see for yourself!
 :::
 
 [gil]: https://wiki.python.org/moin/GlobalInterpreterLock
-
-[^why-gil]: It's possible to turn off the global interpreter lock
-while running foreign C/C++ code linked into a Python library. Skia is
-thread-safe, but DukPy and SDL may not be.
-
-[^why-locks]: Despite the global interpreter lock, we still need locks. Each
-Python thread can yield between bytecode operations, so you can still get
-concurrent accesses to shared variables, and race conditions are still
-possible. And in fact, while debugging the code for this chapter, I often
-encountered this kind of race condition when I forgot to add a lock; try
-removing some of the locks from your browser to see for yourself!
 
 Long-lived threads
 ==================
@@ -645,7 +638,13 @@ But sometimes that task is just running JavaScript that doesn't touch
 the web page, and the `raster_and_draw` call is a waste.
 
 We can avoid this using another dirty bit, which I'll call
-`needs_raster_and_draw`:
+`needs_raster_and_draw`:[^not-just-speed]
+
+[^not-just-speed]: The `needs_raster_and_draw` dirty bit doesn't just
+make the browser a bit more efficient. Later in this chapter, we'll
+add multiple browser threads, and at that point this dirty bit is
+necessary to avoid erratic behavior when animating. Try removing it
+later and see for yourself!
 
 ``` {.python}
 class Browser:
@@ -1384,9 +1383,15 @@ And that's it: we should now be doing render on one thread and raster
 and draw on another!
 
 ::: {.further}
-The `needs_raster_and_draw` dirty bit doesn't just make the browser a
-bit more efficient. Without this bit, our threads would have erratic
-behavior when animating. Try removing it and see for yourself!
+Due to the Python GIL, threading in Python therefore doesn't increase
+*throughput*, but it can increase *responsiveness* by, say,
+running JavaScript tasks on the main thread while the browser does
+raster and draw. It's also possible to turn off the global interpreter
+lock while running foreign C/C++ code linked into a Python library;
+Skia is thread-safe, but DukPy and SDL may not be, and don't seem to
+release the GIL. If they did, then JavaScript or raster-and-draw truly
+could run in parallel with the rest of the browser, and performance
+would improve as well.
 :::
 
 
--- a/book/scheduling.md
+++ b/book/scheduling.md
@@ -1221,7 +1221,7 @@ Communication in the other direction is a little subtler.
 
 Originally, threads were a mechanism for improving *responsiveness*
 via pre-emptive multitasking, not *throughput* (frames per second).
-Nowadays, even phones have several cores plus a highly parallel GPU,
+Nowadays, though, even phones have several cores plus a highly parallel GPU,
 and threads are much more powerful. It's therefore useful to
 distinguish between conceptual events; event queues and dependencies
 between them; and their implementation on a computer architecture.
@@ -1519,11 +1519,14 @@ so, and only store the browser thread's scroll offset if
 
 [^scroll-complicated]: Two-threaded scroll has a lot of edge cases,
 including some I didn't anticipate when writing this chapter. For
-example, it's pretty clear that a load should force scroll to 0, but
+example, it's pretty clear that a load should force scroll to 0
+(unless the browser implements [scroll restoration][scroll-restoration]!), but
 what about a scroll clamp followed by a browser scroll that brings it
 back to within the clamped region? By splitting the browser into two
 threads, we've brought in all of the challenges of concurrency and
 distributed state.
+
+[scroll-restoration]: https://developer.mozilla.org/en-US/docs/Web/API/History/scrollRestoration
 
 ``` {.python}
 class Tab:
--- a/book/scheduling.md
+++ b/book/scheduling.md
@@ -166,8 +166,8 @@ this:[^polling]
 `Task` is supposed to occur, and compare against the current time in
 the event loop. This is called *polling*, and is what, for example,
 the SDL event loop does to look for events and tasks. However, that
-can mean wasting time in a loop until the task to be ready, so I
-expect the `Timer` to be more efficient.
+can mean wasting time in a loop until the task is ready, so I expect
+the `Timer` to be more efficient.
 
 [timer]: https://docs.python.org/3/library/threading.html#timer-objects
 [threading]: https://docs.python.org/3/library/threading.html
@@ -295,13 +295,11 @@ held while the task is running.
 
 ::: {.further}
 Unfortunately, Python currently has a [global interpreter lock][gil],
-so our two Python threads don't truly run in parallel,[^why-gil]
-and our browser's *throughput* won't increase much from multi-threading.
-Nonetheless, the *responsiveness* of the browser thread is still
-massively improved, since it often isn't blocked on JavaScript or the
-front half of the rendering pipeline. This is an unfortunate
-limitation of Python that doesn't affect real browsers, so try to
-pretend it's not there.[^why-locks]
+so Python threads don't truly run in parallel.[^why-gil] Threading in
+Python therefore doesn't increase *throughput*, though it can increase
+*responsiveness*. This is an unfortunate limitation of Python that
+doesn't affect real browsers, so in this chapter just try to pretend
+the GIL isn't there.[^why-locks]
 :::
 
 [gil]: https://wiki.python.org/moin/GlobalInterpreterLock
@@ -461,12 +459,12 @@ to running JavaScript.
 `XMLHttpRequest` played a key role in helping the web evolve. In the
 90s, clicking on a link or submitting a form required loading a new
 pages. With `XMLHttpRequest` web pages were able to act a whole lot
-more like a dynamic application, with GMail one famous early
+more like a dynamic application; GMail was one famous early
 example.[^when-gmail] Nowadays, a web application that uses DOM
 mutations instead of page loads to update its state is called a
-[single-page app][spa]. SPAs enabled more interactive and complex web
-apps, which in turn made browser speed and responsiveness more
-important.
+[single-page app][spa]. Single-page apps enabled more interactive and
+complex web apps, which in turn made browser speed and responsiveness
+more important.
 
 [^when-gmail]: GMail dates from April 2004, [soon after][xhr-history]
 enough browsers finished adding support for the API. The first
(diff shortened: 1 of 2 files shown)


pavpanchekha:
I think once this PR is merged, we should do a PR to release the chapter (this just requires toggling a bit in `config.json`) and then we're set to publish on Monday!

book/scheduling.md
```diff
@@ -167,7 +166,7 @@ this:[^polling]
 `Task` is supposed to occur, and compare against the current time in
 the event loop. This is called *polling*, and is what, for example,
 the SDL event loop does to look for events and tasks. However, that
-can mean wasting time in a loop waiting for the task to be ready, so I
+can mean wasting time in a loop until the task to be ready, so I
```
chrishtr, line 169:
```suggestion
can mean wasting time in a loop until the task is ready, so I
```

book/scheduling.md
```diff
@@ -293,18 +294,28 @@ can't be accessed by another thread, so the lock does not need to be
...
-Event loops often map 1:1 to CPU threads within a single CPU process, but
-this is not required. For example, multiple event loops could be placed together
-on a single CPU thread with yet another scheduler on top of them that
-round-robins between them. It's useful to distinguish between: conceptual
-events, event queues and dependencies between them; and their implementation in
-a computer architecture. This way, the browser implementer (you!) has maximum
-ability to use more or less hardware parallelism as appropriate to the
-situation---some devices have more [CPU cores][cores] than others, or are more
-sensitive to battery power usage.
+Unfortunately, Python currently has a [global interpreter lock][gil],
```
chrishtr, line 297: Isn't this go-further premature here, since we haven't yet talked about performance and parallelism?
Also, have we introduced the rendering pipeline at this point?

book/scheduling.md
```diff
@@ -450,36 +458,36 @@ to running JavaScript.
...
-async version of this API, web pages were able to act a whole lot more like
-an *application* than a page of information. This ushered in a new generation
-of web sites that used this technique; GMail is one famous early example that
-dates from April 2004, [soon after][xhr-history] enough browsers finished adding
-support for the API.
-
+`XMLHttpRequest` played a key role in helping the web evolve. In the
+90s, clicking on a link or submitting a form required loading a new
+pages. With `XMLHttpRequest` web pages were able to act a whole lot
+more like a dynamic application, with GMail one famous early
```
chrishtr, line 464:
```suggestion
more like a dynamic application; GMail was one famous early
```

book/scheduling.md
```diff
@@ -450,36 +458,36 @@ to running JavaScript.
...
+example.[^when-gmail] Nowadays, a web application that uses DOM
+mutations instead of page loads to update its state is called a
+[single-page app][spa]. SPAs enabled more interactive and complex web
```
chrishtr, line 467:
```suggestion
[single-page app][spa]. Single-page apps enabled more interactive and complex web
```

book/scheduling.md
```diff
@@ -450,36 +458,36 @@ to running JavaScript.
...
 The cadence of rendering
 ========================
 
-There's more to tasks than just implementing some JavaScript APIs. Once
-something is a `Task`, the task runner controls when it runs: perhaps now,
-later, at most once a second, at different rates for active and inactive pages,
-or according to its priority. A browser could even have multiple task runners,
-optimized for different use cases.
+There's more to tasks than just implementing some JavaScript APIs.
+Once something is a `Task`, the task runner controls when it runs:
```
chrishtr, line 487: You literally undid the edit I added yesterday. I added "perhaps" because it was unclear whether the use of "now" is in the "now, let me tell you something..." meaning vs "right this moment"

book/scheduling.md
```diff
@@ -557,12 +562,29 @@ if __name__ == "__main__":
...
+        browser.schedule_animation_frame()
 ```
 
 Now we're scheduling a new rendering task every 16 milliseconds, just
 as we wanted to.
 
-Optimizing rendering tasks
+::: {.further}
+
+there's nothing special about 60 frames per second. Many displays
```
chrishtr, line 573:
```suggestion
there's nothing special about 60 frames per second. Some displays
```

book/scheduling.md
```diff
@@ -557,12 +562,29 @@ if __name__ == "__main__":
...
+refresh 72 times per second, and some that [refresh even more
+often][refresh-rate] are becoming more common. Movies are often shot
+in 24 frames per second (though [some directors advocate
+48][hobbit-fps]) while television shows often use 30 frames per
```
chrishtr, line 577:
```suggestion
48][hobbit-fps]) while television shows traditionally use 30 frames per
```

book/scheduling.md
```diff
@@ -897,23 +907,16 @@ function callback() {
...
-but then a script click event handler occurs before *actually* rendering, the
-app might be forced to re-do its DOM mutations to avoid a delayed response to
-the click---yet another example of doubled rendering.
-
-A third is that there is no great way to turn off "rendering" `setTimeout` work
-when a web page tab or window is backgrounded, minimized or otherwise
-throttled. If the browser chooses to stop all tasks, then it would also break
-any important background work the web app might want to do (such as syncing
-your information to the server so it's not lost).
+This sort of worked, there was no guarantee that the callbacks would
```
chrishtr, line 910:
```suggestion
This sort of worked, but there was no guarantee that the callbacks would
```

book/scheduling.md
```diff
@@ -1024,57 +1027,40 @@ raster-and-draw step by adopting a multi-threaded architecture.
...
-`tab_surface` smaller also helps a lot.
-
-Modern browsers go a step further and perform raster and
-draw [on the GPU][skia-gpu], where a lot more parallelism is
-available. Even so, on complex pages raster and draw really do
-sometimes take a lot of time. I'll dig into this more in
+
+Our toy browser spends a lot of time copying pixels. That's why
+[optimizing surfaces][optimize-surfaces] is important! It'll be faster
+by about 30% if you've done the *interest region* exercise from
```
chrishtr, line 1033:
```suggestion
by at least 30% if you've done the *interest region* exercise from
```

book/scheduling.md
```diff
@@ -1135,11 +1121,15 @@ class TaskRunner:
 
     def run(self):
         while True:
-            # ...
+            self.lock.acquire(blocking=True)
+            needs_quit = self.needs_quit
```
chrishtr, line 1125: You'll need to introduce the variable and the code that sets it then. Is that really worth it?

book/scheduling.md
```diff
@@ -1374,37 +1385,15 @@ if __name__ == "__main__":
...
 ::: {.further}
-Unfortunately, Python currently has a [global interpreter lock][gil],
-so our two Python threads don't truly run in parallel,[^why-gil]
-and our browser's *throughput* won't increase much from multi-threading.
-Nonetheless, the *responsiveness* of the browser thread is still
-massively improved, since it often isn't blocked on JavaScript or the
-front half of the rendering pipeline. This is an unfortunate
-limitation of Python that doesn't affect real browsers, so try to
-pretend it's not there.[^why-locks]
+The `needs_raster_and_draw` dirty bit doesn't just make the browser a
```
chrishtr, line 1392: As mentioned in the other comment, I think this is a better place for the global lock go-further. We can move this go-further into a footnote.

book/scheduling.md
```diff
@@ -1423,14 +1412,17 @@ function callback() {
...
 Now, every tick of the counter has an artificial pause during which
-the main thread is stuck running JavaScript. This means it can't
-respond to any events; for example, if you hold down the down key, the
-scrolling will be janky and annoying. I encourage you to try this and
-witness how annoying it is, because modern browsers usually don't have
-this kind of jank.
-
-To fix this, we need to move scrolling from the main thread to the
-browser thread. This is harder than it might seem, because the scroll
+the main thread is stuck running JavaScript.[^adjust] This means it
```
chrishtr, line 1415:
```suggestion
the main thread is stuck running JavaScript. This means it
```

book/scheduling.md
```diff
@@ -1423,14 +1412,17 @@ function callback() {
...
+can't respond to any events; for example, if you hold down the down
+key, the scrolling will be janky and annoying. I encourage you to try
+this and witness how annoying it is, because modern browsers usually
+don't have this kind of jank.
```
chrishtr, line 1419:
```suggestion
don't have this kind of jank.[^adjust]
```

book/scheduling.md
```diff
@@ -1631,52 +1642,47 @@ For example, consider the earlier part of the rendering pipeline:
...
-browser first computing style, and [`getBoundingClientRect`][gbcr]
-needs layout.[^nothing-later] If a web page calls one of these APIs,
-and style or layout is not up-to-date, then is has to be computed
-synchronously, then and there. These computations are called *forced
-style* or *forced layout*. The word "forced" refers to forcing the
-computation to happen right away, as opposed to possibly 16ms in the
-future if it didn't happen to be already computed. Because there are
-forced style and layout situations, browsers have to be able to do
-that work on the main thread if necessary.[^or-stall]
+example, [`getComputedStyle`][gcs] can't be requires first computing
```
chrishtr, line 1660:
```suggestion
example, [`getComputedStyle`][gcs] requires first computing
```

book/scheduling.md
```diff
@@ -1631,52 +1642,47 @@ For example, consider the earlier part of the rendering pipeline:
...
 
 [Servo]: https://en.wikipedia.org/wiki/Servo_(software)
 
-[^nothing-later]: There is no JavaScript API that allows reading back state
-from anything later in the rendering pipeline than layout. It's for
-this reason that we're able to move raster and draw to the browser
-thread.
+[^nothing-later]: There is no JavaScript API that allows reading back
+state from anything later in the rendering pipeline than layout.
+What's why we moved raster and draw to the browser thread.
```
chrishtr, line 1685:
```suggestion
This made it relatively easy for us to move raster and draw to the browser thread.
```

src/lab12.py
```diff
@@ -447,8 +446,7 @@ def run(self):
             needs_quit = self.needs_quit
             self.lock.release()
             if needs_quit:
-                self.handle_quit()
-                return
+                return self.handle_quit()
```
chrishtr, line 449: Is this just a clever trick to avoid two lines?

book/scheduling.md
pavpanchekha: Good point. I still think the GIL should go in front, when we introduce threads. Otherwise readers will feel a bit misled.

pavpanchekha: I've edited the go-further to remove direct references to browser threads while still keeping the general point.

book/scheduling.md
pavpanchekha: Oy, my bad. I'll add back "perhaps" and then make it parallel in the other options.

book/scheduling.md
pavpanchekha: Ok, good point.

src/lab12.py
pavpanchekha: You're right, it's dumb.

pavpanchekha:
Quick question that I just had: what if a page sets a `setTimeout` for a long time, then navigates to another page? We don't handle this correctly, right? Should we add a footnote or something? Is there an easy fix?

book/scheduling.md
chrishtr: How about adding the other sentences from the GIL go-further here? "The go-further in section X said some stuff; here is some more about the implications?" Anyway just an idea, if the sentences fit.

pavpanchekha: See other thread.

chrishtr:
> Quick question that I just had: what if a page sets a `setTimeout` for a long time, then navigates to another page? We don't handle this correctly, right? Should we add a footnote or something? Is there an easy fix?

Good point. Navigation to a new page should clear the task queue. That seems easy enough to just fix?

pavpanchekha:
Ok, let's merge this first, then do that in another PR.

pavpanchekha:
Oh, wait, no, that isn't quite enough, because setting a `setTimeout` doesn't yet add anything to the task queue. There's a subordinate thread, which perhaps we could track and kill, but it's not an easy fix.

chrishtr:
> Oh, wait, no, that isn't quite enough, because setting a `setTimeout` doesn't yet add anything to the task queue. There's a subordinate thread, which perhaps we could track in the `JSContext` and kill, but it's not an easy fix.

Oh. I guess to fix we'd need to save off the URL of the setTimeout and compare when the timer expires. Not too hard?

pavpanchekha:
What if the user follows a link from a page to itself? The most correct token to check is the `JSContext` itself. Actually ...

I just checked and maybe this doesn't happen! The `setTimeout` call creates a task, but that task calls `dispatch_settimeout` on the old `JSContext`, not the new one. That's good, but the old `JSContext` still holds a reference to `tab` and in principle could mess with the new page. (Security issue!!!) I think it might be enough to clear `self.js.tab` when we navigate, but ugh!

chrishtr:
> What if the user follows a link from a page to itself? The most correct token to check is the `JSContext` itself. Actually ...
> 
> I just checked and maybe this doesn't happen! The `setTimeout` call creates a task, but that task calls `dispatch_settimeout` on the old `JSContext`, not the new one. That's good, but the old `JSContext` still holds a reference to `tab` and in principle could mess with the new page. (Security issue!!!) I think it might be enough to clear `self.js.tab` when we navigate, but ugh!

Might be good enough to just note this in a foot note then as one example of the limitations of a toy browser, and that it's not a feasible of our browser to handle every single corner case exactly correctly, just as we don't fix all the memory leaks. But since we noted it, the reader gets a sense of the endless corner cases to make it perfect and secure. I mentioned this in another footnote regarding scrolling as well...

pavpanchekha:
Ok, split the GIL go-further, now to add the `setTimeout` one.

book/scheduling.md
```diff
@@ -1219,15 +1203,40 @@ class Browser:
...
-If you look carefully at our raster-and-draw code, you'll see that to
-draw a display list we also need to know the URL (to update the
-browser chrome), the document height (to allocate a surface of the
-right size), and the scroll position (to draw the right part of the
-surface). Let's make a simple class for storing this data:
+::: {.further}
+
+Originally, threads were a mechanism for improving *responsiveness*
+via pre-emptive multitasking, not *throughput* (frames per second).
+Nowadays, even phones have several cores plus a highly parallel GPU,
```
chrishtr, line 1224:
```suggestion
Nowadays though, even phones have several cores plus a highly parallel GPU,
```

book/scheduling.md
```diff
@@ -1219,15 +1203,40 @@ class Browser:
...
 
-If you look carefully at our raster-and-draw code, you'll see that to
-draw a display list we also need to know the URL (to update the
-browser chrome), the document height (to allocate a surface of the
-right size), and the scroll position (to draw the right part of the
-surface). Let's make a simple class for storing this data:
+::: {.further}
+
+Originally, threads were a mechanism for improving *responsiveness*
+via pre-emptive multitasking, not *throughput* (frames per second).
```
chrishtr, line 1223: [^gil-second]: And our browser still doesn't do that, due to the GIL, lolz.

pavpanchekha:
Added a footnote for the setTimeout bug. The nice thing about adding a footnote is that we don't need to decide on the "best" fix if we just gesture in the direction of multiple 😆

book/scheduling.md
```diff
@@ -1510,17 +1495,37 @@ class Browser:
...
-0 when it loads a new page. We'll set a `scroll_changed_in_tab` flag
-to record when this happens:
+But the main thread also needs to be able to modify the scroll offset.
+We'll add a `scroll_changed_in_tab` flag that tracks whether it's done
+so, and only store the browser thread's scroll offset if
+`scroll_changed_in_tab` is not already true.[^scroll-complicated]
+
+[^scroll-complicated]: Two-threaded scroll has a lot of edge cases,
+including some I didn't anticipate when writing this chapter. For
+example, it's pretty clear that a load should force scroll to 0, but
```
chrishtr, line 1522:
```suggestion
example, it's pretty clear that a load should force scroll to 0 (unless the browser implements 
[scroll restoration](https://developer.mozilla.org/en-US/docs/Web/API/History/scrollRestoration)!, but
```

book/scheduling.md
pavpanchekha: ?

chrishtr: Nevermind, not important.

chrishtr (approved):

PR #373, opened 2022-02-13
Fix the animation frame cadence

chrishtr:
needs_animation_frame is not enough to avoid scheduling tons of extra rendering tasks. For example, once it's true the browser thread will go wild adding more timers. This is what display_scheduled avoided before it was removed last week.

I've re-added display_scheduled via the back door in a way that doesn't look like a dirty bit: put an animation_timer object on Browser that is cleared in commit.

Before this PR, the task list quickly got enormous and it looked like typing into the url bar to load a new page was broken. Now it's very responsive.

This PR also applies some more editorial fixes I found in re-reading today.

### DIFF
--- a/book/scheduling.md
+++ b/book/scheduling.md
@@ -1311,7 +1311,6 @@ class Browser:
             self.active_tab_height = data.height
             if data.display_list:
                 self.active_tab_display_list = data.display_list
-            self.needs_animation_frame = False
             self.animation_timer = None
             self.set_needs_raster_and_draw()
         self.lock.release()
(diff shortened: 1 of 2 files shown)


book/scheduling.md
```diff
@@ -562,8 +560,8 @@ if __name__ == "__main__":
 Now we're scheduling a new rendering task every 16 milliseconds, just
 as we wanted to.
 
-Optimizing rendering tasks
-==========================
+Optimizing rendering
```
chrishtr, line 563: This avoids line wrapping.

pavpanchekha: In my PR I change the title to "Optimizing with dirty bits", take your pick of which you like more.

pavpanchekha (approved):

pavpanchekha:
Approved. I propose you read over mine and, unless there are major issues, do the appropriate merge between the two PRs, since I expect there will be more than the normal number of conflicts. I don't think we need to worry about the exact way wording changes are resolved, as long as we get a reasonable outcome.

PR #374, opened 2022-02-14
Fix the animation frame cadence

chrishtr:
needs_animation_frame is not enough to avoid scheduling tons of extra rendering tasks. For example, once it's true the browser thread will go wild adding more timers. This is what display_scheduled avoided before it was removed last week.

I've re-added display_scheduled via the back door in a way that doesn't look like a dirty bit: put an animation_timer object on Browser that is cleared in commit.

Before this PR, the task list quickly got enormous and it looked like typing into the url bar to load a new page was broken. Now it's very responsive.

This PR also applies some more editorial fixes I found in re-reading today.

### DIFF
--- a/book/scheduling.md
+++ b/book/scheduling.md
@@ -5,11 +5,11 @@ prev: visual-effects
 next: skipped
 ...
 
-The browser must run sophisticated applications while staying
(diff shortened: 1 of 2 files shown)


chrishtr:
Replaces #373.

PR #435, opened 2022-04-01
Strip ports from the top_level_url

### LINKED ISSUE #409: Chapter 10: can't write to the guestbook after adding the SameSite check
After adding the `SameSite` cookie check to the toy browser, I can't write messages on the guestbook server running locally, for example `http://localhost:8000`. I was able to reproduce the issue also with `src/server10.py` and `src/lab10.py` in this repository.

In the following code, `host` doesn't contain a port number while `top_level_host` does when both `url` and `top_level_url` are `http://localhost:8000/`. As a result, `allow_cookie` gets `False` for POST requests.

```py
def request(url, top_level_url, payload=None):
    # ...
    if host in COOKIE_JAR:
        cookie, params = COOKIE_JAR[host]
        allow_cookie = True
        if top_level_url and params.get("samesite", "none") == "lax":
            _, _, top_level_host, _ = top_level_url.split("/", 3)
            allow_cookie = (host == top_level_host or method == "GET")
```

pavpanchekha:
This PR fixes #409 by stripping ports from the top level URL in the SameSite check inside the `request` function. This implements the way SameSite works in Safari and Firefox, but the future looks likely to be a "schemeful" SameSite check already shipped by Chrome and behind a flag in Firefox. I added a footnote about it but don't think we should implement it.

### DIFF
--- a/book/security.md
+++ b/book/security.md
@@ -868,7 +868,15 @@ class JSContext:
 The `request` function can now check the `top_level_url` argument
 before sending `SameSite` cookies. Remember that `SameSite` cookies
 are only sent for `GET` requests or if the new URL and the top-level
-URL have the same host name:
+URL have the same host name:[^schemeful]
+
+[^schemeful]: At I write this, some browsers also check that the new
+    URL and the top-level URL have the same scheme and some browsers
+    ignore subdomains, so that `www.foo.com` and `login.foo.com` are
+    considered the "same site". If cookies were invented today, they'd
+    probably be specific to URL origins, much like CSP policies, but
+    alas historical contingencies and backwards compatibility force
+    rules that are more complex but easier to deploy.
 
 ``` {.python}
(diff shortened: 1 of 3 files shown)


book/security.md
```diff
@@ -868,7 +868,15 @@ class JSContext:
 The `request` function can now check the `top_level_url` argument
 before sending `SameSite` cookies. Remember that `SameSite` cookies
 are only sent for `GET` requests or if the new URL and the top-level
-URL have the same host name:
+URL have the same host name:[^schemeful]
+
+[^schemeful]: At I write this, some browsers also check that the new
```
chrishtr, line 873: Does look like it shipped in M89: https://chromestatus.com/feature/5096179480133632

chrishtr (commented):
LGTM once there is a test :)

PR #438, opened 2022-04-03
Remove an extra new line in the Conclusion list

shuhei:
It's not consistent with the next Appendix list, and the extra line is causing the list items to have `<p>` tags inside.

Screenshot 2022-04-03 at 20 20 17

The extra `<p>` tags don't seem to affect real browsers' rendering, but they pose extra challenges to my toy browser: margin collapsing of parent-child (`<li>` and `<p>`) and then siblings (`<li>` and `<li>`); and `<li>` not being `InlineLayout` 😃 

Screenshot 2022-04-03 at 20 19 04

### DIFF
--- a/book/index.md
+++ b/book/index.md
@@ -82,7 +82,6 @@ Conclusion
 ==========
 
 1. [What Wasn't Covered](skipped.md)
-
 2. [A Changing Landscape](change.md)
 
 Appendix


chrishtr (approved):

shuhei:
Thank you!

PR #440, opened 2022-04-04
Fix Reference to `render` Rather Than `draw`

jerrykuch:
The `draw` function gets referred to as `render` a couple of times.


chrishtr:
Thanks!

PR #470, opened 2022-04-28
Simplify NumericAnimation and running animations

pavpanchekha:
This PR is a big refactor of, basically, the "CSS transitions" section of Chapter 13.

First, I made a lot of changes to how animations are created:

- I've moved the animations from `Tab` to `Element`, so we don't need to pass the `tab` into `style` and clearer naming in some places.
- I've moved responsibility for calling `set_needs_layout` out of `NumericAnimation` and into `Tab`. This has a lot of benefits: `NumericAnimation` now has a reasonable number of arguments and we can move the logic about compositing out of `NumericAnimation` and into `Tab`.
- As a result of that, I've split the `animate` method into two parts. `animate` just steps the frame counter and decides if the animation is still running. `value` produces the new value. This makes the code shorter because it lets us move the node and property name out of `NumericAnimation`.
  - While I didn't do this, I think this could let us fix the edge case where we change one property while another is animating: instead of having the animation write to `node.style`, we can have the `layout` methods query both `node.style[property]` and `node.animations[property].value()`. (We'd write a helper method to do this, of course.) It might be worth doing—more code but fewer confusing edge cases could be a win.
- I've redone the logic for parsing `transition` values, determining what properties are transitioning, and creating animation objects for them. Instead of going property by property, I have a single `diff_styles` method that finds all properties that have changed. Then, because of the changes above, both `TranslateAnimation` and `NumericAnimation` take the same list of arguments, I can treat all animatable properties generically, which eliminates the need for `try_xxx_animation` functions. This saves a lot of repetitive code.

Then, I made a lot of changes to how animations are run:

- I moved running animations into a `run_animations` method; I'm not actually sure this is good but it lets me change the order in which things happen in the text.
- I now run animations *after* `style` instead of *before*. This avoids a kind of circular dependency that would come up earlier and reinforces the pipeline idea of style -> animate -> layout -> paint.
- I now introduce the `needs_layout` and `needs_paint` flags earlier because now it's pretty natural to do so.

### DIFF
--- a/book/animations.md
+++ b/book/animations.md
@@ -871,6 +871,7 @@ class Tab:
         if self.needs_style:
             # ...
             self.needs_layout = True
+            self.needs_style = False
         
         for node in tree_to_list(self.nodes, []):
             # ...
@@ -878,13 +879,12 @@ class Tab:
         if self.needs_layout:
             # ...
             self.needs_paint = True
+            self.needs_layout = False
     
         if self.needs_paint:
             # ...
+            self.needs_paint = False
 
-        self.needs_style = False
-        self.needs_layout = False
-        self.needs_paint = False
         self.measure_render.stop()
 ```
 
--- a/book/animations.md
+++ b/book/animations.md
@@ -691,7 +691,7 @@ is given without units,[^more-units] so `NumericAnimation` looks at
 the old value to determine the unit to use, and then stores the old
 and new values parsed.
 
-[^more-units]: In a real browsers, there are a [lot more][units] units
+[^more-units]: In real browsers, there are a [lot more][units] units
 to contend with. And other constraints, too---like the fact that
 opacity should be a value between 0 and 1.
 
--- a/book/animations.md
+++ b/book/animations.md
@@ -707,15 +707,12 @@ class NumericAnimation:
 
     def animate(self):
         self.frame_count += 1
-        return self.frame_count < self.num_frames
+        if self.frame_count >= self.num_frames: return
 ```
 
-The `animate` method returns a boolean that indicates whether the
-animation is still running. We'll check that before updating the
-actual value of the property being animated.
-
 If the animation is animating, we'll need to compute the new value of
-the property. Let's do that in a `value` method.[^animation-curve]
+the property. Let's return that from the `animate`
+method:[^animation-curve]
 
 [^animation-curve]: Note that this class implements a linear animation
 interpretation (or *easing function*). By default, real browsers use a
@@ -729,7 +726,8 @@ class NumericAnimation:
         total_change = self.new_value - self.old_value
         self.change_per_frame = total_change / num_frames
 
-    def value(self):
+    def animate(self):
+        # ...
         current_value = self.old_value + \
             self.change_per_frame * self.frame_count
         if self.is_px:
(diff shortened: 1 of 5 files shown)


pavpanchekha:
As usual I'll let you review while I work on getting lints to pass and the merge to be clean.

book/animations.md
```diff
@@ -638,226 +634,329 @@ trigger it once every 2 seconds:
...
-def try_transition(property_name, node, old_style, new_style):
-    if not get_transition(property_name, old_style):
-        return None
+def diff_styles(old_style, new_style):
+    old_transitions = parse_transition(old_style.get("transition"))
+    new_transitions = parse_transition(new_style.get("transition"))
+
+    transitions = {}
+    for property in old_transitions:
+        if property not in new_transitions: continue
```
chrishtr, line 791: you forgot old_transitions.

book/animations.md
```diff
@@ -638,226 +634,329 @@ trigger it once every 2 seconds:
...
-        self.node = node
-        self.property_name = property_name
-        self.is_px = is_px
-        self.old_value = old_value
-        self.num_frames = num_frames
-        self.change_per_frame = (new_value - old_value) / num_frames
-        self.tab = tab
-        self.frame_count = 0
-        self.animate()
+However, there's a bit of a catch. Right now, `render()` exits early
```
chrishtr, line 862: Add a sentence calling out that the code above doesn't set any ditty bits at all, and so is definitely incorrect until we decide what bits to set.

infra/compare.py
```diff
@@ -74,8 +74,12 @@ def tangle(file):
     return list(get_blocks(out.stdout.decode("utf8").split("\n")))
 
 def find_block(block, text):
-    differ = difflib.Differ(charjunk=lambda c: c == " ", linejunk=lambda s: "..." in s)
-    d = differ.compare(block.splitlines(keepends=True), text.splitlines(keepends=True))
+    differ = difflib.Differ(charjunk=lambda c: c == " ", linejunk=str.isspace)
```
chrishtr, line 77: Curious what improvements you made to this file.

book/animations.md
```diff
@@ -638,226 +634,329 @@ trigger it once every 2 seconds:
...
-delete and then loop again to delete them. That's why there are two loops and
-the `to_be_deleted` list.
+[^delete-complicated]: For simplicity, this code leaves animations in
+    the `animations` dictionary even when they're done animating.
+    Removing them would be necessary, however, for really long-running
+    tabs where just looping over all the already-completed animations
+    can take a while.
 
+<!--
+This is not actually present in the code; it's hacked in with a
```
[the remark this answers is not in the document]
pavpanchekha: Ah—this comment refers to the fact that the _reader_ should add the `style` and `animations` fields to the `Element` and `Text` classes, but I didn't because it would require copying in most of the browser code from previous chapters before editing it. I think we should just do that, it was a dumb expedient.

book/animations.md
```diff
@@ -638,226 +634,329 @@ trigger it once every 2 seconds:
...
+hasattr() in style() instead. Ugly!
+-->
 
-``` {.python}
-class Tab:
-    def __init__(self, browser):
+``` {.python expected=False}
```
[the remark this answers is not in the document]
pavpanchekha: See comment above.

book/animations.md
pavpanchekha: We're iterating over old_transitions and then additionally filtering by new_transitions. So we end up with things in the intersection.

chrishtr: ah good point

book/animations.md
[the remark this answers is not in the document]
chrishtr: Go it, ok. Acctually, now that you explained it, I'm pretty sure I tried exactly this refactor two months ago and abandoned it for precisely this reason. Do you think moving the dictionaries onto the nodes themselves adds enough value, or should we revert back to storing a double dictionary on the Tab?

infra/compare.py
chrishtr: ok

pavpanchekha:
I updated this branch as we discussed on Friday. It was trickier than I expected, however. They key question was how to start the animation when it is created, without duplicating code or doing anything hard to explain. I finally hit upon a good fix, which is to have `style` reset a property to its old value when creating an animation. That basically starts the animation one frame late. This doesn't require setting any dirty bits or anything, because we're already in `style` overwriting the style information.

src/lab13.py
```diff
@@ -829,6 +832,8 @@ def getAttribute(self, handle, attr):
     def innerHTML_set(self, handle, s):
         doc = HTMLParser(
             "<html><body>" + s + "</body></html>").parse()
+        add_field(doc, "style", lambda x: {})
```
chrishtr, line 835: I thought we had agreed to go back to putting them on the tab and pass some parameters
to `animate`?

book/animations.md
```diff
@@ -638,226 +634,321 @@ trigger it once every 2 seconds:
...
+        if self.is_px:
+            return "{}px".format(current_value)
+        else:
+            return "{}".format(current_value)
 ```
 
-Add some code that detects if a transition remains across a style
-recalc, by comparing two style objects---the ones before and after a style
-update for a DOM node:
+Here I've chosen to compute `change_per_frame` in the constructor. Of
```
chrishtr, line 739: Is this important enough to put in the main text?

book/animations.md
```diff
@@ -638,226 +634,321 @@ trigger it once every 2 seconds:
...
+            if property in ANIMATED_PROPERTIES:
+                AnimationClass = ANIMATED_PROPERTIES[property]
+                animation = AnimationClass(old_value, new_value, num_frames)
+                node.animations[property] = animation
+                node.style[property] = old_value
+```
+
+Now, any time a property listed in `transition` changes its value,
+we'll create an animation and get ready to run it. Note that the
+animation will start running next frame; until then, we want it to
```
chrishtr, line 834:
```suggestion
animation will start running in the next frame; until then, we want it to
```

book/animations.md
```diff
@@ -638,226 +634,321 @@ trigger it once every 2 seconds:
...
+show the old value.
+
+So, let's run the animations! Basically, every frame, we're going to
+want to find all the active animations on the page and call `animate`
+on them. Since these animations are basically a variation of
```
chrishtr, line 839:
```suggestion
on them. Since these animations are a variation of
```

chrishtr (commented):
Did you load all of the example13-*.html files and verify they work correctly?

book/animations.md
pavpanchekha: Doesn't have to be there. I thought it was interesting to point out that some pre-computation could happen here, but now that I think about it a footnote seems more appropriate.

src/lab13.py
pavpanchekha: I think putting animations on tab and passing parameters to animate were separate points.

Passing parameters was the plan for moving the animation loop. I did initially do that, but it lead to code duplication (for the set_needs_layout / set_needs_composite logic). Eventually I hit on the one-more-frame logic to fix that.

But now that I think about it, I think we can put the fields on Element/Text with an even simpler hack than this `add_field` stuff, which should also be very maintainable.

chrishtr: ok I'll wait for that..

pavpanchekha:
I've pushed the new hack for updating `Text` and `Element`. And yes, I've tested all of the examples that are supposed to work.

src/lab13.py
```diff
@@ -16,8 +16,39 @@
...
+        self.animations = {}
+
+    def __repr__(self):
+        attrs = [" " + k + "=\"" + v + "\"" for k, v  in self.attributes.items()]
+        return "<" + self.tag + "".join(attrs) + ">"
+
+# Patch the `Text` and `Element` classes so that all other code that
+# uses them, like HTMLParser, all refer to the patched versions.
+import sys
+sys.modules['lab4'].Text = Text
```
chrishtr, line 49: Interesting. we could do this to a bunch of other chapters to avoid redundant code.

chrishtr (approved):

PR #479, opened 2022-05-04
Fix compositing and transforms

chrishtr:
* Transforms fell of the composited path in PR 470; fixed that
* Somehow PR 470 plus the above fix exhibited a pre-existing bug where the code forgot to skip composited updates with mismatching node ids. Fixed that.


chrishtr:
No test at present.

PR #482, opened 2022-05-04
Add example of multiple layers due to nested effects

pavpanchekha:
Right now the "Compositing Algorithms" section explains that nested effects mean we need to be careful about distributing commands across layers. This PR replaces that discussion with an example.

### DIFF
--- a/book/animations.md
+++ b/book/animations.md
@@ -1148,25 +1148,53 @@ Compositing algorithms
 ======================
 
 The most complex part of compositing and draw is dealing with the hierarchical
-nature of the display list. To deal with this, you might try this algoritihm:
-mark animating visual effects as composited, and raster a `skia.Surface` for
-everything below it. This works fine for the single `DrawText` example we've
-been looking at, but starts to get very complicated when you consider nested
-surfaces. For example, multiple nested DOM nodes could be simultaneously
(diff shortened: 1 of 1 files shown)


book/animations.md
```diff
@@ -1148,25 +1148,53 @@ Compositing algorithms
...
-and *drawing* those surfaces to the screen. Compositing is in charge of finding
-non-animating subtrees of the display list and putting them into groups that
-raster together. Drawing  is in charge of re-creating a tree hierarchy that
-mirrors the hierarchy of animating visual effects in the display list.
-[^temp-surface]
+nature of the display list. For example, consider this web page:
+
+``` {.html}
+<div style="opacity:0.999">
+  <p>
```
chrishtr, line 1155: change to span with color

chrishtr (approved):
Good idea!

book/animations.md
pavpanchekha: Not sure what this is. This bit here is HTML source code; the highlighted bit is lower down.

chrishtr: oh it's a <pre> block below that isn't visible to the reader. I withdraw my comment.

chrishtr (approved):

PR #484, opened 2022-05-04
Reorder the `transform` and overlap testing sections

pavpanchekha:
Right now, the "overlap testing" section is odd because it describes how to handle a case that can't come up. Naturally we should set up the problem first, then describe the solution.

### DIFF
--- a/book/animations.md
+++ b/book/animations.md
@@ -1961,136 +1961,17 @@ and compositing instead.
 
 [WebRender]: https://hacks.mozilla.org/2017/10/the-whole-web-at-maximum-fps-how-webrender-gets-rid-of-jank/
 
-Overlap testing
-===============
-
-The compositing algorithm our browser implemented works great in many cases.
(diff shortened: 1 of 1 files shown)


book/animations.md
```diff
@@ -1961,136 +1961,17 @@ and compositing instead.
...
-
-:::
-
 
 Transform animations
 ====================
 
-The `transform` CSS property lets you apply linear transform visual
+So far, we've been putting two paint commands in the same layer if
+they have the same effects applied. This works---but if paint commands
```
chrishtr, line 1969: "if paint commands overlap"?

The reason I didn't change the order before (even though you suggested it in previous meetings :) ) was that it felt better to properly motivate overlap and why it matters before introducing a method of overlapping. Seems you think the other way around is better...let's just decide together at our meeting today.

pavpanchekha: Ok sounds good.

PR #486, opened 2022-05-05
Remove `repr_recursive`

pavpanchekha:
This PR removes the `repr_recursive` and makes several other simplifications to the `DisplayItem` class:

- [x] Renames the `cmds` field to `children` so that `print_tree` will work
- [x] Adds an `is_paint_command` method so that `children` is always a list and never `None`
- [x] Changes the type-test in `needs_compositing` with overriding in selected classes
- [x] Removes the `noop` field and related `is_noop` method, since they are no longer used anywhere
- [x] Removes the `get_cmds` method; Python style is not to have getters


book/animations.md
```diff
@@ -1088,15 +1097,15 @@ class Tab:
...
         for item in self.display_list:
-            print(item.repr_recursive())
+            print_tree(item)
 ```
 
 When run on the [opacity transition
 example](examples/example13-opacity-transition.html) before the animation has
 begun, it should print something like:
 
-    SaveLayer(alpha=0.999): bounds=Rect(13, 18, 787, 40.3438)
```
chrishtr, line 1098: Removing the rect here invalidates the go further block text a little bit; I'l go back and fix it later.

chrishtr (approved):

book/animations.md
pavpanchekha: Yes—I noticed that but actually the printing the bounds wasn't in the code block cited here! (They're only in your version in the source.) So I agree we should fix up text.

PR #492, opened 2022-05-06
Add parent pointers, thereby removing the need for paint chunks

pavpanchekha:
This is a minimal PR that adds parent pointers to display trees and thereby eliminates the need for paint chunks. It also eliminates the need for the `display_list_to_paint_chunks` method; it is replaced with `tree_to_list`.

I tried to do a deeper refactor and avoid ever creating lists of ancestor effects (that is, replacing all such cases with `while` loops that walk up the parent pointers). However, I got trapped in a rabbit hole and so right now I just call an `ancestor_effects_list` helper method in a few places. The ultimate goal is to not use it anywhere, but it's probably best to do that later, after other refactors happen first.

### DIFF
--- a/book/animations.md
+++ b/book/animations.md
@@ -2356,10 +2356,11 @@ the absolute bounds of a paint chunk:
 
 ``` {.python}
 def absolute_bounds(display_item):
-    ancestor_effects = ancestor_effects_list(display_item)
     retval = display_item.composited_bounds()
-    for ancestor_item in reversed(ancestor_effects):
-        retval = ancestor_item.map(retval)
+    effect = display_item.parent
+    while effect:
+        retval = effect.map(retval)
+        effect = effect.parent
     return retval
 ```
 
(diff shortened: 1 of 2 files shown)


book/animations.md
```diff
@@ -1320,27 +1338,28 @@ class Browser:
...
                 layer = CompositedLayer(skia_context)
-                layer.add_display_item(display_item, ancestor_effects)
+                layer.add_display_item(display_item)
                 composited_layers.append(layer)
 ```
 
+The `can_merge` and `add_display_item` methods use
+`ancestor_effects_list` to make sure we group paint commands
+correctly. If you're not familiar with Python's `for ... else` syntax,
+the `else` block executes only if the loop never executed `break`.
```
chrishtr, line 1356: that's a neat feature, I didn't know about it

chrishtr (approved):

PR #502, opened 2022-05-08
Fix bug in parsing transforms

chrishtr:
Previously, we always returned a (0,0) translation rather than a no-op.
Fixing this revelaed a bug in `composite` where it had the wrong definition of a non-composited subtree.
There is also finally unittesting for some of this.


pavpanchekha:
Oh, great catch!

PR #504, opened 2022-05-09
Merged and tightened the first few sections

pavpanchekha:
This PR merged the sections on "Types of animations", "The animation loop", and "Opacity animations" and tightened up the text so that it is shorter. Parts of those sections were forward reference regarding layout transformations, and those parts were moved forward.

I also simplified the opacity animation to just do a single fade. This does mean the reader will have to refresh the page to see it happen, but it makes the animation code shorter and more similar to the kinds of animations the reader will see later in the chapter. I'm willing to undo this if you don't agree.

### DIFF
--- a/book/animations.md
+++ b/book/animations.md
@@ -23,7 +23,7 @@ Browsers also use animations in response to user actions like
 scrolling, resizing, and pinch-zoom. And some types of animated media,
 like videos, can also be included in web pages.[^video-anim] In this
 chapter we'll focus mostly on web page animations, though we'll touch
-on scrolling at the end.[^excuse]
+on scrolling at the end.
 
 [^general-movement]: Here *movement* should be construed broadly to
 encompass all of the kinds of visual changes humans are used to seeing
--- a/book/animations.md
+++ b/book/animations.md
@@ -30,8 +30,8 @@ encompass all of the kinds of visual changes humans are used to seeing
 and good at recognizing---not just movement from side to side, but
 growing, shrinking, rotating, fading, blurring, and sharpening. The
 point is that an animation is not an *arbitrary* sequence of pictures;
-the sequence must feel, to a human mind trained by experience in the
-real world, to be a continuous motion.
+the sequence must feel continuous to a human mind trained by experience in the
+real world.
 
 [animation]: https://en.wikipedia.org/wiki/Animation
 
@@ -45,7 +45,7 @@ topic is beyond the scope of this book, but it has its own
 Let's write a simple animation using the `requestAnimationFrame` API
 [implemented in Chapter 12](scheduling.md#animating-frames). This
 animation lets us request that some JavaScript code run on the next
-*frame*, and we can have that code change the page slightly.
+frame, and we can have that code change the page slightly.
 To do this repeatedly, we'll need code like this:
 
 ``` {.javascript file=example-opacity-js}
@@ -104,7 +104,7 @@ animation from the beginning.
 
 This animation will almost run in our browser, except that our browser
 doesn't yet support JavaScript changing an element's `style`
-attribute. Let's go ahead and add that feature. We'll need to register
+attribute. Let's go ahead and add that feature. Register
 a setter on the `style` attribute of `Node` in the JavaScript runtime:
 
 ``` {.javascript file=runtime}
@@ -115,7 +115,7 @@ Object.defineProperty(Node.prototype, 'style', {
 });
 ```
 
-Then, inside the browser, we'll need to define a handler for `style_set`:
+Then, inside the browser, define a handler for `style_set`:
 
 ``` {.python}
 class JSContext:
--- a/book/animations.md
+++ b/book/animations.md
@@ -5,56 +5,155 @@ prev: scheduling
 next: skipped
 ...
 
-The UI of a complex web application requires not just fast loading, visually
-interesting rendering, and responsiveness to input and scrolling. It also needs
-smooth *animations* when transitioning between DOM states. These transitions
-improve usability of the web application by helping users understand what
-changed, and improve visual polish by replacing sudden jumps with smooth
-interpolations.
+Complex web application use *animations* when transitioning between
+DOM states. These transitions improve usability by helping users
+understand what changes are occuring. They also improve visual polish
+by replacing sudden jumps with gradual change. But to execute these
+animations smoothly, the browser must make use of the computer's GPU
+and minimize work using compositing.
 
-Modern browsers have APIs that enable animating the styles of DOM elements. To
-implement these APIs performantly, behind the scenes GPU acceleration and
-compositing is needed to make those animations smooth and fast.
+JavaScript Animations
+=====================
 
-Types of animations
-===================
+An [animation] is a sequence of still pictures shown in quick
+succession that create an illusion of *movement* to the human
+eye.[^general-movement] Web pages typically animate effects like
+changing color, fading an element in or out, or resizing an element.
+Browsers also use animations in response to user actions like
+scrolling, resizing, and pinch-zoom. And some types of animated edia,
+like videos, can also be included in web pages.[^video-anim] In this
+chapter we'll focus mostly on web page animations, though we'll touch
+on scrolling at the end.[^excuse]
 
-An [animation] is a sequence of pictures shown in quick succession that create
-an illusion of *movement* to the human eye.[^general-movement] The pixel
-changes in an animation are *not* arbitrary; they are ones that feel logical to
-a human mind trained by experience in the real world.
-
-[^general-movement]: Here *movement* should be construed broadly to encompass
-all of the kinds of visual changes humans are used to seeing and good at
-recognizing---not just movement from side to side, but growing, shrinking,
-rotating, fading, blurring, and sharpening.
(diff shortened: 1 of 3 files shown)


pavpanchekha:
Here's the outline I'm trying to create:

- This "JavaScript animations" section on what animations are, made from the current "Types of animations", "The animation loop", and "Opacity animations" sections
- GPU acceleration section
- A "Compositing" section on what compositing is, made from the current "Compositing" section, the example at the start of the "Compositing Algorithms" section, and the "Width/height animations" example.
- The "CSS transitions" section
- The "Compositing Algorithms" section, renamed to something like "Compositing a Display Tree"
- The "Composited Animations" section
- An "Optimizing compositing" section made from the "Grouping commands" and "Non-composited subtrees" sections
- The "Transform animation" and "Overlap testing" sections in some order
- The "Composited Scrolling" section

I think this should remove some redundant text and it also introduces the width/height animation in a context where it's clear why it matters (because compositing won't work in that case).

After merging and doing a pass for concision we might end up splitting some of the resulting sections into pieces, because some of them are really long (especially the three from "CSS transitions" through "Composited Animations").

book/animations.md
```diff
@@ -5,56 +5,153 @@ prev: scheduling
...
+    div.style = "opacity:" + new_opacity;
+    return current_frame < total_frames;
+}
+```
 
+Here's how it looks; you'll probably need to refresh the page or [open
+it full-screen](examples/example13-opacity-raf.html) to watch the
+animation from the beginning.
 
+<iframe src="examples/example13-opacity-raf.html"></iframe>
```
chrishtr, line 103: I do think it's important for this to keep running forever and not force a reload of the page to restart. Otherwise it'll be too easy to miss it while reading.

chrishtr: Maybe it would be enough to have the iframe page go back and forth, and add sentence saying "here's a demo of it fading out, then fading back in and repeating"

book/animations.md
```diff
@@ -5,56 +5,153 @@ prev: scheduling
...
+
+An [animation] is a sequence of still pictures shown in quick
+succession that create an illusion of *movement* to the human
+eye.[^general-movement] Web pages typically animate effects like
+changing color, fading an element in or out, or resizing an element.
+Browsers also use animations in response to user actions like
+scrolling, resizing, and pinch-zoom. And some types of animated edia,
+like videos, can also be included in web pages.[^video-anim] In this
+chapter we'll focus mostly on web page animations, though we'll touch
+on scrolling at the end.[^excuse]
```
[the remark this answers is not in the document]
pavpanchekha: I'll check and add it back if not.

book/animations.md
```diff
@@ -5,56 +5,153 @@ prev: scheduling
...
+impression of continuous change. It returns `true` until it's done
+animating, and then stops. By changing what `animate` does we can
+change what animation occurs.
 
-[^excuse]: Video animations is a topic unto itself, but is beyond the
-scope of this book. Arguably, canvas is a bit different than the other two,
-since it's implemented by developer scripts. And of course a canvas can have
-animations within it that look about the same as some DOM animations.
+Let's write a fade animation. We can fade in something out by smoothly
+transitioning its `opacity` value from 0.0 to 0.999.[^why-not-one] If we
```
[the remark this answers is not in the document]
pavpanchekha: The old version started opaque and faded out; this starts at transparent and fades in. Do you want it to start at 0.1? (I'm also not sure that 0.999 is necessary in this case, or if actually 0.001 opacity is important instead.)

book/animations.md
pavpanchekha: What if instead of fading back and forth the animation just restarted at transparent? We could even just trigger an iframe refresh. (Mainly, the direction-switching code seemed complicated to me.)

book/animations.md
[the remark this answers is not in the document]
chrishtr: We'll I'd like the demo readers see to be a continuous animation, not a one-shot. And animating from one endpoint that is not completely invisible is better, which is why it was 0.1. 0.999 is necessary to avoid a composited layer pop on real browsers; 0.0 doesn't have the same problem.

book/animations.md
chrishtr: The problem is that once the demo is ported to a css animation in the next section, I wanted it to look identical. Which is why it's an easing function in each direction.

book/animations.md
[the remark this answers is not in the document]
pavpanchekha: Ah—it was this footnote:

> Video animations is a topic unto itself, but is beyond the scope of this book. Arguably, canvas is a bit different than the other two, since it's implemented by developer scripts. And of course a canvas can have animations within it that look about the same as some DOM animations.

I put a reference into the video-animation bullet to VideoNG instead of this footnote. I've removed the footnote reference for now but can put it and the footnote back if you'd like.

book/animations.md
pavpanchekha: Aha! Got it! What if instead of a fully-automated animation that cycles, we had one with buttons to animate one way or the other? That might be better for readers (no distracting animation if they don't want it) and works for both cases.

book/animations.md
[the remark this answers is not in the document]
pavpanchekha: Good to know 0.0 doesn't have that problem. Why isn't there an optimization at that end? (To not draw things that are invisible?) I seem to recall that back in the day all the clickjacking attacks always used like 0.1% opacity for some reason, though perhaps that's optimizations in hit detection, not in compositing...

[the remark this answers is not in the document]
chrishtr: There are optimizations for that 0.0. And you're right that there is no guarantee the composited layer won't be deleted.

book/animations.md
chrishtr: ok sure, if you're up for that.

book/animations.md
```diff
@@ -5,56 +5,153 @@ prev: scheduling
...
+like videos, can also be included in web pages.[^video-anim] In this
+chapter we'll focus mostly on web page animations, though we'll touch
+on scrolling at the end.
+
+[^general-movement]: Here *movement* should be construed broadly to
+encompass all of the kinds of visual changes humans are used to seeing
+and good at recognizing---not just movement from side to side, but
+growing, shrinking, rotating, fading, blurring, and sharpening. The
+point is that an animation is not an *arbitrary* sequence of pictures;
+the sequence must feel, to a human mind trained by experience in the
```
chrishtr, line 33:
```suggestion
the sequence must feel continuous to a human mind trained by experience in the
```

book/animations.md
```diff
@@ -5,56 +5,153 @@ prev: scheduling
...
+real world, to be a continuous motion.
```
chrishtr, line 34:
```suggestion
real world.
```

book/animations.md
```diff
@@ -5,56 +5,153 @@ prev: scheduling
...
+animated canvases. Since our browser doesn't support images, this
+topic is beyond the scope of this book, but it has its own
+[fascinating complexities][videong].
+
+[videong]: https://developer.chrome.com/blog/videong/
+
+Let's write a simple animation using the `requestAnimationFrame` API
+[implemented in Chapter 12](scheduling.md#animating-frames). This
+animation lets us request that some JavaScript code run on the next
+*frame*, and we can have that code change the page slightly.
```
chrishtr, line 48:
```suggestion
frame, and we can have that code change the page slightly.
```

book/animations.md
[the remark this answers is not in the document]
chrishtr: 0.999 vs 1.0 causes a change of stacking context behavior though, which is another reason not to animate to 1.

book/animations.md
```diff
@@ -5,56 +5,153 @@ prev: scheduling
...
+
+``` {.javascript file=runtime}
+Object.defineProperty(Node.prototype, 'style', {
+    set: function(s) {
+        call_python("style_set", this.handle, s.toString());
+    }
+});
+```
+
+Then, inside the browser, we'll need to define a handler for `style_set`:
```
chrishtr, line 118:
```suggestion
Then, inside the browser, define a handler for `style_set`:
```

book/animations.md
pavpanchekha: Ok, I added buttons. It works great in our browser!

chrishtr: \o/

src/example13-opacity-raf.js
```diff
@@ -1,21 +1,43 @@
...
-        (percent_remaining * 0.999 +
-            (1 - percent_remaining) * 0.1);
-    if (frames_remaining-- == 0) {
-        go_down = !go_down
-        frames_remaining = 120;
-    }
-    return true;
+var total_frames = 120;
+var current_frame = 0;
+var change_per_frame = 0.899 / total_frames;
```
chrishtr, line 4:
```suggestion
var change_per_frame = 0.999 / total_frames;
```

chrishtr (approved):

src/example13-opacity-raf.js
pavpanchekha: The animation is now from 0.1 to 0.999. I've clarified this in the code (replacing 0.899 with `(0.999 - 0.1)`).

PR #518, opened 2022-05-24
Megapass on Chapter 13

pavpanchekha:
Hi Chris, sorry that this PR is late (I'm still getting the hang of fatherhood). This is mostly a language pass over Chapter 13. Once something like this is merged I'm ready to release chapter 13. Here are the major changes:

- Fix an issue wherein we used parent pointers earlier in the chapter than we introduced them
- Change the `CompositedLayer` API so that a composited layer always has at least one element (simplifies the code)
- Add extra text clarifying why we need identities on display items
- Update the overlap testing text to no longer refer to paint chunks

Also, a quick question:

- Do we need to make `transform` animatable? I reorganized that section a bit and realized that all of the animation code is now in one chunk at the end. I guess if it's not animatable, the reader might ask why we don't just set its `needs_compositing` to `False` (that wouldn't work, since the transform might contain a composited effect, but the reader might miss that) but we could clarify that. And implementing animated transforms sounds like a great exercise.

### DIFF
--- a/book/animations.md
+++ b/book/animations.md
@@ -6,23 +6,21 @@ next: skipped
 ...
 
 Complex web application use *animations* when transitioning between
-DOM states. These transitions improve usability by helping users
-understand what changes are occurring. They also improve visual polish
-by replacing sudden jumps with gradual changes. But to execute these
-animations smoothly, the browser must make use of the computer's GPU
-and minimize work using compositing.
+states. These animations help users understand the change and improve
+visual polish by replacing sudden jumps with gradual changes. But to
+execute these animations smoothly, the browser must use a technique
+called compositing to minimize redundant work.
 
 JavaScript Animations
 =====================
 
 An [animation] is a sequence of still pictures shown in quick
 succession that create an illusion of *movement* to the human
-eye.[^general-movement] Typical web page animations include
-changing color, fading an element in or out, or resizing an element.
-Browsers also use animations in response to user actions like
-scrolling, resizing, and pinch-zoom. And some types of animated media,
-like videos, can also be included in web pages.[^video-anim] In this
-chapter we'll focus on web page animations.
+eye.[^general-movement] Typical web page animations include changing
+an element's color, fading it in or out, or resizing it. Browsers also
+use animations in response to user actions like scrolling, resizing,
+and pinch-zooming. Plus, some types of animated media (like videos)
+can be included in web pages.[^video-anim]
 
 [^general-movement]: Here *movement* should be construed broadly to
 encompass all of the kinds of visual changes humans are used to seeing
@@ -41,11 +39,12 @@ topic is beyond the scope of this book, but video alone has its own
 
 [videong]: https://developer.chrome.com/blog/videong/
 
-Start with a simple animation using the `requestAnimationFrame` API
-[implemented in Chapter 12](scheduling.md#animating-frames). This
-API lets us request that some JavaScript code run on the next
-frame, and we can have that code change the page slightly.
-To do this repeatedly, we'll need code like this:
+In this chapter we'll focus on animations of web page elements. Let's
+start by writing a simple animation using the `requestAnimationFrame`
+API [implemented in Chapter 12](scheduling.md#animating-frames). This
+method requests that some JavaScript code run on the next frame; to
+run repeatedly over many frames, we can just have that JavaScript code
+call `requestAnimationFrame` itself:
 
 ``` {.javascript file=example-opacity-js replace=animate/fade_out,animation_frame/fade_out}
 function run_animation_frame() {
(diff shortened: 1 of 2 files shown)


book/animations.md
```diff
@@ -1859,23 +1818,23 @@ avoid slowing down transform animations.
 
 [cssfilter]: https://developer.mozilla.org/en-US/docs/Web/CSS/filter
 
-[^overlap-example]: In addition, it's not possible to create the overlapping
-squares example of this section without something like transforms. Real
-browsers have many other methods, such as [position]. In fact, it's very
-difficult to cause overlap at all in our current browser.
+[^overlap-example]: It's not possible to create the overlapping
+squares example of this section without something way of moving an
```
chrishtr, line 1822:
```suggestion
squares example of this section without some way of moving an
```

book/animations.md
```diff
@@ -1859,23 +1818,23 @@ avoid slowing down transform animations.
...
+element around. Real browsers have many methods for this, such as
+[position].
 
 [position]: https://developer.mozilla.org/en-US/docs/Web/CSS/position
 
-The `transform` CSS property in general lets you apply
-[any linear transform][transform-def] in 3D space, but I'll just cover really
-basic 2D translations. Here's HTML for the overlap example mentioned in the
-last section:[^why-zero]
+The `transform` CSS property in quite general and lets you apply [any
```
chrishtr, line 1828:
```suggestion
The `transform` CSS property is quite powerful, and lets you apply [any
```

book/animations.md
```diff
@@ -1887,8 +1846,8 @@ This particular rule is purely a historical artifact.
                 background-color:lightgreen;
                 transform:translate(0px, 0px)"></div>
 
-Adding in support for this kind of transform is not too hard: first
-just parse it:[^space-separated]
+Support these transform is simple. First let's parse the property
```
chrishtr, line 1849:
```suggestion
Supporting these transforms is simple. First let's parse the property
```

book/animations.md
```diff
@@ -2116,7 +2084,7 @@ without just giving up and compositing the scroller.
 Summary
 =======
 
-This chapter introduced the concept of animations. The key takeaways you should
+This chapter introduce animations. The key takeaways you should
```
chrishtr, line 2087:
```suggestion
This chapter introduces animations. The key takeaways you should
```

src/lab13.py
```diff
@@ -109,9 +109,6 @@ def needs_compositing(self):
         return any([child.needs_compositing() \
             for child in self.children])
 
-    def clone(self, children):
```
chrishtr, line 112: Why remove this?

chrishtr:
> would be fine

On further reflection, I remembered additional reasons we should keep transform animations in the book:

* It's weird to have a chapter for animations where we get to the end and there is no animation that moves anything around
* Not too much work

However, I do think we can leave composited transform animations to an exercise. I am sending you a PR now for that.

book/animations.md
```diff
@@ -351,10 +350,10 @@ the GPU. And as expected, speed is much improved:
     Time in render on average: 23ms
 
 That's about three times faster, almost fast enough to hit sixty
-frames per second. And on your computer, you'll likely see even more
-speedup than that, so perhaps your animations are already quite
-smooth. But if we want to go faster yet, we'll need to reduce the
-total amount of work in raster and draw.
+frames per second. (And on your computer, you'll likely see even more
+speedup than I did, so for you it might already be fast enough.) But
```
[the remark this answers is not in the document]
pavpanchekha: Great point.

book/animations.md
```diff
@@ -860,16 +856,12 @@ class Browser:
...
-Some problems remains in the way of great animation performance, though. The
-biggest is that we're doing *all* of these steps on every frame...but the whole
-point of compositing was to avoid re-rastering layers! Let's work on that
-next.
-
-[profiling]: http://localhost:8001/scheduling.html#profiling-rendering
+So simple and elegant! Now, on every frame, we are simply splitting
+the display list into composited layers and the draw display list, and
+then running each of those it their own phase. We're now half-way
+toward getting super-smooth animations. What remains is *not*
```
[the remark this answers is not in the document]
pavpanchekha: Rephrased to make clearer.

book/animations.md
```diff
@@ -1124,29 +1116,27 @@ The `style` function can animate any changed properties listed in
...
 
-Finally, we need to run the animations! Basically, every frame, we're
-going to want to find all the active animations on the page and call
-`animate` on them. Since these animations are a variation of
-JavaScript animations using `requestAnimationFrame`, let's run
-animations right after handling those callbacks:
+With the animations created, we now need to run them on every later
+frame. Basically, on every frame, we iterate through all the active
+animations on the page and call `animate` on them. Since these
+animations cousin to JavaScript using `requestAnimationFrame`, let's
```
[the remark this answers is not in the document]
pavpanchekha: Weird choice of word, rephrased.

src/lab13.py
pavpanchekha: Oh—it doesn't matter either way. In Python it's just a runtime error to call an undefined method; adding `assert False` is only worth it if you write a better error message than the default, and we didn't. So I removed this code block from the chapter (I kept the text, just removed the code block) and then for symmetry removed it from the code as well.

PR #519, opened 2022-05-24
Move composited transform animations to an exercise

chrishtr:


chrishtr:
If you agree with this change, I'll also delete the corresponding code.

pavpanchekha (approved):

pavpanchekha:
Do we even need `TransformAnimation`? Either way, this is good to merge.

pavpanchekha:
Never mind on removing transform animations, just saw your comment on the other PR

PR #545, opened 2022-06-18
Fix bug where InputLayout objects failed to paint or layout correctly

chrishtr:
There were two bugs:

* Putting a block layout mode on an InputLayout
* Painting them twice-- once in InlineLayout and once in InputLayout

Fixing these led to a need to change the focus code to find the object that is an `InputLayout`, not the containing
`InlineLayout`.

I also updated all of the labs that repeated the broken code, added text to forms.md explaining the fixes, and added a unittest.

### DIFF
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -167,7 +167,7 @@ class InlineLayout:
             self.text(node, zoom)
         else:
         	# ...
-            elif node.tag == "input" or node.tag == "button":
+            elif is_input(node):
                 self.input(node, zoom)
             else:
                 for child in node.children:
@@ -409,7 +409,7 @@ This will first require a definition of which elements are focusable:
 
 ``` {.python}
     def is_focusable(node):
-        return node.tag == "input" or node.tag == "button" \
+        return is_input(node) \
         	or node.tag == "a"
 ```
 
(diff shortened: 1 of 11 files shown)


pavpanchekha:
Did we miscommunicate? I left my review above.

chrishtr:
Oops, I missed your comment. It came in 10 minutes before mine and I missed it. I'll revert now and then respond to your comments.

pavpanchekha:
Thanks and sorry about my schedule!

—Pavel Panchekha

chrishtr:
> I hate this system too, but I'm a bit wary about fixing it for `<input>` elements specifically without fixing it elsewhere. It's true that the effects for `InputLayout`s are worse, for example because they have backgrounds, but that just kind of masks an existing bug. Fixing the underlying bug—well, there are multiple options: printing a warning when it occurs? merging `InlineLayout` and `BlockLayout`? implementing anonymous blog boxes?

It seems the only way to fix this issue comprehensively is to add anonymous `LayoutInline` boxes, and also add logic to merge adjacent such inlines together to not introduce incorrect newlines. However, the former problem is worse than the latter, because the page fails to draw elements; adding newlines is minor compared to that.

I can probably implement this via an is_anonymous variable on `LayoutInline`, will try that.

> 
> Separately, two style comments:
> 
> * I've been avoiding writing `is_input` helpers if possible. I know it's good style generally, but it also increases the "weight" of the code substantially, and if we can get away with inlining it (if, say, it's only used in one or two places) I try to do that.

In this case it ended up being three places. In any case, at least one of those will go away if I implement the is_anonymous version.

PR #551, opened 2022-06-20
Re-land: Fix bug where InputLayout objects failed to paint or layout correctly

chrishtr:
This is a re-land of PR #545, with the following changes:

* Restricted to `<input>` elements.
* Adjusts footnote and chapter text to clarify why this is specific to `<input>`

### DIFF
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -105,7 +105,7 @@ class Tab:
 
 Now we just need to pass `zoom` down into `layout`:
 
-``` {.python replace=document/layout_tree}
+``` {.python}
 class Tab:
 	# ...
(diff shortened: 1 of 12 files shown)


chrishtr:
Ok re-land is ready for your review. I responded to your editorial and structural feedback and incorporated it.

Also, regarding input and buttons not being special: I think they are a bit special, and indicated the way I think so in the proposed text. Also, it is true that we don't paint inlines correctly in various other ways. I tried to come up with a simple solution to that but failed to. So I just added a sentence to the footnote admitting as such.

chrishtr:
Going to land this now and discuss later.

PR #564, opened 2022-07-02
Switch the blog to Substack

pavpanchekha:
This PR deletes the blog, which is now moving to Substack, and replaces the signup form with the Substack one.


pavpanchekha:
Once this is approved to merge, I'll push it, import the subscribers from Mailchimp, and publish the explanation post.

chrishtr (approved):

PR #594, opened 2022-08-02
Change accessibility parsing

pavpanchekha:
This PR redoes some of the parsing code in the accessibility chapter.

- [x] Pseudoclasses are represented with a new `PseudoclassSelector`. This avoids the need to modify `TagSelector` and is a bit cleaner in general.
- [x] Parsing for pseudoclasses is moved into a new `simple_selector` method. This avoids duplicate code in descendant selectors.
- [x] The `is_internal` parameter is moved into the `CSSParser` instead of being passed between methods.
- [x] Minor changes to how media queries are parsed, using a local variable instead of parser state.

### DIFF
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -1832,7 +1832,7 @@ class Tab:
                     self.hovered_node.is_hovered = True
 ```
 
-And match it in `TagSelector`:
+And match it in `PseudoclassSelector`:
 
 ``` {.python}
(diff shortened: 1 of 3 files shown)


book/accessibility.md
```diff
@@ -1837,39 +1837,30 @@ And match it in `TagSelector`:
 ``` {.python}
```
chrishtr, line 1837: Update the text on line 1835 to say PseudoclassSelector

chrishtr (approved):

PR #595, opened 2022-08-03
Replace textbox role example with alert

chrishtr:
Textbox (and button) roles are not very useful in reality (confirmed with an a11y expert), but alert is, it's easy to motivate, and easy to implement.


book/accessibility.md
```diff
@@ -2147,14 +2153,69 @@ def announce_text(node):
...
+to alert^[Or the text contexts of the alert changes, but I won't implement
+that.], so we need a way to detect that an attribute. Actually, there isn't
+currently a way to change element attributes other than special ones like
+`style`, so let's first implement that. It will need some runtime code:
+
+``` {.javascript}
+Node.prototype.setAttribute = function(attr, value) {
+    return call_python("setAttribute", this.handle, attr, value);
+}
+```
```
pavpanchekha, line 2173: I don't think this is something we should act on now, but in a later editing pass, do you think it would make sense to implement `setAttribute` instead of `style` in the compositing chapter?

pavpanchekha (approved):
Looks good! I left some minor comments but it is ready to merge.

book/accessibility.md
chrishtr: We could yeah.

chrishtr:
Ready for re-review.

PR #596, opened 2022-08-08
Fix hit testing on transformed content

pavpanchekha:
Hit testing was not properly taking into account transforms. Fix that, add a test, and adjust chapter 13 text accordingly.

### DIFF
--- a/book/animations.md
+++ b/book/animations.md
@@ -1916,7 +1916,8 @@ class Transform(DisplayItem):
 
 We also need to fix the hit testing algorithm to take into account translations
 in `click`. Instead of just comparing the locations of layout objects with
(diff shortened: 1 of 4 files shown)


pavpanchekha:
I think something we should put on the stack "for later" is that this chapter has a lot of code related to different coordinate systems but that code lives in different places and is handled in different ways. This chapter is pretty complex so I'm not thrilled to put more into it, but perhaps there's a way to refactor the code so that converting between trees of coordinate systems isn't so crazy? (Maybe a `CoordinateSystem` object that is relative to a parent `CoordinateSystem`, and a method on each one that walks up the tree? Because we need this in a few places.)

PR #599, opened 2022-08-12
Compiler cleanup

pavpanchekha:
This PR does some minor refactors to the compiler and outline tool to try to make them a bit more regular / more similar to each other. The main new feature is a line counter tool, which counts lines after resolving all imports (so that we get the size a reader would have at any given chapter. Here's the data it makes:

```
~/wbe $ python3 infra/linecount.py src/lab*.py
src/lab1.py	59
src/lab2.py	104
src/lab3.py	195
src/lab4.py	279
src/lab5.py	424
src/lab6.py	595
src/lab7.py	719
src/lab8.py	835
src/lab9.py	919
src/lab10.py	966
src/lab11.py	1218
src/lab12.py	1529
src/lab13.py	1964
src/lab14.py	2295
src/lab15.py	2419
```

### DIFF
--- a/infra/outlines.py
+++ b/infra/outlines.py
@@ -93,21 +93,77 @@ def write_html(objs, indent=0):
             write_html(subs, indent=indent+4)
         print("</code>")
 
+def is_doc_string(cmd):
+    return isinstance(cmd, ast.Expr) and \
+        isinstance(cmd.value, ast.Constant) and isinstance(cmd.value.value, str)
+
+def is_sys_modules_hack(cmd):
+    return isinstance(cmd, ast.Assign) and len(cmd.targets) == 1 and \
+        isinstance(cmd.targets[0], ast.Attribute) and \
+        isinstance(cmd.targets[0].value, ast.Subscript) and \
+        isinstance(cmd.targets[0].value.value, ast.Attribute) and \
+        isinstance(cmd.targets[0].value.value.value, ast.Name) and \
+        cmd.targets[0].value.value.value.id == 'sys' and \
+        cmd.targets[0].value.value.attr == 'modules'
+
+def is_if_main(cmd):
+    return isinstance(cmd, ast.If) and isinstance(cmd.test, ast.Compare) and \
+        isinstance(cmd.test.left, ast.Name) and cmd.test.left.id == "__name__" and \
+        len(cmd.test.comparators) == 1 and isinstance(cmd.test.comparators[0], ast.Constant) and \
+        cmd.test.comparators[0].value == "__main__" and len(cmd.test.ops) == 1 and \
+        isinstance(cmd.test.ops[0], ast.Eq)
+
+def get_name(module, name):
+    assert isinstance(module, ast.Module)
+    for item in module.body:
+        if isinstance(item, ast.Import): pass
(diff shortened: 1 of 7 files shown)


src/lab10.py
```diff
@@ -10,22 +10,14 @@
...
-from lab4 import Element
-from lab4 import Text
-from lab4 import HTMLParser
-from lab5 import DrawRect
-from lab6 import cascade_priority
-from lab6 import resolve_url
-from lab6 import style
-from lab6 import tree_to_list
-from lab6 import CSSParser
+from lab2 import WIDTH, HEIGHT, HSTEP, VSTEP, SCROLL_STEP
```
chrishtr, line 13: Why do you need to change the imports in some of the labs? Were they wrong before?

chrishtr:
Did you verify that all of the labs and widgets still work, and the outlines are unchanged?

infra/outlines.py
```diff
@@ -93,21 +93,77 @@ def write_html(objs, indent=0):
...
+    return isinstance(cmd, ast.If) and isinstance(cmd.test, ast.Compare) and \
+        isinstance(cmd.test.left, ast.Name) and cmd.test.left.id == "__name__" and \
+        len(cmd.test.comparators) == 1 and isinstance(cmd.test.comparators[0], ast.Constant) and \
+        cmd.test.comparators[0].value == "__main__" and len(cmd.test.ops) == 1 and \
+        isinstance(cmd.test.ops[0], ast.Eq)
+
+def get_name(module, name):
+    assert isinstance(module, ast.Module)
+    for item in module.body:
+        if isinstance(item, ast.Import): pass
```
chrishtr, line 119: Is there a way to share code so that you don't have to repeat the calls like is_sys_modules_hack here and in to_item?

src/lab10.py
pavpanchekha: They were. This is a dumb thing that it would be good to automate, but right now the outline only includes things you explicitly `from ... import`. So for example, in one of the labs we actually call `DocumentLayout` but that in turn calls `BlockLayout` etc. If we only `from ... import DocumentLayout`, the outline will have `DocumentLayout` but not `BlockLayout`. So you have to import all of the transitive dependencies as well. Ideally I would construct use-def chains in the compiler so this could be automated but I haven't done that yet.

infra/outlines.py
pavpanchekha: I think avoiding the copy-and-paste here would require more boilerplate (making a custom Visitor?) than it would save.

chrishtr:
> Did you verify that all of the labs and widgets still work, and the outlines are unchanged?

Just this one open comment left..

pavpanchekha:
Sorry, I missed that.  The widgets didn't change at all (the inliner isn't run for widgets). The outlines did change, but I think the changes are acceptable:

- The outlines in general got re-ordered. I don't think we put much thought into the earlier order, so no harm no foul.
- Constants defined together, like in `WIDTH, HEIGHT = 800, 600` now appear as two lines instead of one in the outline. I think this is probably worse, but the difference is super minor. I could fix it if you feel strongly about this.
- The files where I added more imports now have more-complete outlines.

chrishtr:
LGTM then!

PR #601, opened 2022-08-17
Pass on the dark mode section

pavpanchekha:
This PR passes over the "dark mode" and "color scheme" sections. The major changes are:

- I moved the color scheme section to be right after the dark mode section so we don't have to have two copies of the browser color scheme. I thought that was likely to confuse some readers who know about `@media` rules.
- Clarified that the dark mode section has two parts, the chrome and the page contents.
- I modified `DocumentLayout.paint()` to not draw the background rectangle, instead just leaving that to `draw`. Maybe this is bad because it causes a one-frame wrong draw? I don't notice that, but if this was the underlying reason let me know and we'll do this some other way. (Can the one-frame wrong-draw still occur if the web page is shorter than the browser height?)
- I made another small change to how media queries are parsed, thinking about how the reader might want to extend their browser with more media queries.

### DIFF
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -455,16 +455,17 @@ screen is called the [visual viewport][visual-viewport].
 Dark mode
 =========
 
-Next up is helping users who prefer darker screens. The reasons why might
-include extra sensitivity to light, or using a device at night, or at night
-near others without disturbing them. For these use cases, browsers these days
-support a *dark mode* feature that darkens the browser and web pages. For
-example, dark mode pages have a black background and a white foreground
-(as opposed to the default white background and black foreground).
+Another useful visual change is using darker colors to help users who
+are extra sensitive to light, or using a device at night, or just
+prefer a darker color scheme. This browser *dark mode* feature should
+switch both the browser chrome and the web page itself to use white
+text on a black background, and otherwise adjust background colors to
+be darker.
 
-We'll bind the `ctrl-d` keystroke to toggle dark mode:
+We'll trigger dark mode with `Ctrl-d`:
 
 ``` {.python}
+if __name__ == "__main__":
     while True:
         if sdl2.SDL_PollEvent(ctypes.byref(event)) != 0:
             # ...
@@ -474,7 +475,10 @@ We'll bind the `ctrl-d` keystroke to toggle dark mode:
                         browser.toggle_dark_mode()
 ```
 
-Which toggles in the browser:
+When dark mode is active, we need to draw both the browser chrome and
+the web page contents differently. The browser chrome is a bit easier,
+so let's start with that. We'll start with a `dark_mode` field
+indicating whether dark mode is active:
 
 ``` {.python}
 class Browser:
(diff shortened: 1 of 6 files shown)


book/accessibility.md
```diff
@@ -698,6 +638,133 @@ sheets, and instead rely on extensions.
 
 :::
 
+Customizing dark mode
```
chrishtr, line 641: Question: I organized the chapter previously around this order:

Section 1. Some a11y challenges, and how the browser addreseses them
Section 2. APIs for developers to customize.

I think this changes that right? Do you think it'd be better to restructure?

book/accessibility.md
```diff
@@ -455,16 +455,17 @@ screen is called the [visual viewport][visual-viewport].
...
 =========
 
-Next up is helping users who prefer darker screens. The reasons why might
-include extra sensitivity to light, or using a device at night, or at night
-near others without disturbing them. For these use cases, browsers these days
-support a *dark mode* feature that darkens the browser and web pages. For
-example, dark mode pages have a black background and a white foreground
-(as opposed to the default white background and black foreground).
+Another useful visual change is using darker colors to help users who
+are extra sensitive to light, or using a device at night, or just
```
chrishtr, line 459:
```suggestion
are extra sensitive to light, use a device at night, or just
```

book/accessibility.md
```diff
@@ -535,26 +539,19 @@ class Browser:
...
-
-Next up is also informing the `Tab` to switch in or out of dark mode:
+[^more-colors]: Of course, a full-featured browser's chrome has many
+    more buttons than our browser's, and probably uses many more
+    buttons. Most browsers support a theming system that stores all
+    the relevant colors and images, and dark mode switches the browser
+    from one theme to another.
+    
+Now, we want the web page content to change from light mode to dark
+mode as well. To start, let's informing the `Tab` when the user
```
chrishtr, line 553:
```suggestion
mode as well. To start, let's inform the `Tab` when the user
```

book/accessibility.md
```diff
@@ -698,6 +638,133 @@ sheets, and instead rely on extensions.
...
+only in dark mode:
+
+
+``` {.css expected=False}
+@media (prefers-color-scheme:dark) {
+  div { background-color: black; color: white; }
+}
+```
+
+Web developers can use `prefers-color-scheme` queries it in their own
```
chrishtr, line 665:
```suggestion
Web developers can use `prefers-color-scheme` queries in their own
```

book/accessibility.md
chrishtr: You do make a good point though that this change reduces redundancy in the chapter...

src/lab14.py
```diff
@@ -445,16 +445,7 @@ def layout(self, zoom):
...
-        if dark_mode:
-            background_color = "black"
-        else:
-            background_color = "white"
-        display_list.append(
-            DrawRect(skia.Rect.MakeLTRB(
-                self.x, self.y, self.x + self.width,
-                self.y + self.height),
-                background_color, background_color))
+    def paint(self, display_list):
```
chrishtr, line 448: You didn't indicate this change in your chapter text edits. How about just editing earlier chapters to remove the white background from DocumentLayout's paint method and leave it in draw?

chrishtr:
Per offline discussion, LGTM to commit once the smaller issues I noted are fixed. You can do the rework to put customizations next to the features they customize in another PR.

src/lab14.py
pavpanchekha: This is a great find because it turns out we added it in Chapter 13 without explaining it.

PR #602, opened 2022-08-19
Implement Dukpy in widgets using SharedArrayBuffers

pavpanchekha:
This PR implements a mock `dukpy` module for our online widgets. It is pretty tricky, but it works like a dream:

```
> i = dukpy.JSInterpreter()
> i.export_function("ret1", function() { return 1; });
> await i.evaljs("call_python('ret1')")
1
```

Here's how it works. When a `JSInterpreter` is created, it spawns a `Worker` (running the code in `dukpy.js`) and a `SharedArrayBuffer` that it shares with the worker. Every `evaljs` call sends a message to the worker with the code, which then `eval`s it and sends a `postMessage` back with the result. So this works pretty much how you'd expect it to. However, `call_python` is trickier. When a worker runs `call_python`, it sends a `postMessage` to the main thread with the function it wants to run and the arguments it wants to run it with. It then immediately waits on the first dword of the SAB using `Atomics.wait`. This effectively pauses the thread until awoken by the main thread. The main thread, at some point, receives the `postMessage`, runs the function, JSON-encodes the result, writes it to the SAB (writing the length to the first dword and the JSON result to the rest of the dwords), and wakes the worker. The worker then JSON-decodes the result and returns it.

One annoying limitation is that there is a maximum length for responses. In principle, this would break a very long `querySelectorAll`. That said, right now responses are allowed to run up to 255 characters, which should be plenty for widgets. It's easy to bump if we need to.

### DIFF
--- a/www/widgets/rt.js
+++ b/www/widgets/rt.js
@@ -2,7 +2,7 @@
 
 
 export { breakpoint, comparator, filesystem, http_textarea, lib, pysplit,
-    pyrsplit, rt_constants, socket, ssl, tkinter, truthy, Widget };
+    pyrsplit, rt_constants, socket, ssl, tkinter, truthy, Widget, dukpy };
 
 window.TKELEMENT = null;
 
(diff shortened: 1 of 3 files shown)


pavpanchekha:
There's no widget that uses this yet, but I've at least tested it, and I can work on more widgets later.

www/widgets/rt.js
```diff
@@ -291,6 +291,74 @@ static tkinter(options) {
...
+                "type": "array",
+                "buffer": this.buffer,
+            });
+        }
+
+        export_function(name, fn) {
+            this.function_table[name] = fn;
+        }
+
+        evaljs(code, replacements) {
```
chrishtr, line 324: Doesn't evaljs need to be sync? e.g. in `JSContext.displatch_event` ?

pavpanchekha: Ah—our compiler compiles everything to async, so `evaljs` is sync in Python but that becomes async in JS.

chrishtr: I see. Does that mean that preventDefault doesn't work in the widgets?

pavpanchekha: Which `preventDefault` do you mean? When the user clicks in the Canvas, that generates an event, which doesn't have a default action. When we send that to the browser, all of the calls use `await f()` and `async function ...`, but it works the same way. I think `preventDefault` should work; that's just something returned to the caller of `evaljs` and then to the caller of `dispatch_event` which then knows whether to do the default action.

Here's a way to think about it. Imagine a subset of JS where every function call is written `await f()` and every function definition is written `async function`... That's the subset we compile Python to. (There are a few workarounds for parts of JS where this doesn't work.)

chrishtr: ah right. thanks for explaining.

PR #609, opened 2022-08-24
Pass on keyboard shortcuts section

pavpanchekha:
This PR splits the keyboard shortcuts and the customizing tabindex sections in half and recombines them into two new sections. The first section implements keyboard navigation, tab focus, and tabindex. The second half implements focus outlines.

The code stays mostly the same, except that clearing a text entry is moved from focus to activation, which simplifies `click` by a lot.


book/accessibility.md
```diff
@@ -731,36 +739,108 @@ indicated.)
...
+back, typing a URL, or quitting the browser, and also web page
+interactions such as submitting forms, typing in text areas, and
+navigating links.
+
+Let's start with the browser chrome, since it's easiest. Here, we need
+to allow the user to go back, to type in the address bar, and to
+create and cycle through tabs, all with the keyboard. We'll also add a
+keyboard shortcut for quitting the browser.[^one-more] Let's make all
+these shortcuts use the `Ctrl` modifier key so they don't interfere
+with normal typing: `Ctrl-Left` to go back, `Ctrl-L` to type in the
```
chrishtr, line 757: Technically the code uses the lowercase letter, change the text to match?

book/accessibility.md
```diff
@@ -770,185 +850,193 @@ class Browser:
...
                     self.submit_form(elt)
-                    return None
                 elt = elt.parent
-        return elt
 ```
 
-With these methods, we can also avoid a bit of duplicated code in `click`, which
-of course already handled the activation concept---via mouse input---even if it
-didn't have a name at the time:
+All of this action code is copied from the `click` method on `Tab`s,
```
chrishtr, line 928:
```suggestion
All of this activation code is copied from the `click` method on `Tab`s,
```

book/accessibility.md
```diff
@@ -770,185 +850,193 @@ class Browser:
...
-in order. This will first require a definition of which elements are
-focusable:
-
-``` {.python expected=False}
-def is_focusable(node):
-    return node.tag == "input" or node.tag == "button" \
-    	or node.tag == "a"
-```
+We now have the ability to focus on links, buttons, and text entries.
+But as with any browser feature, it's worth asking whether we should
```
chrishtr, line 961: You pose the question but don't answer it I think?

book/accessibility.md
```diff
@@ -770,185 +850,193 @@ class Browser:
...
-        self.lock.release()
-```
+::: {.further}
+Why send the `click` event when an element is activated, instead of a
+special `activate` event? Internet Explorer [did send][onactivate]
+this event, and other browsers used to send a
+[DOMActivate][domactivate] event, but it's been deprecated in favor of
+sending the `click` event even if the element was activated via
+keyboard, not via a click. This works better when the developers aren't
+thinking much about accessibility.
```
chrishtr, line 1032:
```suggestion
thinking much about accessibility, and only register the `click` event listener.
```

src/lab14-tests.md
```diff
@@ -226,7 +226,7 @@ Tab order causes focus:
 
     >>> browser.tabs[0].advance_tab()
     >>> browser.tabs[0].focus
-    <input value="">
```
chrishtr, line 229: I don't remember why this was here; why is it now breaking things?

src/lab14.py
```diff
@@ -1214,28 +1217,27 @@ def render(self):
...
         self.focus = node
         if node:
-            if node.tag == "input":
-                node.attributes["value"] = ""
             node.is_focused = True
         self.set_needs_render()
 
     def activate_element(self, elt):
-        if elt.tag == "a" and "href" in elt.attributes:
+        if elt.tag == "input":
```
chrishtr, line 1229: This is called by `enter()` also; is this behavior appropriate in that situation?

src/lab14.py
```diff
@@ -1723,9 +1712,11 @@ def reset_zoom(self):
         active_tab.task_runner.schedule_task(task)
 
     def load(self, url):
+        self.lock.acquire(blocking=True)
```
chrishtr, line 1715: If this is a bug, fix the previous chapters also?

book/accessibility.md
pavpanchekha: We should standardize how we write keyboard shortcuts. But, for now, done.

src/lab14-tests.md
pavpanchekha: Ah—you used to clear `value` on focus, the new code does it on activate (a purposeful change in semantics, the old semantics were very hard to use) so `value` isn't set yet after focusing.

src/lab14.py
pavpanchekha: Well, I do call it out in the text as a bit of a quirk. In a normal browser `enter` in a text area submits the form, but our browser has never had that behavior. And we need the clearing behavior. It's definitely better to have it in activate than in focus; otherwise, just tabbing through a form requires re-entering every text entry you go through.

src/lab14.py
pavpanchekha: I'm not sure if it's a bug—honestly, the locking discipline in `Browser` is a bit confusing to me. Let's discuss.

chrishtr: ok...why did you add it then? You have to be careful to make sure that all callsites expect no locks.

pavpanchekha: I, uh, think it's a bug, but I'm not sure. Basically, I looked over the code and I think we're not correctly locking when tabs are added or removed, which is a problem because in `commit` we do things based on the active tab. But I don't remember the locking discipline well and coming up with a reproducible test case for a locking issue seemed too hard. Let's discuss tomorrow and go from there—whatever the resolution is, we'll reflect that in this chapter.

chrishtr: Ok. My recommendation is to remove this part of the change for now and I'll take care of investigating this code path comprehensively and fixing any locking bug.

The TL;DR of the locking discipline is: any code in Browser that touches data structures that are changed during commit needs to be enclosed in a lock.

pavpanchekha: Done, I believe this is mergable.

src/lab14.py
```diff
@@ -1628,12 +1620,11 @@ def go_back(self):
         active_tab.task_runner.schedule_task(task)
         self.clear_data()
 
-    def add_tab(self):
-        self.load("https://browser.engineering/")
-
     def cycle_tabs(self):
+        self.lock.acquire(blocking=True)
```
chrishtr, line 1624: Also remove this

chrishtr (approved):

PR #610, opened 2022-08-24
Many and various fixes to the feedback tools

pavpanchekha:
This PR fixes a lot of small issues, like comments in a narrow window, not shadowing Ctrl-E on macOS, not showing the pop-up if it's already active, and so on.

### DIFF
--- a/infra/api.py
+++ b/infra/api.py
@@ -8,6 +8,8 @@ import difflib
 import html
 import hashlib
 
+NOPASSWORD = False
+
 bottle.TEMPLATE_PATH.append(".")
 
 class Data:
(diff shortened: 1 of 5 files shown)


pavpanchekha:
Frankly this code is pretty bad and should be rewritten, but for now this at least fixes some major issues.

infra/api.py
```diff
@@ -249,10 +253,15 @@ def status():
...
-        good = f.read(256)
-    # Equivalent to `good == heng.digest()` but constant-time-ish
-    if sum([0 if a == b else 1 for a, b in zip(good, heng.digest())]) == 0:
+    try:
+        with open("pw.hash", "rb") as f:
+            good = f.read(256)
+        # Equivalent to `good == heng.digest()` but constant-time-ish
+        allowed = sum([0 if a == b else 1 for a, b in zip(good, heng.digest())]) == 0
+    except FileNotFoundError:
+        allowed = True
```
chrishtr, line 262: What effect does this have? Is it a security hole perhaps?

www/feedback.js
```diff
@@ -1,22 +1,45 @@
...
+    if (navigator.platform.indexOf("Mac") === 0 || navigator.platform === "iPhone") {
+        return e.metaKey;
+    } else {
+        return e.ctrlKey;
+    }  
+}
+
+function ctrl_key_name() {
+    if (navigator.platform.indexOf("Mac") === 0 || navigator.platform === "iPhone") {
+        return "Ctrl";
```
chrishtr, line 17: Is this the opposite of what you want?

infra/api.py
pavpanchekha: Ok, I've made it so you have to pass a `--no-password` flag.

www/feedback.js
pavpanchekha: Good catch, fixed.

PR #611, opened 2022-08-24
Rehost the text of xiyouji to fix #606

### LINKED ISSUE #606: In chapter 2, the request threw an exception because the test site's certificate was updated
Python version: 3.10.1
System: macos 12.3.1

When python requests the website https://www.zggdwx.com through ssl, an exception will be thrown:
```
ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:997)
```
I tried running the `Install Certificates.command` but it still didn't solve the problem.
Then try to open the website through firefox browser and find that there is a problem with the certificate: 
image

After investigation, it was found that the https://www.zggdwx.com website has recently updated the certificate, and the issuer of the certificate is TrustAsia Technologies, Inc. which is not in the list of certificates trusted by mozilla (it looks like they are trying to request mozilla to add trust: [link](https://bugzilla.mozilla.org/show_bug.cgi?id=1688854), but no success yet), so python throws an exception.

So I think it's better to put the "西游记" content of the website on github, and the test website in the course also requests github instead.

pavpanchekha:
The site we'd been using has certificate issues, plus it's always been slow.


src/xiyouji.html
```diff
@@ -0,0 +1,184 @@
+<!doctype html>
```
chrishtr, line 1: I think the name of this fiile (xiyouji.html) needs to start with "example" to meet the Makefile rule?

pavpanchekha: Turns out no; the Make rule is:

```
www/examples/%.html: src/%.html
	cp $< www/examples
```

It works:

```
~/wbe $ make www/examples/xiyouji.html
cp src/xiyouji.html www/examples
```

PR #612, opened 2022-08-26
Get Chapter 8 code to compile into a widget

pavpanchekha:
This wasn't that bad; mostly, I needed to implement a way for the compiled browser and server to talk to each other.

### DIFF
--- a/www/widgets/rt.js
+++ b/www/widgets/rt.js
@@ -82,7 +82,6 @@ static socket() {
             if (this.host == "localhost" && rt_constants.URLS["local://" + this.port]) {
                 let s = new socket({family: "inet", type: "stream", proto: "tcp"});
                 s.is_proxy_socket = true;
-                console.log(this.input);
                 s.output = this.input;
                 await rt_constants.URLS["local://" + this.port](s)
                 this.output = s.input;
@@ -104,7 +103,6 @@ static socket() {
                 this.idx = 0;
                 return this;
             } else {
-                console.log(this);
                 throw new WidgetXHRError(this.host);               
             }
             return this;
--- a/www/widgets/rt.js
+++ b/www/widgets/rt.js
@@ -2,7 +2,7 @@
 
 
 export { breakpoint, comparator, filesystem, http_textarea, lib, pysplit,
-    pyrsplit, rt_constants, socket, ssl, tkinter, truthy, Widget, dukpy };
+    pyrsplit, rt_constants, socket, ssl, tkinter, truthy, Widget, dukpy, urllib };
 
 window.TKELEMENT = null;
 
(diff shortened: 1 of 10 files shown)


www/widgets/rt.js
```diff
@@ -66,13 +66,28 @@ static socket(URLS) {
...
+            if (this.is_proxy_socket) return this;
 
             let [line1] = this.input.split("\r\n", 1);
             let [method, path, protocol] = line1.split(" ");
             this.url = this.scheme + "://" + this.host + path;
-            if (rt_constants.URLS[this.url]) {
+            if (this.host == "localhost" && rt_constants.URLS["local://" + this.port]) {
+                let s = new socket({family: "inet", type: "stream", proto: "tcp"});
+                s.is_proxy_socket = true;
+                console.log(this.input);
```
chrishtr, line 85: Needed?

chrishtr (commented):
Can you point me to the code that causes the client and server to connect, or just summarize in words how it works? I can't find it yet.

www/widgets/rt.js
pavpanchekha: Yes, there are a few moving pieces. The basic idea is simple: the server indicates a function to call per connection. When the browser connects to it, we take its request, stuff it in a new `socket` that we hand the server, `await` the server handler to finish, pull out its response from what it wrote to this new `socket`, and pass that back to the browser. But there are some quirks.

First, for the server I'm not using the existing server mainloop (which `bind`s a socket and then `accept`s from it) because it's blocking. Instead, I just have a single library method, `socket.accept(port, fn)` which says, "run this socket for this port". In the background, this saves a fake URL, `local://<port>/<path>`, to connect to. (It needs to be distinct from `https://` because the code for handling it is different. It should be invisible to users.)

Second, when the browser requests a connection to `https://localhost:<port>`, we look up this fake `local://` URL, and pull out the function that responds to it. We now need to call that function with a new socket object that represents the "other end" of the connection. Here the code gets a little nasty, because right now our mock `socket` is meant to do a request followed by response, whereas on the server side the order is backwards. (This is the code you're commenting on here.) So I make a new `socket`, save the incoming request as its `output` (so that it can be read by the server), call the server with it, and read the `input` (what the server sent) back. This almost works, except that right now we have some code that looks at a socket's `input` to determine what URL is being requested; I added a new `is_proxy_socket` flag to skip that step.

chrishtr: Ok makes sense.

Remaining question: Do you need the console.log here?

pavpanchekha: Good catch, woops, I'll remove it.

pavpanchekha: The console.log is fixed, should be mergable.

PR #616, opened 2022-08-30
Get Chapter 9 compiling and running in a widget

pavpanchekha:
Getting Chapter 9 running required a bunch of changes throughout the compiler, including some small bug fixes and also:

- Better handling of list comprehensions, which now compile in a way that allows calling async functions
- Better support for `try`/`catch`, which now correctly checks the type of the exception
- Fix `dukpy.js` to use async handler functions
- In `dukpy.js`, save a private copy of `console` and `postMessage` for when they are overwritten by the browser runtime
- Avoid accidentally returning the last-defined function from `dukpy.js`
- Using indirect eval, a truly hellish hack, in `dukpy.js` for ""reasons""

### DIFF
--- a/infra/compiler.md
+++ b/infra/compiler.md
@@ -407,7 +407,7 @@ The `with` statement in Python has no analog in JS (which also has a
 compile it into `open` and `close` statements:
 
     >>> Test.stmt("with open('browser.css') as f:\n f.read()")
-    f = (filesystem.open("browser.css"));
+    let f = (filesystem.open("browser.css"));
     (f.read());
     f.close();
 
--- a/infra/compiler.md
+++ b/infra/compiler.md
@@ -409,7 +409,7 @@ compile it into `open` and `close` statements:
     >>> Test.stmt("with open('browser.css') as f:\n f.read()")
     f = (filesystem.open("browser.css"));
     (f.read());
-    (f.close());
+    f.close();
 
 But `if` statements are weird. First, there's the top-level `if`
 statement, which we elide. But then also there are one-line `if`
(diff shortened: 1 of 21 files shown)


infra/compiler.md
```diff
@@ -234,12 +234,19 @@ Conditional expressions are also supported:
     >>> Test.expr("x if y else z")
     (y ? x : z)
 
-One weird case is list comprehensions. We compile those to `map` and `filter`:
-
+One weird case is list comprehensions. We compile those to `map` and
+`asyncfilter`, which can get a bit ugly:
+
+    >>> Test.expr("[x for x in v]")
+    v
```
chrishtr, line 241: aside: did you special-case no-ops somewhere?

chrishtr (approved):
Cool!

Please make sure all widgets still work from previous chapters.

infra/compiler.md
pavpanchekha: Not really. Before all list comprehensions had a “map” component but not necessarily a “filter” component. So I didn’t output “filter” code if it wasn’t there, but I always output “map”. Now there are some pure-filter list comprehensions, plus the “map” code got uglier to deal with async, so I made it not output a “map” unless it does something. Put it together, and no-ops now do nothing.

pavpanchekha:
The good news is: I checked and all of the other labs work. I also messed with the Makefile a bit because lab4-browser.html and lab9-browser.html are different enough to not be captured by the template.

The bad news is that I found a new bug in *this* chapter, regarding the `node_to_handle` map, which needs to be a JS `Map` object, not a regular object, because it maps `Element` objects to integers. I fixed it with hints, which required a lot of work. I'll be fixing this PR and asking you to re-review.

PR #620, opened 2022-08-31
Pass over the outlining section

pavpanchekha:
This PR has good and bad parts. The good part is it does a pass over the tab-outline section of Chapter 14, and I think improves a few things about. (The biggest change, I think, is splitting `DrawRect` with no fill but with a stroke into a `DrawOutline` command.)

But the bad part is that I tried to get rid of our use of `hasattr` by patching `Element`, and while that worked it was nastier than I thought and I had to fix some very subtle bugs first. So we need to find a replacement for this "patching" idea.


pavpanchekha:
@chrishtr, please review this and I think for now we should merge (once you review & I fix the conflicts) and deal with the redefining-`Element` issue later. I'm sure it will bite us again.

book/accessibility.md
```diff
@@ -770,280 +849,327 @@ class Browser:
...
+        if parent:
+            paint_outline(parent, display_list, outline_rect)
 ```
 
-Iterating through the focusable elements with the keyboard, and highlighting
-them with a focus rect, should now work. Try it in
-this [example](examples/example14-focus.html). And if you zoom in enough, you
-should be able to make the link cross multiple lines and use multiple
-focus rectangles.
+You can also add a `paint_outline` call to `BlockLayout`,
```
chrishtr, line 1162:
```suggestion
You should also add a `paint_outline` call to `BlockLayout`,
```

book/accessibility.md
```diff
@@ -1058,56 +1184,230 @@ important for browsers and web sites to provide keyboard input alternatives.
...
+    outline: 3px solid red;
 
-So what we need to do is bundle up the semantics of the DOM and present them to
-the user in some way they can access. For a user that can't see the screen,
-the simplest approach will be to read the content out loud. This functionality
-is called a *screen reader*. Screen readers are typically[^why-diff] a different
-application than the browser. But it's actually not so hard to build a
-(very simple) one directly into our browser, and doing so will give you
-insight into how accessibility actually works in a browser.[^os-pain] 
+Which means "make the outline red 3px thick". First parse it:
```
chrishtr, line 1195:
```suggestion
Which means "make the outline red and 3 pixels thick". First parse it:
```

book/accessibility.md
```diff
@@ -1058,56 +1184,230 @@ important for browsers and web sites to provide keyboard input alternatives.
...
+```
 
-[^os-pain]: Another reason is that it's quite a lot of work to directly
-integrate a browser with the accessibility APIs of each OS. Further, it's not
-very easy to find Python bindings for these APIs, especially Linux. And as
-you'll see, it really isn't very hard to get the basics working, though a big
-reason is that these days there are high-quality text-to-speech libraries
-available for non-commercial use.
+Now to use it. The outline will be present if the element is focused or has the
+`outline` CSS property generally. While we could finish implementing that via
```
chrishtr, line 1210:
```suggestion
`outline` CSS property generally. However: while we could finish implementing that via
```

src/lab14.py
```diff
@@ -550,8 +544,11 @@ def paint(self, display_list):
         cmds.append(DrawText(self.x, self.y,
                              text, self.font, color))
 
-        paint_outline(self.node, cmds, rect)
+        if self.node.is_focused and self.node.tag == "input":
```
chrishtr, line 547: Should this ideally be back-ported to the chapter in which it was introduced?

book/accessibility.md
```diff
@@ -770,280 +849,327 @@ class Browser:
...
+``` {.python}
+def is_focusable(node):
+    return node.tag in ["input", "button", "a"] \
+        or "tabindex" in node.attributes
 ```
 
- When `tab` is pressed and we're at the
-end of the focusable list, focus the address bar:
+Next, we need to make sure to send a `click` event when an element is
+activated:
```
chrishtr, line 1013: Add a sentence about the JS behavior, since just like click  enter can be preventDefaulted

chrishtr (approved):

book/accessibility.md
```diff
@@ -770,280 +849,327 @@ class Browser:
...
+author might want certain links not to be focusable (like "permalinks"
+to a section heading, which would just be noise to most users), or
+might want to change the order in which the user tabs through
+focusable items.
 
-``` {.python expected=False}
-def is_focusable(node):
-    return node.tag == "input" or node.tag == "button" \
-    	or node.tag == "a"
+Browsers support the `tabindex` HTML attribute to make this possible.
```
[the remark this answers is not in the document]
pavpanchekha: I rephrased the next sentence on focusability to stand out a bit more, and I've implemented the -1 no focus thing.

book/accessibility.md
```diff
@@ -770,280 +849,327 @@ class Browser:
...
+added by drawing a verticle line in the `Tab`'s `render` method. We
+could extend that code to draw a cursor or an outline, but before we
+make that method more complicated let's move it into the `InputLayout`
+so we have easier access to size and position information.[^effects]
+To do that, we'll need each `InputLayout` to know whether or not it is
+currently focused:
+
+[^effects]: As a side effect, this change will also mean text cursors
+    are now affected by visual effects, including blends, opacity, and
+    translations. Translations in particular are important.
```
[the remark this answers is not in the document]
pavpanchekha: Yes, as of Chapter 13. More generally, however, I think it tells me that _way back_ in Chapter 7, we should have each `InputLayout` draw its text cursor in `paint` by consulting an `is_focused` field (like we now do in this chapter). I think at the time we checked Chrome and it does it the way we do it in the book, but that looks like a mistake now.

src/lab14.py
pavpanchekha: See above: yes, but this will require some reasonably large changes so I'd rather push it on the stack for that chapter. I've added it to our chapter notes document.

PR #622, opened 2022-09-02
Get a chapter 10 widget running

pavpanchekha:


infra/compile.py
```diff
@@ -509,6 +520,10 @@ def compile_expr(tree, ctx):
             return "Error"
         elif tree.id in OUR_CONSTANTS:
             return "constants.{}".format(tree.id)
+        elif tree.id == "_":
```
chrishtr, line 523: what is this feature?

src/lab10.py
```diff
@@ -166,7 +166,7 @@ def XMLHttpRequest_send(self, method, url, body):
         full_url = resolve_url(url, self.tab.url)
         if not self.tab.allowed_request(full_url):
             raise Exception("Cross-origin XHR blocked by CSP")
-        headers, out = request(full_url, self.tab.url, payload=body)
```
chrishtr, line 169: Do named parameters not work in this case?

www/widgets/rt.js
```diff
@@ -325,6 +325,24 @@ static urllib() {
     return { parse: parse };
 }
 
+static html() {
```
chrishtr, line 328: What does this new method do?

pavpanchekha: It implements Python’s html.escape function by leveraging the browser’s html unpaused.

src/lab10.py
pavpanchekha: Well, ugh, the compiler compiles named arguments to passing a dictionary as the last argument, but when it compiles the function being called it doesn’t know which arguments will be passed by keyword. In principle I can compile our functions and library functions differently, so we can keep using keyword arguments, but I’ve been thinking we need to refactor the networking part of the book anyway so I’d rather punt the keyword argument question for now.

infra/compile.py
pavpanchekha: Oh lol. I sometimes use _ as an “ignore” argument (it’s a convention, not a language feature) but JS doesn’t like the same variable being bound twice in the same assignment, so I compile every _ variable to a different variable name in JS. This would be a problem if we ever wanted to use this variable, but the whole point of an “ignore” variable is that we won’t.

www/widgets/rt.js
chrishtr: Is that a well-documented technique? If so, can you add a comment to documentation for it?

chrishtr (approved):

pavpanchekha:
I got the rest of this all working; Chapter 10 seems to run well, though if you type in the wrong password it is pretty user-unfriendly. (The Python version just hard-crashes. The browser version does the same, but the user probably doesn't have their browser console open to see what happened.) I propose we make the widget runner handle crashes in a separate PR.

Interestingly, I found (and fixed) a Chapter 10 bug. I think it was just a bad / confused copy/paste from previous chapters. (It caused us to not clear the display list when you typed into a text box, leading to ever-darker text.)

chrishtr (approved):

PR #625, opened 2022-09-07
Add a requirements.txt

pavpanchekha:
This PR adds a `requirements.txt` file to the repository, which allows us to fix the versions of our dependencies. This should help guard against possible forwards-incompatibility and also documents our dependencies.

I've generally included transitive dependencies, but this may not be wise—we can discuss later.

### DIFF
--- a/.github/workflows/compare.yml
+++ b/.github/workflows/compare.yml
@@ -11,7 +11,7 @@ jobs:
         with:
           python-version: 3.9
       - name: Install dependencies
-        run: python -m pip install -r requirements.txt
+        run: python -m pip install -r requirements.txt --no-deps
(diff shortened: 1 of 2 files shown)


chrishtr (commented):
Question: how did you generate this file? Is there a way to enforce it being complete or failing a test if we accidentally add a new thing without updating this file?

pavpanchekha:
Good question. The CI should check—I just double-checked the documentation and fixed it, so that if we forget something in the `requirements.txt` file either the "Install dependencies" line will fail or, later the "make book draft lint test" line will fail. So I think we'll know.

To generate it, I used `pip freeze`, which outputs all the Python packages installed on your system, and then I stripped out the ones that I didn't think were relevant.

PR #630, opened 2022-09-12
Pass on screen-reader section

pavpanchekha:
I'm not really sure this PR is even good, but hear me out.

My biggest concern with the accessibility tree section is that we haven't really justified why there needs to be a tree. My attempted solution to this was to actually implement moving in / out of the accessibility tree, which a real screen reader does support. I do think this is probably a good idea.

To implement this, you need a pointer for where in the document you currently are reading. (In the existing version of the chapter, that pointer can actually only point to focusable things.) I ended up making an `accessibility_focus` field that points to an accessibility node instead of just an HTML element. This way you can easily go to the previous / next / parent / child node. That works nicely.

However, since we rebuild the accessibility tree every so often, we need to "fix up" the `accessibility_focus` pointer (make it point into the new, not the old, tree). I think this is kind of ugly and I wish I could avoid it.

Plus, I also need to sync the accessibility focus back to the actual focus when you move over to a focusable element with the accessibility controls. I don't implement that because it's not 100% obvious the best way to do this. I think this is another ugliness of this approach.

Anyway, I'm looking for feedback on this direction. I think letting the user traverse the accessibility tree is good, but the implementation is not as clean as we'd like.

### DIFF
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -1680,14 +1680,15 @@ class Tab:
         # ...
         self.accessibility_is_on = False
 
+    def render(self):
+        if self.needs_accessibility:
+            if self.accessibility_is_on:
+                task = Task(self.speak_update)
+                self.task_runner.schedule_task(task)
+
     def toggle_accessibility(self):
         self.accessibility_is_on = not self.accessibility_is_on
         self.set_needs_render()
-
-    def render(self):
-        if self.needs_accessibility:
-            # ...
-            self.speak_task()
 ```
 
 Let's now use this code to speak the whole document once after it's been loaded:
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -1470,8 +1470,7 @@ screen-reader users. The browser therefore builds a separate
 several other ways in real browsers that elements can be made
 invisible, such as with the `visibility` or `display` CSS properties.
 
-Let's implement an accessibility tree in our browser. It is built in a
-rendering phase just after layout:
+The accessibility tree is built in a rendering phase just after layout:
 
 ``` {.python}
 class Tab:
@@ -1681,15 +1680,14 @@ class Tab:
         # ...
         self.accessibility_is_on = False
 
-    def render(self):
-        if self.needs_accessibility:
-            if self.accessibility_is_on:
-                task = Task(self.speak_update)
-                self.task_runner.schedule_task(task)
-
     def toggle_accessibility(self):
         self.accessibility_is_on = not self.accessibility_is_on
         self.set_needs_render()
+
+    def render(self):
+        if self.needs_accessibility:
+            # ...
+            self.speak_task()
 ```
 
 Let's now use this code to speak the whole document once after it's been loaded:
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -1498,9 +1498,9 @@ The accessibility tree is built out of `AccessibilityNode`s:
 
 ``` {.python}
 class AccessibilityNode:
-    def __init__(self, node):
+    def __init__(self, node, parent=None):
         self.node = node
-        self.previous = None
+        self.parent = parent
         self.children = []
 ```
 
@@ -1557,7 +1557,7 @@ class AccessibilityNode:
             self.build_internal(child_node)
 
     def build_internal(self, node):
-        child = AccessibilityNode(node)
+        child = AccessibilityNode(node, self)
         if child.role != "none":
             self.children.append(child)
             parent = child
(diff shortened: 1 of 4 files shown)


src/lab14.py
```diff
@@ -1153,20 +1215,25 @@ def render(self):
...
-            if self.accessibility_is_on:
-                self.needs_accessibility = True
-            else:
-                self.needs_paint = True
+            self.needs_accessibility = True
+            self.needs_paint = True
             self.needs_layout = False
 
         if self.needs_accessibility:
+            old_afocus = self.accessibility_focus
```
chrishtr, line 1223: could accessibility_node be an actual Node? And we dynamically compute which a11y node it refers to?

book/accessibility.md
```diff
@@ -1610,166 +1653,287 @@ used, so let's bind turning it on to the `ctrl-a` keystroke:
...
-The `build` method on `AccessibilityNode` recursively creates the tree. Whether
-a node is represented in the accessibility tree depends on its tag and style,
-which in turn determine not only whether the element is presentational, but if
-not what its semantics are. For example, a `<div>` by itself is by default
-considered presentational or otherwise having no accessibility semantics, but
-`<input>`, `<a>` and `<button>` have them by default. The semantics are called
-the *role* of the element---the role it plays in the meaning of the document.
-And these roles are not arbitrary text; they are specified in a
-[standard][aria-roles] just like rendering.[^not-exposed]
+Now we can add navigation key bindings for the accessibility tree.
```
chrishtr, line 1791: Does really a11y actually work like this? I thought it navigated conceptually by landmarks.

book/accessibility.md
```diff
@@ -1610,166 +1653,287 @@ used, so let's bind turning it on to the `ctrl-a` keystroke:
...
-        for child_node in node.children:
-            AccessibilityNode.build_internal(child_node, parent)
+class Tab:
+    def toggle_accessibility_reading(self):
+        self.accessibility_reading = not self.accessibility_reading
+        self.set_needs_accessibility()
 ```
 
-Let's now use this code to speak the whole document once after it's been loaded:
+If we bind this to `Ctrl-R`, the user will be able to start reading
```
chrishtr, line 1892: ctrl-r is usually for reloading the page...

chrishtr: ctrl-s? For start and stop.

chrishtr (commented):
I like the idea of something that explicitly traverses the tree. I added the focus-on-hover feature as a way to justify the tree, but it doesn't do a good enough job of using it's tree structure...

book/accessibility.md
pavpanchekha: Well... There are lots of navigation options in most screen readers, and landmarks are often more efficient than directly navigating the tree. But this one is supported at least by VoiceOver (which is the only screen reader I know), though it's probably not the most common feature.

I guess I'm dissuading myself that this was a good example. Why implement an obscure option instead of the most common one? Because we want to demonstrate the accessibility tree, sure, but surely there is a better reason to build the accessibility tree than supporting this one obscure option? I think this brings us back to the goal stated in the PR: we need better motivation for the accessibility tree being a tree.

src/lab14.py
pavpanchekha: We could, definitely, but that raises the question of why we need the accessibility tree at all. As far as I can tell, this is a constraint imposed by operating systems: their accessibility technologies rquire you to provide an accessibility tree...

chrishtr: Sorry I was unclear. I meant to cache a Node, and then whenever code currently writes `self.accessibility_focus` we would instead write `self.accessibility_node.layout_object.accessiblity_object`, via tree forward pointers.

chrishtr: Operating systems do not actually require there be a tree. They do require that the browser respond to requests synchronously though, and this requires them to have all the necessary parts of the DOM ready at hand. In a multi-process browser, this requires either serializing the DOM to another process, or some subset of it (the accessibility tree). It's a tree only because that is the reasonable and obvious choice for representing some subset of DOM stuff.

book/accessibility.md
chrishtr: I would rather we implement important features and not obscure ones.

I added the hover feature a couple of months ago precisely for this reason -- it's an important a11y feature, and "makes use of" the a11y tree in a real way, in the sense that it hit tests it.

The a11y tree could probably have been built with notations on the layout tree about which nodes to skip, etc. But it does seem a bit cleaner to separate the concepts, and also note that it does become more hairy and complicated the more features you add. It's kind of like how in the early days, there probably was not a layout tree and instead browsers hung metadata off of DOM nodes. But nowadays that feels crazy because layout has become complex.

chrishtr:
Just checking: will this PR be abandoned given our side discussion about the browser thread?

pavpanchekha:
Yes. I'm going to keep the PR alive until we merge the screen-reader-browser-thread PR so I can steal words if necessary.

PR #631, opened 2022-09-16
Pass over alert section

pavpanchekha:
This PR reacts to the fact that the `role` attribute is now covered in the accessibility tree section. It also makes a large change to how alerts are implemented. Instead of catching `role` changes in `setAttribute`, we compute the set of all alerts, keep track of which ones we've already read, and read the diff. This is more robust when, for example, an `alert` element is added with `innerHTML`.

Other changes include:
- Point out why we can't read every alert immediately
- Adds detail to a footnote about alert timing
- Gives an example of where an `alert` might be used
- Addresses the case of an element becoming, then un-becoming, then again becoming an alert.

### DIFF
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -2167,7 +2167,8 @@ accessiblity tree for elements with the `alert` role:
 
 ``` {.python}
 class Tab:
-    def __init__(self):
+    def __init__(self, browser):
+        # ...
         self.active_alerts = []
 
     def render(self):
@@ -2185,10 +2186,11 @@ them twice:
 
 ``` {.python}
 class Tab:
-    def __init__(self):
+    def __init__(self, browser):
+        # ...
         self.spoken_alerts = []
 
-    def speak_update(self)
+    def speak_update(self):
         for alert in self.active_alerts:
             if alert not in self.spoken_alerts:
                 self.speak_node(alert, "New alert")
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -2175,7 +2175,7 @@ class Tab:
         if self.needs_accessibility:
             # ...
             self.active_alerts = [
-                node for node in self.accessibility_tree
+                node for node in tree_to_list(self.accessibility_tree, [])
                 if node.role == "alert"
             ]
 ```
@@ -2209,7 +2209,7 @@ class Tab:
             new_spoken_alerts = []
             for old_node in self.spoken_alerts:
                 new_nodes = [
-                    node for node in self.accessiblity_tree
+                    node for node in tree_to_list(self.accessibility_tree, [])
                     if node.node == old_node.node
                     and node.role == "alert"
                 ]
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -2102,27 +2102,29 @@ knowing that it was implemented by a browser is a good sign that it's not
 :::
 
 
-Live Regions
-============
+Accessible alerts
+=================
 
 Scripts do not interact directly with the accessibility tree, much
 like they do not interact directly with the display list. However,
 sometimes scripts need to inform the screen reader about *why* they're
 making certain changes to the page to give screen-reader users a
 better experience. The most common example is an alert[^toast] telling
-you that some action you just did failed. A screen-reader user needs
+you that some action you just did failed. A screen reader user needs
 the alert read to them immediately, no matter where in the document
 it's inserted.
 
 [^toast]: Also called a "toast", because it pops up.
 
-The `alert` role addresses this need. A screen-reader will
-immediately[^alert-css] read an element with that role, no matter
+The `alert` role addresses this need.[^other-live] A screen reader
+will immediately[^alert-css] read an element with that role, no matter
 where in the document the user currently is. Note that there aren't
 any HTML elements whose default role is `alert`, so this requires
-setting the `role` attribute. There are also other "live" roles like
-`status` for less urgent information or `alertdialog` if the keyboard
-focus should move to the alerted element.
+setting the `role` attribute.
+
+[^other-live]: There are also other "live" roles like `status` for
+less urgent information or `alertdialog` if the keyboard focus should
+move to the alerted element.
 
 [^alert-css]: The alert is only triggered if the element is added to
     the document, has the `alert` role (or the equivalent `aria-live`
@@ -2337,7 +2339,7 @@ instead. Implement it.
 
 [hover-pseudo]: https://developer.mozilla.org/en-US/docs/Web/CSS/:hover
 
-* *Alert updates*: Right now, the screen-reader immediately reads an
+* *Alert updates*: Right now, the screen reader immediately reads an
 alert when an element gains the `alert` role. But if that element's
 contents change, the alert isn't re-read. In real browsers, changes to
 an alert cause the alert to be re-read. Implement that in your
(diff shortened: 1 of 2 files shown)


pavpanchekha:
One thing I'd like to do in this section, but haven't in this PR, is to move it right after the screen reader section, before hover. I can't do that until the hover section is cleaned up a bit, so let's merge it here, then move the section later.

book/accessibility.md
```diff
@@ -2095,103 +2102,52 @@ knowing that it was implemented by a browser is a good sign that it's not
...
-        text += " is focused"
-    return text
-```
-
-These alerts are supposed to happen only when the `role` attribute changes
-to alert,^[Or the text contexts of the alert changes, but I won't implement
-that.] so we need a way to detect that an attribute changed. But there isn't
-currently a way to change element attributes other than special ones like
-`style`, so let's first implement that. It will need some runtime code:
+[^aom]: As of this writing, there's a draft [Accessibility Objet
```
chrishtr, line 2118: This was abandoned, you should just remove it.

book/accessibility.md
```diff
@@ -2095,103 +2102,52 @@ knowing that it was implemented by a browser is a good sign that it's not
...
-Now that roles and element tags need not be the same, we need to redefine
-`announce_text` to look only at the computed role, and not the element tag:
+Scripts do not interact directly with the accessibility tree,[^aom]
+much like they do not interact directly with the display list.
+However, sometimes scripts need to inform the screen reader about
+*why* they're making certain changes to the page to give screen-reader
+users a better experience. The most common example is an alert[^toast]
+telling you that some action you just did failed. Typically, these are
+inserted at some fixed place in the document, and thus in the
+accessibility tree; there's little chance a screen-reader user would
```
chrishtr, line 2115: The second half of this sentence is confusing. I think you're trying to say that you can't just rely on the regular document-reading order for alerts, but it's not clear enough.

book/accessibility.md
```diff
@@ -2095,103 +2102,52 @@ knowing that it was implemented by a browser is a good sign that it's not
...
+
+[^alert-css]: The alert is only triggered if the element is added to
+    the document, has the `alert` role (or the equivalent `aria-live`
+    value, `assertive`), and is visible in the layout tree (meaning it
+    doesn't have `display: none`), or if its contents change. In this
+    chapter, I won't handle all of these cases and just focus on new
+    elements with an `alert` role, not changes to contents or CSS.
+    
+Before we jump to implementation, we first need to make it possible
+for scripts to change the `role` attribute. Let's add support for the
```
chrishtr, line 2140:
```suggestion
for scripts to change the `role` attribute. To do that we'll need to add support for the
```

book/accessibility.md
```diff
@@ -2203,38 +2159,73 @@ class JSContext:
...
         elt = self.handle_to_node[handle]
-        if attr == "role" and value == "alert" and \
-            self.getAttribute(handle, attr) != "alert":
-            self.tab.queue_alert(elt)
         elt.attributes[attr] = value
 ```
 
-Which then queues the alert if accessibility is on:
+Now we can implement the `alert` role. To do so, we'll search the
+accessiblity tree for elements with the `alert` role:
```
chrishtr, line 2166:
```suggestion
accessiblity tree for elements with that role:
```

book/accessibility.md
```diff
@@ -2203,38 +2159,73 @@ class JSContext:
...
-        for alert in self.queued_alerts:
-            self.speak_node(alert, "New alert")
+    def __init__(self, browser):
         # ...
+        self.spoken_alerts = []
+
+    def speak_update(self):
+        for alert in self.active_alerts:
+            if alert not in self.spoken_alerts:
+                self.speak_node(alert, "New alert")
```
chrishtr, line 2196: perhaps add an exercise to allow setting the text content of an element in order to update the alert. Otherwise the page will have to remove the alert, wait a rAF loop, then re-add it with new innerHTML.

pavpanchekha:
Done. The confusing sentence is totally rewritten and, I think, much better. In the new exercise I explicitly call out `aria-relevant`.

book/accessibility.md
```diff
@@ -2351,6 +2337,18 @@ instead. Implement it.
 
 [hover-pseudo]: https://developer.mozilla.org/en-US/docs/Web/CSS/:hover
 
+* *Alert updates*: Right now, the screen-reader immediately reads an
+alert when an element gains the `alert` role. But if that element's
+contents change, the alert isn't re-read. In real browsers, changes to
+an alert cause the alert to be re-read. Implement that in your
+browser: if the text description of any node below an `alert` node
+changes in the accessibility tree, re-read the `alert`. For an
```
chrishtr, line 2345: I don't think our runtime currently supports changing out the text without the element...also add that API?

book/accessibility.md
```diff
@@ -2095,103 +2102,47 @@ knowing that it was implemented by a browser is a good sign that it's not
...
-        text += " is focused"
-    return text
-```
-
-These alerts are supposed to happen only when the `role` attribute changes
-to alert,^[Or the text contexts of the alert changes, but I won't implement
-that.] so we need a way to detect that an attribute changed. But there isn't
-currently a way to change element attributes other than special ones like
-`style`, so let's first implement that. It will need some runtime code:
+Live Regions
```
chrishtr, line 2105:
```suggestion
Accessibility alerts
```

book/accessibility.md
```diff
@@ -2095,103 +2102,47 @@ knowing that it was implemented by a browser is a good sign that it's not
...
+============
+
+Scripts do not interact directly with the accessibility tree, much
+like they do not interact directly with the display list. However,
+sometimes scripts need to inform the screen reader about *why* they're
+making certain changes to the page to give screen-reader users a
+better experience. The most common example is an alert[^toast] telling
+you that some action you just did failed. A screen-reader user needs
```
chrishtr, line 2113:
```suggestion
you that some action you just did failed. A screen reader user needs
```

book/accessibility.md
```diff
@@ -2095,103 +2102,47 @@ knowing that it was implemented by a browser is a good sign that it's not
...
+the alert read to them immediately, no matter where in the document
+it's inserted.
+
+[^toast]: Also called a "toast", because it pops up.
+
+The `alert` role addresses this need. A screen-reader will
```
chrishtr, line 2119:
```suggestion
The `alert` role addresses this need. A screen reader will
```

book/accessibility.md
```diff
@@ -2095,103 +2102,47 @@ knowing that it was implemented by a browser is a good sign that it's not
...
+immediately[^alert-css] read an element with that role, no matter
+where in the document the user currently is. Note that there aren't
+any HTML elements whose default role is `alert`, so this requires
+setting the `role` attribute. There are also other "live" roles like
```
chrishtr, line 2123: Put this last sentence in a footnote perhaps?

book/accessibility.md
pavpanchekha: You can change the `innerHTML` of an existing element to regular text. It won't change the element itself, just delete existing children and add a new one.

chrishtr: Right, forgot about that.

chrishtr (approved):

PR #632, opened 2022-09-17
Commit the a11y tree to the browser process and access it there

chrishtr:
This CL still creates the a11y tree on the `Tab` but then sends it over to the `Browser` for actual interaction, without
reference to the `Tab`. The purpose is to demonstrate why there is a tree at all instead of accessing the layout tree.

It also updates the chapter text to match the new implementation, plus adds a paragraph and some footnotes about why
it has to be in the browser thread.

This necessitates making the a11y tree self-contained, which is most of the cocde changes.

### DIFF
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -1500,8 +1500,8 @@ The accessibility tree is built out of `AccessibilityNode`s:
 class AccessibilityNode:
     def __init__(self, node):
         self.node = node
-        self.previous = None
         self.children = []
+        self.text = None
 ```
 
 The `build` method on `AccessibilityNode` recursively creates the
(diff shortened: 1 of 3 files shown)


pavpanchekha (approved):
I think we can go ahead and merge this as is. But a question: in a real browser, is the hover outline drawn by the screen-reader or by the browser or by the tab? If it's by the browser or the screen-reader, why not do it that way here?

chrishtr:
> I think we can go ahead and merge this as is. But a question: in a real browser, is the hover outline drawn by the screen-reader or by the browser or by the tab? If it's by the browser or the screen-reader, why not do it that way here?

By the browser. I mentioned this in a footnote that I'm taking a shortcut. We can add the browser implementation. A reason to do that would be that it requires using the bounds field of `AccessibilityNode` in one more case..

PR #634, opened 2022-09-20
Pass on the new screen-reader section

pavpanchekha:
I did a re-pass over the screen-reader section. I think I fixed a bug. I also removed the `accessibility_is_on` flag on `Tab`; why not just build the accessibility tree every time?

However, I think `has_spoken_document` isn't set back to False when you navigate between tabs. Am I right? One fix for this is to make it a `last_url` field. This wouldn't work for refreshes, but maybe that's OK?

### DIFF
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -1867,7 +1867,7 @@ class Browser:
 ```
 
 There's a lot more in a real screen reader: landmarks, navigating text
-at different granularities, repeating recently spoken text, and so on.
+at different granularities, repeating text when requested, and so on.
 Those features make various uses of the accessibility tree and the
 roles of the various nodes. But since the focus of this book is on the
 browser, not the screen reader itself, let's focus for the rest of
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -1402,8 +1402,8 @@ colors is something that page authors also have to pay attention to.
 :::
 
 
-Screen readers
-==============
+The accessibility tree
+======================
 
 Zoom, dark mode, and focus indicators help users with difficulty
 seeing fine details, but if the user can't see the screen at all,^[The
@@ -1569,14 +1569,14 @@ class AccessibilityNode:
 The user can now direct the screen reader to walk up or down this
 accessibility tree and describe each node to the user. 
 
-Using the accessibility tree
-----------------------------
+Screen readers
+==============
 
-Typically, the screen reader is a separate application from the browser,
-[^why-diff] which the browser communicates with through OS-specific APIs. To
-keep this book platform-independent, our discussion of screen reader support
-will instead include a minimal screen reader integrated directly into the
-browser.[^os-pain]
+Typically, the screen reader is a separate application from the
+browser,[^why-diff] which the browser communicates with through
+OS-specific APIs. To keep this book platform-independent, our
+discussion of screen reader support will instead include a minimal
+screen reader integrated directly into the browser.[^os-pain]
 
 But should our built-in screen reader live in the `Browser` or each `Tab`? Real
 browsers implement it in the `Browser`, so we'll do that too.^[And therefore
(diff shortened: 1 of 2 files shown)


book/accessibility.md
```diff
@@ -1569,14 +1569,14 @@ class AccessibilityNode:
...
+Screen readers
+==============
 
-Typically, the screen reader is a separate application from the browser,
-[^why-diff] which the browser communicates with through OS-specific APIs. To
-keep this book platform-independent, our discussion of screen reader support
-will instead include a minimal screen reader integrated directly into the
-browser.[^os-pain]
+Typically, the screen reader is a separate application from the
+browser,[^why-diff] which the browser communicates with through
```
chrishtr, line 1576:
```suggestion
browser,[^why-diff] with which the browser communicates through
```

book/accessibility.md
```diff
@@ -1812,65 +1802,77 @@ class Browser:
...
+            text += " " + \
+            node.children[0].text
+
+        if text:
+            speak_text(text)
+```
+
+There's a lot more in a real screen reader: landmarks, navigating text
+at different granularities, repeating recently spoken text, and so on.
+Those features make various use of the accessibility tree and the
```
chrishtr, line 1871:
```suggestion
Those features make various uses of the accessibility tree and the
```

src/lab14.py
```diff
@@ -613,7 +613,28 @@ def build(self):
         for child_node in self.node.children:
             self.build_internal(child_node)
 
-        self.text = announce_text(self.node, self.role)
```
chrishtr, line 616: Are you deleting announce_text?

pavpanchekha: Woops, yes

pavpanchekha:
Ah, no, I can't delete `announce_text` yet, it's used in the section on alerts, which is a separate PR.

book/accessibility.md
```diff
@@ -1760,31 +1734,23 @@ class Browser:
...
-    elif node.tag == "button":
-        text = "Button"
-    elif node.tag == "link":
-```
+To speak the whole document, we need to know how to speak each
+`AccessibilityNode`. This has to be decided back in the `Tab`, since
+the text will include DOM content that is not accessible to the
+browser thread. So let's add a `text` field to `AccessibilityNode` and
+set it according to role and surrounding DOM context. For each node,
+we'll figure out its text via the `announce_text` function. For text
```
chrishtr, line 1749: Rephrase to remove `announce_text`

chrishtr:
> However, I think has_spoken_document isn't set back to False when you navigate between tabs. Am I right?

I think so.

> One fix for this is to make it a last_url field. This wouldn't work for refreshes, but maybe that's OK?

I think something simple (maybe even what's already implemented) is good enough.

book/accessibility.md
```diff
@@ -1812,65 +1801,77 @@ class Browser:
...
 ```
 
-Let's next use this to also speak the focused element to the user. But right now
-the browser thread doesn't actually know what in the `Tab` is focused, so we'll
-need to commit that to a `tab_focus` field on `Browser` as well.^[By adding a
-`focus` field to `CommitData`, etc., just like the accessibilty tree, and
-storing it in a new `tab_focus` property on the `Browser`. Since it's
-repetitive I've omitted that code.]
+Speaking the whole document happens only once. But the user might need
+feedback as they browse the page. For example, let's speak the focused
```
chrishtr, line 1817:
```suggestion
feedback as they browse the page, such as which element is focused. Let's speak the focused
```

book/accessibility.md
```diff
@@ -1812,65 +1801,77 @@ class Browser:
...
+            node.children[0].role == "StaticText":
+            text += " " + \
+            node.children[0].text
+
+        if text:
+            speak_text(text)
+```
+
+There's a lot more in a real screen reader: landmarks, navigating text
+at different granularities, repeating recently spoken text, and so on.
```
chrishtr, line 1869: *not* repeating recently spoken text?

chrishtr (approved):
Approved once those comments I just left are resolved.

book/accessibility.md
pavpanchekha: Screen readers have key bindings for "repeat that word again" or "repeat that sentence you just said", which is what I meant to point to here. Can you suggest a better phrasing?

book/accessibility.md
pavpanchekha: I rephrased in a different way.

PR #642, opened 2022-09-28
Pass over the hover section

pavpanchekha:
This PR reorganizes the flow in the hover section a bit but the only two code changes are making `layout_object` a proper field on `Element` (to avoid using `hasattr`, which I'm trying to avoid in the book) and rewriting `hit_test` to be a bit simpler.

### DIFF
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -1200,7 +1200,7 @@ class Tab:
     def run_animation_frame(self, scroll):
         # ...
         if self.focus_changed and self.focus:
-            if hasattr(self.focus, "layout_object"):
+            if self.focus.layout_object:
                 layout_object = self.focus.layout_object
                 if layout_object.y - self.scroll < 0:
                     self.scroll = \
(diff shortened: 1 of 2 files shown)


book/accessibility.md
```diff
@@ -2122,99 +2122,161 @@ also defining new and fully stylable [form control elements][openui].
...
-the user wants to [hit test][hit-test] a place on the screen to see what is
-there:  user who can't see the screen still might want to do things like touch
-exploration of the screen, or being notified what is under the mouse as they
-move it around.
+Thanks to our work in this chapter, our rendering pipeline now
+basically have two different outputs: a display list for visual
+interaction, and an accessibility tree for screen-reader interaction.
+Many users will use just one or the other. However, it can also be
+valuable to use both together. For example, a user might have limited
+vision, able to make out the general items on a web page but unable to
```
chrishtr, line 2134:
```suggestion
vision---able to make out the general items on a web page but unable to
```

book/accessibility.md
```diff
@@ -2122,99 +2122,161 @@ also defining new and fully stylable [form control elements][openui].
...
+        # ...
+        self.layout_object = None
 ```
 
+Now, when we construct a layout object, we can fill in the
+`layout_object` field of its `Element`. In `BlockLayout`, it looks
+like this:
+
 ``` {.python}
+class Element:
```
chrishtr, line 2171: remove?

chrishtr (approved):

book/accessibility.md
pavpanchekha: Why? You mean, just don't show the code?

pavpanchekha: My inclination is to show it. If it's a bit obvious, that also means it'll be easy to read, which means the cost of showing it is low, but the benefit to the uncareful reader is potentially large.

chrishtr: This code block says:

class Element:
 
class BlockLayout:
   .. some code

Why is the "class Element" part there? not sure what it indicates.

pavpanchekha: Oh wow, that was such a crazy mistake I totally missed it. Thanks!!!

PR #643, opened 2022-09-30
Fix iframes

pavpanchekha:
This PR stops embedding the Chapter 9 and 10 widgets. This is necessary to fix iframes, for complicated reasons:

- Our widget, located in an `iframe`, requires `SharedArrayBuffer`
- That requires the top-level document (the book chapter) to be `crossOriginIsolated`
- That requires the top-level document to set its `Cross-Origin-Embedder-Policy` to `require-corp`
- That bans all iframes that don't send a `Cross-Origin-Resource-Policy` header
- Substack and Youtube don't send that header

There are some work-arounds, like credentialless mode, but they only work in Chrome right now.

So I've removed the COEP header from the server, and this PR removes the iframes for the JS and Security chapters, and replaces them with a note that you have to open them in their own window. If open in their own window, the server will send the COEP header and they'll work.

Longer term I'm going to investigate using a busy-wait loop around `localStorage` for the synchronous communication instead. It'll be much more CPU-intensive, which is a big loss. @chrishtr perhaps you can ask someone you know for any other workarounds?

### DIFF
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -2142,8 +2142,6 @@ Now, when we construct a layout object, we can fill in the
 like this:
 
 ``` {.python}
-class Element:
-    
 class BlockLayout:
     def __init__(self, node, parent, previous):
         # ...
--- a/book/scripts.md
+++ b/book/scripts.md
@@ -1137,8 +1137,8 @@ The server's outline is unchanged from the last chapter:
 :::
 
 If you run it, it should look something like [this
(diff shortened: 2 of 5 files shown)


chrishtr:
I've asked a couple of Chrome security experts for advice, will come back with what they say soon.

pavpanchekha:
Ok. I've pushed this PR to the website as a stop-gap, let's hear from them and modify or merge.

chrishtr (approved):

PR #645, opened 2022-10-01
Editorial and spelling fixes; fix to hit testing

chrishtr:
I also added a paragraph at the beginning of the chapter putting the concerns of permanently-disabled people
first. This is in response to feedback from an a11y expert review who suggested that change.

### DIFF
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -24,9 +24,9 @@ What is accessibility?
 
 Accessibility means that the user can change or customize how they
 interact with a web page in order to make it easier to
-use.[^other-defs][^not-just-screen-reader] The web's uniquely-flexible
+use.[^other-defs] The web's uniquely-flexible
 core technologies mean that browsers offer a lot of accessibility
-features[^not-just-screen-reader] to allow a user to customize the
+features[^not-just-screen-reader] that allow a user to customize the
 style, layout, and rendering of a web page, as well as interact with a
 web page with their keyboard, by voice, or using some kind of helper
 software.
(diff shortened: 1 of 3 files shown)


src/lab14.py
```diff
@@ -668,9 +668,10 @@ def intersects(self, x, y):
         return False
 
     def hit_test(self, x, y):
-        for node in tree_to_list(self, []):
-            if node.bounds.intersects(x, y):
-                return node
+        nodes = [node for node in tree_to_list(self, [])
```
chrishtr, line 671: The old code was wrong because you need to hit test from the bottom of the tree up.

src/lab14.py
```diff
@@ -1742,7 +1739,8 @@ def handle_click(self, e):
         self.lock.release()
 
     def handle_hover(self, event):
-        if not self.accessibility_is_on:
+        if not self.accessibility_is_on or \
```
chrishtr, line 1742: This fixes a race condition that happened sometimes when enabling a11y.

book/accessibility.md
```diff
@@ -106,7 +115,7 @@ In the United States, the European Union, and many other countries,
 website accessibility is legally required. For example, United States
 Government websites are required to be accessible under [Section
 508][sec508] of the [Rehabilitation Act Amendments of
-1993][rehab-act], and the government has a bunch of
+1973][rehab-act], and the government has a bunch of
```
pavpanchekha, line 118: I think, confusingly, these are two similarly-named acts. We should double-triple check.

book/accessibility.md
```diff
@@ -883,7 +894,7 @@ class Tab:
 
 Finally, we just need to focus on the chosen element. If we've reached
 the last the focusable node (or if there weren't any focusable nodes
-to begin with), we'll unfocus the page and move focus to the address
+to begin with), we'll UN-focus the page and move focus to the address
```
pavpanchekha, line 897: Would prefer italics rather than capitalization for emphasis here.

src/lab14.py
pavpanchekha: Oh yeah, you are right.

pavpanchekha (approved):
Please consider at least re-checking which Rehabilitation Act we want. I think a semi-authoritative source is this: https://www.section508.gov/manage/laws-and-policies/

Which makes me think it is 1998, not 1993 but also not 1973, that we need to cite.

pavpanchekha:
And yes, the alert examples works for me in VoiceOver+Firefox

pavpanchekha:
But it doesn't work in VoiceOver+Safari.

pavpanchekha:
I guess either way we should explain that this example will only work in some browsers and screen readers.

pavpanchekha:
Well, to finish data collection, the example works on Chrome and Firefox, but not Safari, with VoiceOver.

book/accessibility.md
chrishtr: I looked at the site you recommended, and it's the ...act of 1973, with amendments made in 1998. I added a parenthetical.

book/accessibility.md
chrishtr: That was a typo - changed to "un-focus"

PR #646, opened 2022-10-02
fixed typo in preface

PaulineNemchak:
I fixed a small typo.
I've added `.idea` to `.gitignore` in case I will add any other fixes, but you can reject this change and just fix the typo.


chrishtr:
Thanks.

PR #648, opened 2022-10-04
Some typo and editorial fixes

chrishtr:


book/accessibility.md
```diff
@@ -2145,13 +2149,13 @@ be configured on a more granular level by setting their "politeness"
...
 `aria-atomic` and `aria-relevant`; and whether the live region is in a
-finished or intermediate state, via `aria-busy`. In fact, `aria-live`
+finished or intermediate state, via `aria-busy`. In addition, `aria-live`
 is all that's necessary to create a live region; no role is necessary.
 
 :::
 
 
-Mixed voice / visual interaction
+Voice + visual interaction
```
pavpanchekha, line 2158: Would prefer to write out "and" here. (Here's the difference I was taught: the slash is a punctuation character, you read it as a small pause, but the plus is read as "and" so should be written out in words. Frankly, these nitpicky rules don't matter but perhaps it keeps the book to a consistent style.)

pavpanchekha (approved):
LGTM, one super dumb tiny nitpicky comment.

book/accessibility.md
chrishtr: I changed to & instead. The goal is just to make it fit on one line.

PR #655, opened 2022-11-02
Fix two small bugs

pavpanchekha:
This PR fixes two minor bugs.

1) We need to import `ANIMATED_PROPERTIES` into lab 14, because we use it
2) The `TaskRunner` deadlocks when `needs_quit` is set but the task queue is empty.


chrishtr:
Why closed?

pavpanchekha:
Bad branch. Will redo, one second :)

PR #659, opened 2022-11-08
Unify InlineLayout and BlockLayout

pavpanchekha:
I've been working on invalidation. I've run into an annoying issue that goes back to a decision made in Chapter 5, which I think we should revisit. I can continue working on invalidation without messing with this decision, so it's not urgent, but I'd like your thoughts.

*Chapter 5* is the Layout chapter. It introduces two classes: BlockLayout and InlineLayout. Despite the name, they both represent what you and I would call "block" boxes; BlockLayout represents block boxes with block children, while InlineLayout represents block boxes with inline children.

In the WBE browser, layout objects are created by its parent layout object. So the parent layout object looks at each of its children, determines whether that child's children (the parent block's grandchildren) make it BlockLayout or InlineLayout, and instantiates the relevant class.

*The problem* is that a node can switch between needing BlockLayout or InlineLayout, for example if an empty block gains text content (BlockLayout to InlineLayout) or if a text block gains <div> children (InlineLayout to BlockLayout). But recall that the decision of which class to use is actually made in the node's parent's layout object. So once we add dirty bits, we need to rerun this logic if any child is dirty, which also recreates layout objects for any siblings without need. (Or we use even more complex matching logic...)

If, instead, each node decided for itself whether it was using block or inline layout inside itself, a node could only switch if its list of children changed, and in WBE that always involves the node itself (because it's basically only `innerHTML`). This makes managing dirty bits easy. In a more complex browser, such as something we might want to discuss in an exercise, the previous layout mode could be cached and compared to determine dirtiness.

*The solution* is to unify the InlineLayout and BlockLayout classes. It'll be a single class, called BlockLayout I suppose, which will test its own node's layout mode and do either block or inline layout for the children.
Why do it? Normally I wouldn't change Chapter 5 to make Chapter 16 work, but in this case I think there are compelling reasons. The split between BlockLayout and InlineLayout is already super confusing, and I think Chapter 5 is consistently one of the most confusing. So I'm not too worried about breaking something good. Plus, loads of other chapters currently request the reader make identical changes to BlockLayout and InlineLayout. If they were one class, this would go away.

In terms of code changes, this requires limited changes in Chapter 5 (because a lot of code is already copy-and-pasted between the classes) and some savings in later chapters. The text changes are, however, more substantial, and deserve careful review.

*This PR* shows the relevant code changes as a start. If you approve of the idea, I'll work on propagating this through later chapters and making the relevant text changes, for eventually merging.


pavpanchekha:
@chrishtr here I'm requesting a review of the general idea—to merge this we'll need to do changes in a lot of other chapters, make the text fit the new code, and so on. Code for Chapter 5 hopefully gives you something concrete to look at to gauge whether this is a good idea. Happy to discuss in more detail Friday.

chrishtr:
I think this is a good idea, since as you say BlockLayout and InlineLayout are both blocks.

This seems very similar to [LayoutBlockFlow](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/layout/layout_block_flow.h?q=layoutblockflow&ss=chromium), the Chromium class that implements a block container, and has all inline or block descendants (which is also the same rule we have for `layout_mode`).

If it simplifies the dirty bits, all the better.

Random aside: Chromium also recently (this year) changed its implementation of the mixed-block-and-inline-siblings problem to get rid of continuations and instead use anonymous inline wrappers for blocks.

pavpanchekha:
Great, I'll work on finishing out the PR.

Just to make sure I understand with the recent Chromium change—you mean wrap inline elements in an anonymous block box? Or wrapping blocks with anonymous inline wrappers? The first one makes sense to me, the second one seems pretty weird!

pavpanchekha:
Ok: I've pushed fixes to Chapters 5–14 to unify BlockLayout and InlineLayout across all of them.

I *also* discovered that Chapters 13 and 14 contain implementations of CSS `width`/`height` properties even though that's been removed from the text. I removed those implementations. (If you want that as a separate PR that's probably doable.)

I haven't yet touched the text; I'll be working on that next.

pavpanchekha:
Looks like Chapter 5 will need major edits, Chapter 7 minor edits, and every other chapter just very small fix ups.

pavpanchekha:
Ok, only Chapter 5 left, which will require major reworking. Having looked over it, I think it's no worse, maybe a little better, if we make the `cursor_x/y` variables relative, so I went ahead and did that.

pavpanchekha:
Ok, I've now rewritten Chapter 5 and now the whole book uses the unified `BlockLayout`. I'll work on resolving conflicts but the code is ready for review.

The Chapter 5 edits were a bit bigger than expected.

Basically, in the current version of the book, we introduce `BlockLayout` first while keeping the existing inline layout implementation (called `Layout`) on the side, and eventually convert that into `InlineLayout`. I first tried keeping that structure but it was too weird, because instead of converting `Layout` into a separate `InlineLayout` class it would have to get merged into the `BlockLayout` class.

So instead, I now start by converting `Layout` into a fully-fledged layout object, and then add the normal block layout mode later, basically flipping the order of the first two sections of the book. This required some reworking, but I think the chapter still basically hangs together.

pavpanchekha:
I think the failure in the CI isn't our fault, and if we wait a bit it'll fix itself.

book/layout.md
```diff
@@ -30,103 +30,102 @@ to produce the layout tree, then computes the size and position for
 each layout object, and finally draws each layout object to the
 screen.
 
-[^no-box]: Elements like `<script>` don't generate layout objects, and
```
chrishtr, line 33: Isn't this still referenced on line 27?

book/layout.md
```diff
@@ -30,103 +30,102 @@ to produce the layout tree, then computes the size and position for
...
         self.children.append(child)
         child.layout()
+        self.display_list = child.display_list
 ```
 
-So we're building a layout tree with one layout object per HTML node,
-plus an extra layout object at the root, by recursively calling
-`layout`. It looks like this:
+Note an interesting thing about this new `layout` method: its role is
+to _create_ the child layout objects, and then _recursively_ calls
```
chrishtr, line 108:
```suggestion
to _create_ the child layout objects, and then _recursively_ call
```

book/layout.md
```diff
@@ -232,54 +225,66 @@ BLOCK_ELEMENTS = [
 ]
 ```
 
-We'll use `BlockLayout` for elements with children in that list, and
-`InlineLayout` otherwise. Put that logic in a new `layout_mode`
-function:
+Our `layout_mode` function has to handle one tricky case, where a node
+contains both block objects like a `<p>` element but also text objects
```
chrishtr, line 229:
```suggestion
contains both block children like a `<p>` element but also text ones
```

book/layout.md
```diff
@@ -232,54 +225,66 @@ BLOCK_ELEMENTS = [
...
-elements.
+``` {.python expected=False}
+class BlockLayout:
+    def layout(self):
+        # ...
+        for child in self.children:
+            self.display_list.extend(child.display_list)
+```
 
-::: {.widget big-height=490px small-height=860px}
```
chrishtr, line 277: Don't want this example any more?

book/layout.md
```diff
@@ -296,87 +301,154 @@ actually implementing inline and block layout.
 Size and position
 =================
 
-By default, layout objects are greedy and take up all the horizontal
-space they can.[^until-css] So their width is their parent's width:
+In the [previous chapter](html.md), the `Layout` object was
+responsible for whole web page, so it just laid out its content
```
chrishtr, line 305:
```suggestion
responsible for the whole web page, so it just laid out its content
```

book/layout.md
```diff
@@ -296,87 +301,154 @@ actually implementing inline and block layout.
...
+starting at the top of the page. Now that we have multiple
+`BlockLayout` objects each containing a different paragraph of text,
+we're going to need to do things a little differently, computing a
+size and position for each layout object independently.
 
-[^until-css]: In the [next chapter](styles.md), we'll add support for
-author-defined styles, which modify these rules and allow setting custom widths,
-borders, or padding.
+Let's start with the `cursor_x` and `cursor_y`. Instead of having them
```
chrishtr, line 311:
```suggestion
Let's start with `cursor_x` and `cursor_y`. Instead of having them
```

book/layout.md
```diff
@@ -296,87 +301,154 @@ actually implementing inline and block layout.
...
-An `InlineLayout` computes `width`, `x`, and `y` the same way, but
-`height` is a little different: an `InlineLayout` has to contain all
-of the text inside it, which means its height must be computed from
-its *y*-cursor.
+However, a `BlockLayout` that contains text doesn't have children;
+instead, it needs to be tall enough to contain all its text, which we
+can conveniently read off of `cursor_y`:[^why-two-fields]
+
+[^why-two-fields]: Since the height is just equal to `cursor_y`, why
+    not renamed `cursor_y` to `height` instead? You could, it would
```
chrishtr, line 426:
```suggestion
    not rename `cursor_y` to `height` instead? You could, it would
```

book/layout.md
```diff
@@ -296,87 +301,154 @@ actually implementing inline and block layout.
...
         # ...
-        self.height = self.cursor_y - self.y
+        else:
+            self.height = self.cursor_y
 ```
 
-Again, `width`, `x`, and `y` have to be computed before text is laid
-out, but `height` has to be computed after. It's all about that
-dependency order.
+Let's think again about dependencies. While the `x`, `y`, and `width`
```
chrishtr, line 444: Is this paragraph partially repeating the one starting at line 384?

book/layout.md
pavpanchekha: No, you're right, this one is still good (but I need to edit the caption). The other one (which had just `BlockLayout`s) is no longer good, since now the first stage is only inline, not only block.

pavpanchekha:
I fixed all the comments except the duplicate paragraph, which I'll look into when I have a moment.

pavpanchekha:
I fixed the redundant paragraph by deleting the repetitive parts and focusing on how `height` is different from `x`, `y`, and `width`.

book/layout.md
```diff
@@ -139,83 +143,77 @@ Safari a [render tree][webkit-tree], and Firefox a [frame tree][gecko-tree].
...
-we'll implement something a little simpler.
-
-[anon-block]: https://developer.mozilla.org/en-US/docs/Web/CSS/Visual_formatting_model#anonymous_boxes
-
-Here's a list of block elements:[^from-the-spec]
+Here the list of `BLOCK_ELEMENTS` is basically what you expect, a list
+of all the tags that describe parts of a page instead of
+formatting:[^from-the-spec]
 
 [^from-the-spec]: Taken from the [HTML5 living standard][html5-elts].
```
chrishtr, line 218:
```suggestion
[^from-the-spec]: Taken from the [HTML living standard][html5-elts].
```

book/layout.md
```diff
@@ -232,54 +230,70 @@ BLOCK_ELEMENTS = [
...
+Our `layout_mode` function has to handle one tricky case, where a node
+contains both block children like a `<p>` element but also text
+children like a text node or a `<b>` element. I've chosen to use block
+mode in this case, but it's probably best to think of this as a kind
+of error on the part of the web developer. And just like with implicit
+tags in [Chapter 4](html.md), we use a repair mechanism to make sense
+of the situation.[^anon-block]
+
+[^anon-block]: In real browsers, that repair mechanism is called
+"[anonymous block boxes][anon-block]" and is more complex than what's
```
chrishtr, line 242: This link doesn't work in the rendered version.

book/layout.md
```diff
@@ -232,54 +230,70 @@ BLOCK_ELEMENTS = [
...
+        for child in self.children:
+            self.display_list.extend(child.display_list)
+```
+
+Our browser is now constructing a whole tree of `BlockLayout` objects;
+in fact, if you add a `print_tree` call to `Browser`'s `load` method,
+you'll see that large web pages like this chapter produce large and
+complex layout trees!
 
 ::: {.widget big-height=490px small-height=860px}
```
chrishtr, line 290: This widget is out of date - it references InlineLayout.

book/layout.md
```diff
@@ -296,87 +310,151 @@ actually implementing inline and block layout.
...
+class BlockLayout:
+    def text(self, node):
+        for word in node.text.split():
+            # ...
+            if self.cursor_x + w > self.width:
+                # ...
+            # ...
+```
+
+So now that sets us the problem of computing these `x`, `y`, and
```
chrishtr, line 367:
```suggestion
So now that leaves us with the problem of computing these `x`, `y`, and
```

book/layout.md
```diff
@@ -296,87 +310,151 @@ actually implementing inline and block layout.
...
+Height, meanwhile, is a little tricky. A `BlockLayout` that contains
+other blocks should be tall enough to contain all of its children, so
+its height should be the sum of its children's heights:
+
+``` {.python}
+class BlockLayout:
+    def layout(self):
+        # ...
+        if mode == "block":
+            self.height = sum([child.height for child in self.children])
```
chrishtr, line 427: Needs more line wrapping to avoid a horizontal scrollbar.

book/layout.md
```diff
@@ -399,22 +477,30 @@ run into the very edge of the window and get cut off.
...
 
++ When `layout` is called, it first computes the `width`, `x`, and `y`
+  fields, reading from the `parent` and `previous` layout objects.
++ Next, it creates a child layout object for each child element.
++ Then, the child layout nodes are recursively laid out by calling
+  their `layout` methods.
++ Finally, `layout` computes the `height` field, reading from the
+  child layout objects.
+
+You can see these steps in action in this widget:
```
chrishtr, line 488: This widget is broken in my local build on your branch.

book/layout.md
```diff
@@ -468,11 +548,15 @@ class BlockLayout:
             child.paint(display_list)
 ```
 
-Finally, `InlineLayout` is already storing things to paint in its
-`display_list` variable, so we can copy them over:
+You can now delete the line that computes a `BlockLayout`'s
```
chrishtr, line 551: Rephrase? Repetitive with the previous "you can now" sentence.

pavpanchekha:
Good catch on the widgets. Fixed.

book/layout.md
```diff
@@ -232,54 +230,72 @@ BLOCK_ELEMENTS = [
...
+```
 
-[^or-empty]: Or, the leaf nodes could be `BlockLayout`s for empty
-elements.
+Our browser is now constructing a whole tree of `BlockLayout` objects;
+in fact, if you add a `print_tree` call to `Browser`'s `load` method,
+you'll see that large web pages like this chapter produce large and
+complex layout trees!
 
 ::: {.widget big-height=490px small-height=860px}
```
chrishtr, line 292: The widget iframe is still not tall enough to avoid a vertical scrollbar.

book/layout.md
```diff
@@ -296,87 +312,152 @@ actually implementing inline and block layout.
...
-the parent and previous sibling, so the parent has to compute it
-before recursing, and when recursing it has to lay out the children in
-order.
+Note that in each of these cases, to compute one block's `x` and `y`,
+the `x` and `y` of its parent block must _already_ have been computed.
+That means these computations have to go *before* the recursive
+`layout` call, so those children can compute their `x` and `y` based
+on this block's `x` and `y`. Similarly, since the `y` position of a
+block depends on its previous sibling's `y` position, the recursive
+`layout` calls have to start and the first sibling and iterate through
```
chrishtr, line 401:
```suggestion
`layout` calls have to start at the first sibling and iterate through
```

book/layout.md
```diff
@@ -399,22 +480,30 @@ run into the very edge of the window and get cut off.
...
++ When `layout` is called, it first computes the `width`, `x`, and `y`
+  fields, reading from the `parent` and `previous` layout objects.
++ Next, it creates a child layout object for each child element.
++ Then, the child layout nodes are recursively laid out by calling
+  their `layout` methods.
++ Finally, `layout` computes the `height` field, reading from the
+  child layout objects.
+
+You can see these steps in action in this widget:
+
```
chrishtr, line 492: Make this widget taller to avoid vertical scrolling.

pavpanchekha:
Fixed, the two widgets had different issues but in both cases I fixed the underlying issue, not just changed the widget height.

PR #661, opened 2022-11-16
Add tests to catch + fix a bug in Chapter 4

pavpanchekha:
Chapter 4 has a bug: it doesn't call `flush` when it's done processing a web page, meaning that in some circumstances the last line of text won't be printed. Classic buffer problem! I added tests for it and fixed it. Interestingly, the code for this never quite appears in the book, because it's a simple modification of Chapter 3, and Chapter 3 does have the necessary `flush` call.

### DIFF
--- a/src/lab4-tests.md
+++ b/src/lab4-tests.md
@@ -61,9 +61,29 @@ Missing end tags are added:
 
 Attributes can be set on tags:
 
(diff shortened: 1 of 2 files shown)


chrishtr (approved):

chrishtr:
Is this bug just in chapter 4?

PR #662, opened 2022-11-17
Update blog link

chrishtr:


book/index.md
```diff
@@ -17,7 +17,7 @@ misbehave. Read them at your own risk.
 ::: {.signup}
 :::
 
-Follow this book's [blog](blog/) or
```
chrishtr, line 20: We should also take down the blog/ URL and add a redirect from it to substack. Agree?

pavpanchekha: Absolutely, good find.

PR #666, opened 2022-11-21
Fix animation examples in chapter 13 to not need width/height

chrishtr:


book/animations.md
```diff
@@ -2260,7 +2260,9 @@ and use it to cause animated scroll in `handle_down`, by delegating scroll to
 a main thread animation.[^main-thread-scroll] You'll need to implement a new
 `ScrollAnimation` class and some logic in `run_animation_frame`. Scrolling in
 the [transform transition](examples/example13-transform-transition.html)
-example should now be smooth, as that example uses `scroll-behavior`.
+example should now be smooth, as that example uses `scroll-behavior`. (You'll
+have to have done the width/height exercise to see a scroll, because of the
```
pavpanchekha, line 2264: Would be good to name the chapter and link it.

www/examples/example13-transform-transition.html
```diff
@@ -1,7 +1,8 @@
 <link rel="stylesheet" href="example13-transform-transition.css">
 <body style="scroll-behavior:smooth">
+<div style="height:100px"></div>
```
pavpanchekha, line 3: Can we make this work without `height` using, like, `<br>` and some small character like a period?

pavpanchekha (approved):

PR #668, opened 2022-11-23
Commit the start of Chapter 16

pavpanchekha:
This PR has two things in it. The first is `reflow-test.py`. This is just a file where I'm testing invalidation implementations; it'll be deleted later, so please ignore it. The second and more important thing is a tentative start to `invalidation.md`, our future Chapter 16. Ideally you'd read over it and give any general feedback on the approach.

This is a _very_ rough draft so I don't think it's worth fixing typos and phrasing issues at the moment. But it is important to know whether you think I'm explaining ideas like idempotency and dirty flags in a clear way, and whether you think the emphasis on correctness and approaching things in a systematic way comes across.

### DIFF
--- a/book/embeds.md
+++ b/book/embeds.md
@@ -2,7 +2,7 @@
 title: Supporting Embedded Content
 chapter: 15
 prev: accessibility
-next: skipped
+next: invalidation
 ...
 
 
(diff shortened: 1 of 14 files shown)


pavpanchekha:
In retrospect, let me pull just the chapter into a new text and keep the "testing" code separate.

PR #670, opened 2022-11-23
Commit the start of Chapter 16

pavpanchekha:
This PR contains a tentative start to invalidation.md, our future Chapter 16. Ideally you'd read over it and give any general feedback on the approach. This is a very rough draft so I don't think it's worth fixing typos and phrasing issues at the moment. But it is important to know whether you think I'm explaining ideas like idempotency and dirty flags in a clear way, and whether you think the emphasis on correctness and approaching things in a systematic way comes across.

### DIFF
--- a/book/invalidation.md
+++ b/book/invalidation.md
@@ -1,0 +1,379 @@
+---
+title: Reusing Previous Computations
+chapter: 16
+prev: embeds
+next: skipped
+...
+
+So far, we've focused on improving animation performance by improving
+our browser's graphics subsystem. But that hasn't helped
+layout-inducing animations, since those typically spend most of their
+time in layout, not graphics. Yet just like compositing enabled smooth
+animations by avoiding redundant raster work, we can avoid redundant
+layout work using a technique called invalidation.
+
+Idempotency
+===========
+
+Invalidation is a pretty simple idea: don't redo redundant layout
+work. It means not recreating layout objects if the existing ones are
+already the right ones, and not recomputing layout values (like a
+width or a *y* value) if it is already up to date. Yet invalidation
+has a well-earned reputation as a rats' nest, with even real browsers
+having dozens of known invalidation bugs as of this writing. That's
+because of a real risk that an invalidation bug accidentally forgets
+to recompute a value that changed, causing layout glitches, or
+recomputes way too many values that didn't change, dramatically
+slowing down layout.
+
+So in this chapter, the challenge won't just be implementing
+invalidation. It'll be making sure that we implement invalidation
+_correctly_. That's going to involve thinking carefully about how each
+layout value is computed, and what values it depends on.
+
+Let's start by thinking about how layout objects are created. Right
+now, layout objects are created by `Tab`s when `render` is called:
+
+``` {.python}
+class Tab:
+    def render(self):
+        if self.needs_layout:
+            self.document = DocumentLayout(self.nodes)
+            self.document.layout(self.zoom)
+            # ...
+```
+
+Every time layout runs, the whole layout tree is discarded and a new
+one is created, starting with the root `DocumentLayout`. But this is
+wasteful: when layout _does_ need to run, it's typically due to a
+small change, like `innerHTML` or `style` being set on some element.
+Recreating the layout tree is wasteful, and we'd like to avoid it.
+
+But before jumping right to coding, let's read over the existing code
+to understand where layout objects are created: search the code for
+`Layout`, which all layout class names end with. You should see that
+layout objects are created only in a few places:
+
+- `DocumentLayout` objects are created by the `Tab` in `render`
+- `BlockLayout` objects are created by either:
+  - A `DocumentLayout`, in `layout`
+  - A `BlockLayout`, in `layout`
+- `LineLayout` objects are created by `BlockLayout`s in `new_line`
+- `TextLayout` objects are created by `BlockLayout`s in `text`
+- `InputLayout` objects are created by `BlockLayout`s in `input`
+
+Let's start with the first location. The `DocumentLayout` is created
+the same way each time, and its argument, `nodes`, is never assigned
+to once the page finishes loading. And the `DocumentLayout`
+constructor just copies its arguments into the `DocumentLayout`'s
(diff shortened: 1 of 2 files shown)


book/invalidation.md
```diff
@@ -0,0 +1,379 @@
...
+Here, the arguments to `BlockLayout` are a little more complicated
+than the ones passed in `DocumentLayout`. The second argument, `self`,
+is never assigned to, but `child` and `previous` both contain elements
+read from `node.children`, and that `children` array can change as a
+result of `innerHTML`. Moreover, in order to even run this code, the
+node's `layout_mode` has to be `"block"`, and `layout_mode` itself
+also reads the node's `children`.[^and-tags]
+
+[^and-tags]: It also looks at the node's `tag` and the node's
+    childrens' `tag`s, but luckily in our browser a node's `tag` can't
```
chrishtr, line 223: random note: tags can't change in a real browser without destroying the DOM node.

chrishtr: no such DOM API

chrishtr: literally this (implicitly in the context of idempotence, without saying that word) came up in a design review for a new DOM API a month ago.

book/invalidation.md
```diff
@@ -0,0 +1,379 @@
...
+    change, so we don't need to worry about this. If our browser
+    supported changing a node's `tag`, that's something we would need
+    to worry about, but to my knowledge no real browser does that,
+    even in APIs like `outerHTML`.
+
+So to avoid redundant `BlockLayout` creation, we're going to have to
+know whether the `children` array has changed. To do that, let's add a
+boolean field called a "dirty flag":
```
chrishtr, line 231: chapter 13 has these

book/invalidation.md
```diff
@@ -0,0 +1,379 @@
...
+also reads the node's `children`.[^and-tags]
+
+[^and-tags]: It also looks at the node's `tag` and the node's
+    childrens' `tag`s, but luckily in our browser a node's `tag` can't
+    change, so we don't need to worry about this. If our browser
+    supported changing a node's `tag`, that's something we would need
+    to worry about, but to my knowledge no real browser does that,
+    even in APIs like `outerHTML`.
+
+So to avoid redundant `BlockLayout` creation, we're going to have to
```
chrishtr, line 229: I think you should tie this back to idempotence more clearly:
- Idempotence means you can store off the result of the first function call in a cache 
- The functions should be idempotent (and if not you should go edit them to be so), but that doesn't mean the output doesn't change if the input does
- When the input changes, you have to pair idempotence of the function with a dirty flag indicating the input might have changed.
- You might want more than one dirty flags to reduce the likelihood of an invalidation

These together form the basis for a layout invalidation strategy.

These points together are a lot to understand at once, so at this point I think the first 3 bullets and how they are related are the important things to explain.

chrishtr: Maybe this justifies pseudocode before getting to the main action?

function idempotent(arg1, arg2) {
  //
}

function run_idempotent(arg1, arg2) {
  if (not saved)
    saved = idempotent(arg1, arg2)
  return saved
}

if (arg1 changes or arg2 changes)
   saved = undefined

Then if this is super clear, we can point out that the really difficult thing is that not everything is an explicit argument to a function, and sometimes variables are stored off here and there. It's easy to forget they are in fact implicit arguments, and also easy to forget their dependencies in turn.

you say all this below, just wondering out loud if the pseudocode is helpful to explain that the core concept is actually very simple, it's the implementation in a complex code base without making really slow or unreadable, that is  hard.

Maybe not, just an idea.

book/invalidation.md
```diff
@@ -0,0 +1,379 @@
...
+our browser's graphics subsystem. But that hasn't helped
+layout-inducing animations, since those typically spend most of their
+time in layout, not graphics. Yet just like compositing enabled smooth
+animations by avoiding redundant raster work, we can avoid redundant
+layout work using a technique called invalidation.
+
+Idempotency
+===========
+
+Invalidation is a pretty simple idea: don't redo redundant layout
```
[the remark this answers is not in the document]
chrishtr: Or exercises. Regardless, I don't think we want to give the impression that this is special to layout.

chrishtr (approved):
I really like spending as much time as needed to carefully walk through the concepts. And structure the code very carefully.

They can learn for themselves how difficult it is to get it right by trying to implement more caches for other things like styles or paint.

book/invalidation.md
pavpanchekha: I added a discussion of invalidation vs dirty flags. I think it's a really helpful point and improves the flow.

I'm going to add a discussion of implicit arguments once we get to actual layout invalidation, because that's where we first see them. I've went back and rewrote the definition of idempotence to be "same output if called with the same inputs *and dependencies*", and then go back to that in sec 2.

I'm hesitant to do pseudo-code because that introduces a whole extra layer of mapping the pseudo-code to the actual code. Let me write more of the chapter and see if it would help or hurt. I did consider some kind of abstract `ProtectedField` class, where you explicitly list the inputs & outputs and it does the dirty flag stuff. That way the "pseudo-code" becomes actual code. But the layout code ended up looking kind of ugly and I either needed to define a class per field or use lambdas and other more Python-specific features, neither of which I was thrilled with. But happy to discuss!

book/invalidation.md
```diff
@@ -0,0 +1,379 @@
...
+prev: embeds
+next: skipped
+...
+
+So far, we've focused on improving animation performance by improving
+our browser's graphics subsystem. But that hasn't helped
+layout-inducing animations, since those typically spend most of their
+time in layout, not graphics. Yet just like compositing enabled smooth
+animations by avoiding redundant raster work, we can avoid redundant
+layout work using a technique called invalidation.
```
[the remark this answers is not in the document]
pavpanchekha: I guess you're right. I would like to give it a name, though. "Cache" makes things sound easy; I want it to sound a bit scary.

pavpanchekha:
I made several changes based on your suggestions. Since this chapter is still WIP I'm going to merge and continue writing, but happy to discuss more about it in meeting.

PR #691, opened 2022-12-21
Ban bots from indexing /draft/ and /samples/

pavpanchekha:
This doesn't really change anything, but it hopefully will cut down on the number of Googlebot indexing errors we see.

### DIFF
--- a/www/robots.txt
+++ b/www/robots.txt
@@ -1,0 +1,3 @@
+User-agent: *
+Disallow: /draft/
+Disallow: /examples/


www/robots.txt
```diff
@@ -0,0 +1,3 @@
+User-agent: *
+Disallow: /draft/
+Disallow: /examples/
```
chrishtr, line 3: Why disallow /examples?

pavpanchekha: Pretty funny story, but Google thinks some of them are "Soft 404s" because they contain just the word "Test" or something similar. I guess we could alternatively add more text to them to convince Google otherwise, but that seems counter to the actual purpose of the example.

PR #732, opened 2023-02-16
Pass over the first section on downloading & decoding images

pavpanchekha:
This PR is done, sort of, but it has two "TODO" blocks inside for changes I want @chrishtr's approval on.

At a high level, this PR makes the following changes:

*MIME types*: I moved `decode` on the body out of `request` and into its callers. This avoids the whole discussion of MIME types, since we now `decode` when we expect text and don't `decode` when we expect an image. I think this change is actually important to make, because as it stands, if you put an image URL into, say, a `<script src>`, you'll get a type error inside the browser! That said, we can fix that bug without necessarily removing MIME types; perhaps the caller to `request` can indicate the MIME type it wants, and then `request` can use that information to decide whether or not to decode internally. I'm calling this out explicitly because it's removing a _lot_ of content and I want to make sure it wasn't central.

*PIL/Skia*: I'm not sure why we're using PIL to open and resize images, where as far as I can tell Skia also has an `Image` class that can load encoded data with `MakeFromEncoded`. It might be that PIL is clearer about the decoding, resizing, and so on steps, and in case that's the reason I didn't change anything. (Because that content is clearly central to this section.) But it's worth thinking about whether Skia's `Image` could have the same discussion, at the very least because it's one less library to install.

*Resizing/Quality*: I split the discussion of resizing and resizing quality into two separate steps. I also fixed a bug regarding the `image-rendering` property; we were treating `crisp-edges` as Lanczos filtering, but it should be nearest neighbor according to MDN.

*Download method*: I added a `download_image` method to parallel `decode_image`, so it felt more clearly like this section had two major parts which resulted in two methods.

*Image layout*: I also added an extra test and in the process found a bug. We were decoding images to a height of `self.height`, but that isn't actually the image height—it's the larger of image or font height. This was leading to all images smaller than current text height showing up in the wrong place (`vertical-align: top` instead of `vertical-align: baseline`). This required some changes in the next section as well as multiple changes to the tests. I haven't done a pass over the next section, so if my code changes are kind of ugly, I can change it when passing over that section.

-----

So, before we merge, we need to make at least the following decisions:

- [x] Do we keep or drop MIME types? If we drop it, should we move it into a "Go Further" block or similar?
- [x] Do we keep PIL or switch to Skia? If we switch to Skia, we need to remove the corresponding "Go Further" block. (We have plenty of replacements.)
- [x] Is the image layout change fixing a bug, or did I misinterpret things?

### DIFF
--- a/book/embeds.md
+++ b/book/embeds.md
@@ -172,11 +172,11 @@ several algorithms for decoding, some of which result in a
 worse-looking image but are faster than the default.[^lossy] For
 example, just for resizing an image, there's fast, simple, "nearest
 neighbor" resizing and the slower but higher-quality "bilinear" or
-even "[Lanczos][lanczos]" resizing algorithm.
+even "[Lanczos][lanczos]" resizing algorithms.
 
 [lanczos]: https://en.wikipedia.org/wiki/Lanczos_resampling
 
-To give web page authors control over this performance bottleneck, can
+To give web page authors control over this performance bottleneck,
 there's an [`image-rendering`][image-rendering] CSS property that
 indicates which algorithm to use. Let's add that as an argument to
 `DrawImage`:
@@ -199,9 +199,9 @@ class DrawImage(DisplayItem):
         canvas.drawImageRect(self.image, self.rect, paint)
 ```
 
-So we've now got the helper functions we need to downloading and
-decoding images. But to actually put images into web pages, we're
-going to need to add images into our browser's layout tree.
+So we've now got the helper functions we need to download and decode
+images. But to actually put images into web pages, we're going to need
+to add images into our browser's layout tree.
 
 [^lossy]: Image formats like JPEG are [*lossy*][lossy], meaning that
     they don't faithfully represent all of the information in the
@@ -223,8 +223,8 @@ Interestingly, when we used Skia's `MakeFromEncoded`, we didn't need
 to pass in the image format. That's because many image formats start
 with ["magic bytes"][magic-bytes]; for example, PNG files always start
 with byte 137 followed by the letters "PNG". These magic bytes are
-often more reliable than server-send MIME types, so this kind of
-"format sniffing" is common inside browsers and their supporting
+often more reliable than web-server-provided MIME types, so this kind
+of "format sniffing" is common inside browsers and their supporting
 libraries.
 :::
 
@@ -252,10 +252,11 @@ it reaches the end of the available space.
 
 But it's different than a text *node*, because the text in a text node is not
 just one glyph, but an entire run of text of a potentially arbitrary length,
-and that can be split into words and lines across multiple lines. An image, on
-the other hand, is an [atomic inline][atomic-inline]---it doesn't make sense to
-split it across multiple lines.^[There are other elements that can be atomic
-inlines, and we'll encounter more later in this chapter.]
+and that can be split into multiple words and lines. An image, on the
+other hand, is an [atomic inline][atomic-inline]---it doesn't make
+sense to split it across multiple lines.^[There are other elements
+that can be atomic inlines, and we'll encounter more later in this
+chapter.]
 
 
 [atomic-inline]: https://drafts.csswg.org/css-display-3/#atomic-inline
--- a/book/embeds.md
+++ b/book/embeds.md
@@ -46,25 +46,7 @@ started. There are four steps to displaying images in our browser:
 3. Lay the image out on the page.
 4. Paint the image in the display list.
 
-You probably know that there are lots of image formats (like JPG, PNG,
-and many more obscure ones), so for the decoding step^[I'll get into
-how this works in a bit; for now just think of decoding as something
-like decompressing a zip file.] we'll use the [Pillow/PIL][pillow]
-library. You can install it like so:
-
-[pillow]: https://pillow.readthedocs.io/en/stable/reference/Image.html
-
-    pip3 install Pillow
-
-Import it to make sure the installation worked:^[Pillow is a fork of a
-project called PIL---for Python Image Library---which is why the
-import says PIL.]
-
-``` {.python}
-import PIL.Image
-```
-
-Now let's start with downloading images from a URL. Naturally, that
+Let's start with downloading images from a URL. Naturally, that
 happens over HTTP, which we already have a `request` function for.
 However, all of the content we've downloaded so far---HTML, CSS, and
 JavaScript---has been textual, but images typically use binary data
@@ -96,7 +78,7 @@ def request(url, top_level_url, payload=None):
 ```
 
 Note that I _didn't_ add a `decode` call when we read the body; that's
-because the body actually might be binary data, and we want to return
+because the body might actually be binary data, and we want to return
 that binary data directly to the browser. Now, every existing call to
 `request`, which wants textual data, needs to `decode` the response.
 For example, in `load`, you'll want to do something like this:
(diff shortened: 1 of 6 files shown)


book/embeds.md
```diff
@@ -102,26 +91,75 @@ def request(url, top_level_url, payload=None):
...
+    except:
+        print("Failed to load image: " + image_url)
+```
+
+This method wraps the image bytes as a `BytesIO` object, and then
+passes that to `open` to turn it into an Pillow `Image` object.
+
+::: {.todo}
+
+Chris, do we actually need MIME types? It seems kind of dangerous the
```
chrishtr, line 136: I guess we don't need it after your proposed refactoring.

chrishtr: We could remove it or move it to a go-further block.

book/embeds.md
```diff
@@ -145,104 +183,133 @@ def request(url, top_level_url, payload=None):
...
 
-``` {.python expected=False}
-def decode_image(image_bytes):
-    picture_stream = io.BytesIO(image_bytes)
-    pil_image = PIL.Image.open(picture_stream)
-    if pil_image.mode == "RGBA":
-        pil_image_bytes = pil_image.tobytes()
-    else:
-        pil_image_bytes = pil_image.convert("RGBA").tobytes()
+Before we can use a downloaded image, we need to _decode_ it. That's
```
chrishtr, line 193:
```suggestion
After we can use a downloaded image, we need to _decode_ it. That's
```

book/embeds.md
```diff
@@ -145,104 +183,133 @@ def request(url, top_level_url, payload=None):
...
+_when_ Pillow actually performs the decoding: eagerly when we `open`
+the file or lazily when we call `tobytes`. This may seem like an
+obscure issue, but in a real browser the timing of a decode is
+important for performance. That's also why there is an [HTML
+API][html-image-decode] to control decoding.
+
+[html-image-decode]: https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/decoding
+
+::: {.todo}
+I'm a little confused why we use PIL at all, seeing as Skia seems to
```
chrishtr, line 229: I didn't know it had this support. If it works then we can get rid of PIL.

book/embeds.md
```diff
@@ -145,104 +183,133 @@ def request(url, top_level_url, payload=None):
...
 
+[make-from-encoded]: https://kyamagu.github.io/skia-python/reference/skia.Image.html#skia.Image.MakeFromEncoded
+:::
+
+Because image formats use many clever and sophisticated algorithms,
+image decoding is slow and consumes a lot of memory.^[More generally,
+images are more expensive than text in a lot of ways. They take a long
+time to download; decoding is slow; resizing is slow; and they can
+take up a lot of both CPU and GPU memory. Optimizing image handling is
+essential to a performance browser.] Decoding might involve resolving
```
chrishtr, line 249:
```suggestion
essential to a performant browser.] Decoding might involve resolving
```

chrishtr:
I just landed a PR to get rid of PIL images. I made only the minimum changes to the chapter text though, hopefully it should be easy for you to rebase your PR.

book/embeds.md
pavpanchekha: I'll make a go-further block out of it.

book/embeds.md
pavpanchekha: Don't think the proposal is correct but will rephrase.

pavpanchekha:
Ok, I think the text rewrite will be a little subtle, because there's no step where decoding "happens" now. What I might do is move painting up from the next section somehow, so we can talk about it in this section and leave the next section for life cycle + layout.

pavpanchekha:
This is rewritten and ready for re-review.

book/embeds.md
```diff
@@ -88,147 +73,193 @@ def request(url, top_level_url, payload=None):
...
 ```
 
-Now define a method that decodes a response body that we know is an image
-(even if we don't know its format).^[Interestingly, to make it work for our toy
-browser we don't need to consult `content-type`. That's because Skia already
-auto-detects the image format by peeking at the first few bytes of the binary
-data, which varies for each image format.][^maybe-decode]
+Make sure to make this change everywhere in your browser that you call
+`request`, including inside `XMLHttpRequest_send` and in several other
+places in `request`. When we download images, however, we _won't_ call
```
chrishtr, line 97:
```suggestion
places in `load`. When we download images, however, we _won't_ call
```

book/embeds.md
```diff
@@ -88,147 +73,193 @@ def request(url, top_level_url, payload=None):
...
+
+`MakeWithoutCopy` means that the `Data` object just stores a reference
+to the existing `body` and doesn't own that data. That's essential,
+because encoded image data can be large---maybe megabytes---and
+copying that data wastes memory and time. But that means that the
+`data` is invalid if `body` is ever garbage-collected, so we return
+the `body` from `download_image` and need to make sure to store it
+somewhere for at least as long as we're using the image.[^memoryview]
+
+[^memoryview]: This is a bit of a hack. Perhaps a better solution
```
chrishtr, line 125: I don't quite understand why it's a hack, can you explain?

book/embeds.md
```diff
@@ -88,147 +73,193 @@ def request(url, top_level_url, payload=None):
...
-because many encoded image formats are [*lossy*][lossy], meaning that
-they "cheat": they don't faithfully represent all of the information in the
-original picture, in cases where it's unlikely that a human viewing the decoded
-image will notice the difference.
+But because image decoding can be so expensive, Skia actually has
+several algorithms for decoding, some of which result in a
+worse-looking image but are faster than the default.[^lossy] For
+example, just for resizing an image, there's fast, simple, "nearest
+neighbor" resizing and the slower but higher-quality "bilinear" or
+even "[Lanczos][lanczos]" resizing algorithm.
```
chrishtr, line 175:
```suggestion
even "[Lanczos][lanczos]" resizing algorithms.
```

book/embeds.md
```diff
@@ -88,147 +73,193 @@ def request(url, top_level_url, payload=None):
...
-    elif attr == "crisp-edges":
-        return skia.FilterQuality.kLow_FilterQuality
-    else:
-        return skia.FilterQuality.kMedium_FilterQuality
-```
-
-Now, of course, the `width` and `height` have to be specified to the
-"painted" size on the screen. Which means it's now time to figure out image
-layout and painted sizing.
+To give web page authors control over this performance bottleneck, can
```
chrishtr, line 179:
```suggestion
To give web page authors control over this performance bottleneck,
```

book/embeds.md
```diff
@@ -88,147 +73,193 @@ def request(url, top_level_url, payload=None):
...
+            self.quality = skia.FilterQuality.kLow_FilterQuality
+        else:
+            self.quality = skia.FilterQuality.kMedium_FilterQuality
+
+    def execute(self, canvas):
+        paint = skia.Paint(FilterQuality=self.quality)
+        canvas.drawImageRect(self.image, self.rect, paint)
+```
+
+So we've now got the helper functions we need to downloading and
```
chrishtr, line 202:
```suggestion
So we've now got the helper functions we need to download and
```

book/embeds.md
```diff
@@ -88,147 +73,193 @@ def request(url, top_level_url, payload=None):
...
+decoding images. But to actually put images into web pages, we're
```
chrishtr, line 203:
```suggestion
decode images. But to actually put images into web pages, we're
```

book/embeds.md
```diff
@@ -88,147 +73,193 @@ def request(url, top_level_url, payload=None):
...
+whether a document contains text or binary data. The header contains a
+value called a [MIME type][mime-type]: `text/html`, `text/css`, and
+`text/javascript` for HTML, CSS, and JavaScript; `image/png` and
+`image/jpeg` for PNG and JPEG images; and [many others][mime-list] for
+different font, video, audio, and data formats.[^mime-history]
+Interestingly, when we used Skia's `MakeFromEncoded`, we didn't need
+to pass in the image format. That's because many image formats start
+with ["magic bytes"][magic-bytes]; for example, PNG files always start
+with byte 137 followed by the letters "PNG". These magic bytes are
+often more reliable than server-send MIME types, so this kind of
```
chrishtr, line 226:
```suggestion
often more reliable than web-server-provided MIME types, so this kind of
```

book/embeds.md
```diff
@@ -88,147 +73,193 @@ def request(url, top_level_url, payload=None):
...
 ===============
 
+An `<img>` is a leaf element of the DOM. In some ways, it's similar to a single
+font glyph that has to paint in a single rectangle sized to the image (instead
+of the glyph), takes up space in a `LineLayout`, and causes line breaking when
+it reaches the end of the available space.
+
+But it's different than a text *node*, because the text in a text node is not
+just one glyph, but an entire run of text of a potentially arbitrary length,
+and that can be split into words and lines across multiple lines. An image, on
```
chrishtr, line 255:
```suggestion
and that can be split into multiple words and lines. An image, on
```

book/embeds.md
pavpanchekha: Uhh, well, IDK, I think a "nice" solution wouldn't involve working around garbage collector weaknesses. In C++ I would "move" the request object into the Skia Data, and or writing directly to a Data would be nicer. But perhaps what you really mean is—let's explain the footnote a bit better so people aren't confused. Will do.

chrishtr: ok sounds good

pavpanchekha:
Done with all the suggestions. Only significant change is rewriting the footnote labeled `memoryview`.

book/embeds.md
chrishtr: What about this proposed change?

pavpanchekha:
Woops

PR #735, opened 2023-02-23
Pass over second section on image layout

pavpanchekha:
This PR contains a pass over the second section on image layout. It's probably worth reading, but there are some TODO items below that need discussion:

- [x] I moved loading out of `ImageLayout` and into the main `Frame.load` method. This make the already-long `load` method even longer, which I'm not thrilled with, but at least it parallels existing code and gets all the loading done on initial page load. It also didn't make sense to have a separate `download_image` method at this point.
- [x] I also described `EmbedLayout` a bit different, as if it's a renamed `InputLayout` that's been hollowed out. I'm not sure if this is exactly right, but we should try to make it right.
- [x] I split the `width` and `height` attributes to their own section. I'll maybe merge it back after I pass on that.

Still needs discussion:

- [x] We need to update Chapter 15 to use the unified `BlockLayout`
- [x] We also don't have good behavior when an image fails to load. At the very least, we need to make that not crash
- [x] I'm not thrilled that we have parallel `image`, `input`,`text`, and `iframe` methods in `InlineLayout` soon to be `BlockLayout`—could reinforce the abstraction layer here.
- [-] It would be nice to clean up `load`, probably in the `Task` chapter.

### DIFF
--- a/book/embeds.md
+++ b/book/embeds.md
@@ -46,25 +46,7 @@ started. There are four steps to displaying images in our browser:
 3. Lay the image out on the page.
 4. Paint the image in the display list.
 
-You probably know that there are lots of image formats (like JPG, PNG,
-and many more obscure ones), so for the decoding step^[I'll get into
-how this works in a bit; for now just think of decoding as something
-like decompressing a zip file.] we'll use the [Pillow/PIL][pillow]
-library. You can install it like so:
-
-[pillow]: https://pillow.readthedocs.io/en/stable/reference/Image.html
-
-    pip3 install Pillow
-
-Import it to make sure the installation worked:^[Pillow is a fork of a
-project called PIL---for Python Image Library---which is why the
-import says PIL.]
-
-``` {.python}
-import PIL.Image
-```
-
-Now let's start with downloading images from a URL. Naturally, that
+Let's start with downloading images from a URL. Naturally, that
 happens over HTTP, which we already have a `request` function for.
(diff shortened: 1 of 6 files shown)


book/embeds.md
```diff
@@ -88,198 +73,226 @@ def request(url, top_level_url, payload=None):
...
-        self.width = None
-        self.height = None
-        self.font = None
-
-    def get_ascent(self, font_multiplier=1.0):
-        return -self.height
-
-    def get_descent(self, font_multiplier=1.0):
-        return 0
+[^atomic-inline]: Images aren't quite like *text* because text node is
```
chrishtr, line 280:
```suggestion
[^atomic-inline]: Images aren't quite like *text* because a text node is
```

pavpanchekha:
I'll take the first todo on BlockLayout

pavpanchekha:
@chrishtr will do task 2 with a placeholder image

pavpanchekha:
Let's not do Task 4

pavpanchekha:
For Task 3 on unifying `input`, `text`, `iframes`, `inputs`, @chrishtr will do that. The goal is to unify all of them (maybe with an extra `for` loop for `text`) and put the unification text in Chapter 15 (with maybe a forward reference in Chapter 8).

pavpanchekha:
This PR has been split into #752 and #751; closing this one as it's been obsoleted.

PR #736, opened 2023-02-24
Add alt text to images in the book

chrishtr:


book/history.md
```diff
@@ -21,7 +21,7 @@ The Memex concept
 =================
 
 <figure>
-  <img src="im/memex.jpg">
+  <img src="im/memex.jpg" alt="A photograph of the first few lines of *As You May Think*">
```
pavpanchekha, line 24: I think the stars will go into the HTML. I don't think we want that.

pavpanchekha (approved):

PR #737, opened 2023-02-24
Fix #734

### LINKED ISSUE #734: We don't support our own iframe
If you use the Chapter 15 browser to load `https://browser.engineering/`, it crashes when trying to render a button that has HTML inside it:

```
 <button tabindex="0" class="butto" rightbutton="" primary="" subscribe-btn"="" type="submit">
   <b class="button-tex" "="">
     'Subscribe'
```

That's because our browser assumes a button only has text inside. We don't need to solve this the "right" way (that's an exercise) but we should at least make it not crash!

pavpanchekha:
As discussed in meeting today, buttons with rich markup should just not print their contents. I added this change to the book, every chapter where it comes up, and added a test for it too.

### DIFF
--- a/book/forms.md
+++ b/book/forms.md
@@ -165,7 +165,7 @@ Note that `<button>` elements can in principle contain complex HTML,
 not just a text node. I'm having the browser print a warning and skip
 the text in that case.[^exercises]
 
-[^exercises]: There's an [exercise](#exercises) on this
+[^exercises]: There's an [exercise](#exercises) on this.
 
 Finally, we draw that text:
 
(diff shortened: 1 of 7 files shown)


book/forms.md
```diff
@@ -153,8 +153,25 @@ class InputLayout:
...
+            else:
+                print("Ignoring HTML contents inside button")
+                text = ""
+```
+
+Note that `<button>` elements can in principle contain complex HTML,
+not just a text node. I'm having the browser print a warning and skip
+the text in that case.[^exercises]
+
+[^exercises]: There's an [exercise](#exercises) on this
```
chrishtr, line 168:
```suggestion
[^exercises]: There's an [exercise](#exercises) on this.
```

PR #739, opened 2023-02-26
Refactor child methods of InlineLayout to reduce repeated code

chrishtr:
...and emphasize to the reader how much they have in common.

I made the minimum edits to the chapter to make linting pass and remove/update obviously out of date commentary. 

I also moved image loading to `load`, because it was necessary to complete the refactoring.

And I found and fixed a bug where keypress after focus tabbing didn't activate an input element, and so
text entry didn't work. (unlike focusing it via click ). This bug is present in chapter 14 also, and I fixed it in both chapters.

### DIFF
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -941,7 +941,7 @@ class Tab:
 
 All of this activation code is copied from the `click` method on
 `Tab`s, which can now be rewritten to call `activate_element`
-directly:
+directly. Code is also needed in `keypress` to activate if needed:
 
 ``` {.python}
 class Tab:
@@ -955,6 +955,11 @@ class Tab:
                 self.set_needs_render()
                 return
             elt = elt.parent
+
+    def keypress(self, char):
+        if self.focus and self.focus.tag == "input":
+            if not "value" in self.focus.attributes:
+                self.activate_element(self.focus)
 ```
 
 Note that hitting `Enter` when focused on a text entry clears the text
--- a/book/embeds.md
+++ b/book/embeds.md
@@ -367,12 +367,15 @@ varies depending on the child type.
 
 ``` {.python}
 class BlockLayout(LayoutObject):
-    def add_inline_child(self, node, zoom, w, child_class, extra_param):
+    def add_inline_child(self, node, zoom, w, child_class, frame, word=None):
         font = self.font(node, zoom)
         if self.cursor_x + w > self.x + self.width:
             self.new_line()
         line = self.children[-1]
-        child = child_class(node, line, self.previous_word, extra_param)
+        if word:
+            child = child_class(node, line, self.previous_word, word)
+        else:
+            child = child_class(node, line, self.previous_word, frame)
         line.children.append(child)
         self.previous_word = child
         self.cursor_x += w + font.measureText(" ")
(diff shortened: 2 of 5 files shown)


chrishtr:
I have rebased this PR on top of yours that landed just now. Ready for review.

pavpanchekha (commented):
Definitely we should merge this or something like it. I like the idea of computing the width in the `input`/`text`/etc helper methods to avoid a two-phase layout.

Some questions, you probably considered them, but I do want your thoughts:

- Why not pass the `tab` to `TextLayout`, so that these classes are more uniform?
- Why not compute the `font` inside `add_inline_child` instead of passing it through as an argument (we'd end up computing the font once per word instead of once for the whole text node, but that should hit the cache)
- Is there any way we can avoid passing the `word` into `TextLayout`? All the options I can think of are kind of ugly, but the way `extra_param` can mean different things in different call sites is also ugly...

chrishtr:
Answers below. I don't have strong opinions.

> * Why not pass the `tab` to `TextLayout`, so that these classes are more uniform?

Because it would result in two parameters, and it's not needed.

> * Why not compute the `font` inside `add_inline_child` instead of passing it through as an argument (we'd end up computing the font once per word instead of once for the whole text node, but that should hit the cache)

For performance, but you're right that it should hit the cache. Fixed.

> * Is there any way we can avoid passing the `word` into `TextLayout`? All the options I can think of are kind of ugly, but the way `extra_param` can mean different things in different call sites is also ugly...

I tried something. Better now?

book/accessibility.md
```diff
@@ -955,6 +955,11 @@ class Tab:
                 self.set_needs_render()
                 return
             elt = elt.parent
+
+    def keypress(self, char):
+        if self.focus and self.focus.tag == "input":
+            if not "value" in self.focus.attributes:
+                self.activate_element(self.focus)
```
pavpanchekha, line 962: Just so I understand the bug—is it that tabbing into something and then typing into it would crash (because the `value` attribute didn't exist?).

book/embeds.md
```diff
@@ -333,71 +334,93 @@ class InputLayout(EmbedLayout):
...
+        if isinstance(node, Text):
+            self.text(node, zoom)
+        else:
+            if node.tag == "br":
+                self.new_line()
+            elif node.tag == "input" or node.tag == "button":
+                self.input(node, zoom)
+            else:
+                for child in node.children:
+                    self.recurse(child, zoom)
```
pavpanchekha, line 361: Now, this part doesn't change, right? We're just adding a `font` method?

Also, IIUC the `font` method doesn't rely on the `BlockLayout` at all. It's fine to keep it here, but what do you think of moving it to a stand-alone method or a method on `Element` or `Text` or something? That can be a different PR.

book/embeds.md
```diff
@@ -333,71 +334,93 @@ class InputLayout(EmbedLayout):
...
-        line.children.append(img)
-        self.previous_word = img
-        weight = node.style["font-weight"]
-        style = node.style["font-style"]
-        size = device_px(float(node.style["font-size"][:-2]), zoom)
-        font = get_font(size, weight, size)
+        if word:
+            child = child_class(node, line, self.previous_word, word)
+        else:
+            child = child_class(node, line, self.previous_word, frame)
```
pavpanchekha, line 378: Ugh, this isn't much better—we still have nearly-parallel interfaces that aren't quite parallel. I think neither of us have a great idea for what to do here, and the overall PR is still a net improvement, so we should merge. But this is definitely kind of ugly.

pavpanchekha (approved):

book/accessibility.md
chrishtr: Yes, exactly.

book/embeds.md
chrishtr: good point. I'll delete the recurse part in this PR at least.

book/embeds.md
chrishtr: ack

PR #740, opened 2023-02-27
Unify Block- & InlineLayout in Chapter 15

pavpanchekha:
This PR unifies `BlockLayout` and `InlineLayout` in `lab15.py` without making any other changes.

It's best to read only the second commit in the PR because it has a much cleaner diff. The first commit just moves `InlineLayout` and `BlockLayout` to be next to each other in the source.


pavpanchekha:
I'll add that I'll handle whatever merge conflicts come from this and the `refactorinline` branch.

PR #742, opened 2023-02-28
Feedback pass

pavpanchekha:
These changes were suggested by readers via WBE's feedback mechanism. Thank you, readers!


pavpanchekha:
I especially want to thank Julian Geppert, who pointed out that we had too big a click area for the address bar, across 8+ chapters and even relied on in two chapters' tests.

chrishtr (approved):
That's an impressive number of "the the" errors!

PR #745, opened 2023-03-03
Refactor font code to a static method

chrishtr:
*  I also removed some incorrect code for getting fonts back to chapter 11, which replaced tkinter with Skia, and also changed `get_font`
* I didn't add a method on `Element` or `Text` because it would have been code duplication on the two classes, and there is no `Node` class defined.

### DIFF
--- a/book/embeds.md
+++ b/book/embeds.md
@@ -306,13 +306,7 @@ class EmbedLayout(LayoutObject):
         return 0
 
     def layout(self, zoom):
-        weight = self.node.style["font-weight"]
-        style = self.node.style["font-style"]
-        if style == "normal": style = "roman"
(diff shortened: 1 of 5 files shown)


book/embeds.md
```diff
@@ -366,7 +358,7 @@ class BlockLayout(LayoutObject):
             child = child_class(node, line, self.previous_word, frame)
         line.children.append(child)
         self.previous_word = child
-        self.cursor_x += w + font.measureText(" ")
+        self.cursor_x += w + font(node, zoom).measureText(" ")
```
pavpanchekha, line 361: This is really nice

pavpanchekha (approved):

PR #751, opened 2023-03-06
Align image downloading with `image-layout`

pavpanchekha:
I'm converting #735 into two PRs. This is one of them: it redoes how `download_image` works, merging it into `Frame.load`. Along the way, I incorporate the placeholder changes. I also fixed some minor bugs:

- If the image fails to decode, we now show the placeholder instead of crashing
- If the image has an empty `src` attribute, we no longer try to load an image


src/lab15.py
```diff
@@ -1234,25 +1234,23 @@ def load(self, url, body=None):
...
 
         images = [node
-                   for node in tree_to_list(self.nodes, [])
-                   if isinstance(node, Element)
-                   and node.tag == "img"
-                   and "src" in node.attributes]
+            for node in tree_to_list(self.nodes, [])
+            if isinstance(node, Element)
+            and node.tag == "img"
+            and node.attributes.get("src")]
```
chrishtr, line 1240: You should show the broken image icon even if the `src` attribute is omitted

book/embeds.md
```diff
@@ -95,54 +95,75 @@ class Tab:
...
+            and node.attributes.get("src")]
+        for img in images:
+            image_url = resolve_url(img.attributes["src"], self.url)
+            assert self.allowed_request(image_url), \
+                "Blocked load of " + image_url + " due to CSP"
+            header, body = request(image_url, self.url)
+```
 
-``` {.python replace=tab/frame}
-def download_image(image_src, tab):
```
chrishtr, line 104: Also delete the code

book/embeds.md
```diff
@@ -95,54 +95,75 @@ class Tab:
...
+    def load(self, url, body=None):
+        for img in images:
+            try:
+                # ...
+            except Exception as e:
+                print(e)
+                img.image = skia.Image.open("Broken_Image.png")
+```
+
+Note that when errors don't occur, created the `Image` object using
```
chrishtr, line 160:
```suggestion
Note that when errors don't occur, we created the `Image` object using
```

src/lab15.py
pavpanchekha: I did some tests in a Chrome variant and it looked like you needed a nonempty src. Just to be sure, I just looked it up in the spec:

https://html.spec.whatwg.org/multipage/embedded-content.html#the-img-element

Way down in this section it tells what the image presents as:

> If the [src](https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-src) attribute is not set and either the [alt](https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-alt) attribute is set to the empty string or the [alt](https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-alt) attribute is not set at all
> The element [represents](https://html.spec.whatwg.org/multipage/dom.html#represents) nothing.

> Otherwise
> The element [represents](https://html.spec.whatwg.org/multipage/dom.html#represents) the text given by the [alt](https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-alt) attribute.

which points out an interesting thing, that an Img element without src but with alt is actually a kind of text node… I don’t feel like implementing this but it could be a footnote.

another thing not totally clear in the spec is what it means by a the src attribute not being set. In my tests, an empty, set src attribute still doesn’t trigger fallback behavior, and text earlier in the spec suggests you should strip whitespace, but it’s not obvious to me. I haven’t read this part of the spec in full so I might be missing something.

chrishtr: The part of the spec you mentioned is in an exercise already. But `data:text/html,` does show the broken image in Chrome.

pavpanchekha: Ok, this definitely isn't too important but we should understand for ourselves. Here's a series of tests I did, all with a doctype so we're not in quirks mode:

[ ] ``
[ ] ``
[ ] ``
[x] ``
[ ] ``
[ ] ``
[ ] ``
[ ] ``
[x] ``
[x] ``

pavpanchekha: The weirdest one is ``, which seems like it should show a placeholder but doesn't. I can't explain that. If you ignore that, then a nonempty `src` or a nonempty `alt` is enough, but perhaps we should leave the non-empty `alt` case as an exercise?

chrishtr: I think it's because a bare `alt` isn't nonempty.

pavpanchekha: Ok, let's discuss this tomorrow.

pavpanchekha:
Ok, I made it do `"src" in node.attributes` again. Along the way I also put that in the conditional in `BlockLayout` so we won't make `ImageLayout`s for `img` elements without `src` attributes. I also did the same for `iframe`s. This used to cause a crash.

src/lab15.py
```diff
@@ -1234,25 +1222,23 @@ def load(self, url, body=None):
...
 
         images = [node
-                   for node in tree_to_list(self.nodes, [])
-                   if isinstance(node, Element)
-                   and node.tag == "img"
-                   and "src" in node.attributes]
+            for node in tree_to_list(self.nodes, [])
+            if isinstance(node, Element)
+            and node.tag == "img"
+            and "src" in node.attributes]
```
chrishtr, line 1228: I thought we decided to always show the broken image icon, even if `src` was specified?

pavpanchekha: Oh—my bad. Fine, always show the `src`. When you modify the exercise, can you have me review that PR?

chrishtr: ok

pavpanchekha:
Ok, @chrishtr, I'm a little lost on what the status of this PR is, but it should now pass tests and lints again, and every place where we handle images we just test `node.tag == "img"`. I believe that's what we decided? If so, this should be ready to merge if you approve it.

PR #756, opened 2023-03-08
Pass on iframe construction & rendering chapters

pavpanchekha:
This PR doesn't have too many code changes or very significant text changes---I broke up and reordered a few code blocks and shortened the text, but that's about it.

Some questions for discussion:

* Does CSP block iframe loading? If so, we need to add code to do that.
* Why increment the window ID in the `Frame` constructor? I'd rather do it when the `Frame` is constructed. Any objections?
* I added a sentence on the order in which you need to lay out frames. Can you confirm my logic?

All of these should be easy fixes.


book/embeds.md
```diff
@@ -984,7 +1014,13 @@ class IframeLayout(EmbedLayout):
         self.node.frame.frame_width = self.width - 2
 ```
 
-The root frame is sized to the window:
+Note that there's a tricky dependency order here. We need the parent
+frame to do layout before the child frame, so the child frame has an
+up to date width and height when it does layout. That order is
```
chrishtr, line 1019:
```suggestion
up-to-date width and height when it does layout. That order is
```

book/embeds.md
```diff
@@ -1018,19 +1068,49 @@ class IframeLayout(EmbedLayout):
...
 
-        cmds = [Transform(
-            (self.x + 1 , self.y + 1), rect, self.node, cmds)]
+Note the last line, where we recursively paint the child frame.
 
-        paint_outline(self.node, cmds, rect)
+Before putting those commands in the display list, though, we need to
+add a border and transform the coordinate system:
 
+[box-model]: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Introduction_to_the_CSS_box_model
```
chrishtr, line 1081: Nit: move this below line 1103.

book/embeds.md
```diff
@@ -984,7 +1014,13 @@ class IframeLayout(EmbedLayout):
         self.node.frame.frame_width = self.width - 2
 ```
 
-The root frame is sized to the window:
+Note that there's a tricky dependency order here. We need the parent
+frame to do layout before the child frame, so the child frame has an
+up to date width and height when it does layout. That order is
+guaranteed for us by Python (where dictionaries are sorted by
```
chrishtr, line 1020:
```suggestion
guaranteed for us by Python 3.7 or later (where dictionaries are sorted by
```

chrishtr: (read this on stack overflow)

chrishtr:
> Does CSP block iframe loading? If so, we need to add code to do that.

It seems yes: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-src

> Why increment the window ID in the Frame constructor? I'd rather do it when the Frame is constructed. Any objections?

What does "when the frame is constructed" mean, and how does that differ from the constructor?

> I added a sentence on the order in which you need to lay out frames. Can you confirm my logic?

Yes, that is correct. However, it seems I got lucky, as I didn't think about this detail..

book/embeds.md
```diff
@@ -652,71 +652,82 @@ media controls---requires. In modern browsers, these are handled by
...
-or DOM in the containing page, nor can scripts in the containing page access
-the iframe's variables or DOM.
-
-[^iframe-event-loop]: For example, if an iframe has the same origin as the web
-page that embeds it, then scripts in the iframe can synchronously access the
-parent DOM. That means that it'd be basically impossible to put that iframe in
-a different thread or CPU process, and in practice it ends up in the same
-rendering event loop as a result.
+* Iframes can *share a rendering event loop*.[^iframe-event-loop] In
+  real browsers, [cross-origin] iframes are often "site isolated",
```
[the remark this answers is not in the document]
pavpanchekha: Why capitalize this?

pavpanchekha:
Ok, I added CSP blocking and fixed the nits. Regarding window IDs, I meant computing and updating the window ID in `load`. It does mean duplicating the window ID, so maybe you'd rather not.

book/embeds.md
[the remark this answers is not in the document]
chrishtr: oh oops. I thought it was the start of a sentence.

book/embeds.md
```diff
@@ -595,71 +595,82 @@ media controls---requires. In modern browsers, these are handled by
...
 [cant-access]: https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy#cross-origin_script_api_access
 
-Since iframes are HTML documents, they can contain iframes. So in general each
-`Tab` has a tree of objects---*frames* containing HTML documents---nested
-within each other. Each node in this tree will be an object from a new `Frame`
-class. We'll use one rendering event loop for all `Frame`s.
+We'll get to these differences, but for now, let's start working on
+the idea of a `Tab` within a `Tab`. What we're going to do is split
+the concept of a `Tab` into two pieces: a `Tab` will own the event
+loop and script environments, and will contain a tree---yes, a tree,
```
chrishtr, line 638:
```suggestion
loop and script environments, and will contain a tree
```

book/embeds.md
```diff
@@ -595,71 +595,82 @@ media controls---requires. In modern browsers, these are handled by
...
+because an iframe can itself contain nested iframes---of `Frame`s.
```
chrishtr, line 639:
```suggestion
of `Frame`s.
```

chrishtr: (this is redundant, already mentioned earlier)

chrishtr (approved):

chrishtr:
Re window id in load: if it makes the code better I don't oppose it, send as a separate PR?

pavpanchekha:
Agreed, let's merge this and do a separate PR.

PR #759, opened 2023-03-09
Fix attribute parser in Chapter 15

pavpanchekha:
I assume this is here for testing? Or left over? Perhaps we should just delete it...


chrishtr:
I referred to it in my chapter text, saying it was a lot like the CSS parser but left it out because it was redundant. I added it because alt text is unreasonably restrictive otherwise, as it can't contain a space character.

chrishtr (approved):

Issue #758, opened 2023-03-09
Clicking on links in Chapter 15 draft doesn't work


pavpanchekha:
Clicking on links was broken by the `dispatch` change. The core reason is thus:

- When you click on something, we resolve it to a /layout/ object
- You then walk up and dispatch based on the /element/ tree
- This is important because an `<a>` tag does not have a layout object in our browser

My best suggestion is that we undo the `dispatch` changes, which is also shorter.

PR #760, opened 2023-03-09
Pass on most of iframe events

pavpanchekha:
This PR does a pass on most off the iframe events section, except the part on dispatching clicks because (see #758) I think that has to be redone, and I want this PR to be independently mergable.

I fixed one minor bug (switching focus between two elements in different frames didn't redraw the blurred frame) and also redid how the `AccessibilityTree` stuff worked. Instead of making `AccessibilityTree` a sister class to `AccessibilityNode`, I made it a subclass (called `FrameAccessibilityTree`), which also let me keep `tree_to_list`.


pavpanchekha:
Let me add that I'd like to move the initial build of the accessibility tree into the earlier rendering section, but I'm saving it until its PR ends up merged to avoid conflicts.

chrishtr (commented):
I assume you tested the accessibility code with frames?

pavpanchekha:
Uhh no, I need to test before we merge, you're right.

pavpanchekha:
Ok, it's good you reminded me to test.

1. I found and fixed some random unrelated issues, by accident
2. I also fixed how the frame accessibility nodes work—they were wrong
3. Finally, I found what I think was an *existing* bug where hovered nodes inside a child frame are drawn in the wrong place. I think we should merge this PR and then work on that.

pavpanchekha:
@chrishtr I think it would be ideal if you could push commits onto this PR (or merge it and put in a new PR) for the a11y hover issue. Basically, the `hit_test` method currently returns an `AccessibilityNode`, which the browser then gets the `bounds` field from to draw the hover outline. This `bounds` field is computed from the underlying layout object, which means it's in the child frame's coordinate system, not the parent frame's coordinate system. I think it also doesn't respond correctly to scrolling. (Again, I believe all these bugs pre-exist this PR.) Here's my thoughts on possible fixes:

- We can do a post-pass to adjust the bounds in `FrameAccessibilityNode.build`. I implemented this and it works, but it's ugly. This can't handle threaded scrolling on child frames, which I guess we don't support but still suggests it's the wrong solution.
- We can add a `transform` method to `FrameAccessibilityNode`s, similar to what we have on the display tree. It's duplicative, but not the end of the world, and would work correctly with threaded scrolling.
- Longer term, I think we need a `CoordinateSystem` abstraction, but we can discuss that in meeting; we won't do that in the next few days.

book/embeds.md
```diff
@@ -1088,14 +1091,8 @@ class Tab:
...
-Rendering now functions properly in iframes, but user input does not:
-it's not (yet) possible to click on a element in an iframe in our toy browser,
-iterate through its focusable elements, scroll it, or generate an accessibility
-tree.
-
-Let's fix that. It's as simple as checking for an iframe, and if found
-sending the click event to it and immediately returning after (because
-iframes capture click events):
+When an iframe is clicked, pass the click through to the child frame,
+and immediately returning after (because iframes capture click events):
```
chrishtr, line 1095:
```suggestion
and immediately return after (because iframes capture click events):
```

book/embeds.md
```diff
@@ -1127,99 +1125,81 @@ class Tab:
...
-                self.document.height) - self.frame_height))
+class Tab:
+    def advance_tab(self):
+        frame = self.focused_frame or self.root_frame
+        frame.advance_tab()
 ```
 
-Now change all call sites of `clamp_scroll` to use the method rather than the
-global function:
+We'll do the same exact thing for `keypress` and `enter`.
```
chrishtr, line 1163:
```suggestion
Also do the same exact thing for `keypress` and `enter`.
```

book/embeds.md
```diff
@@ -1127,99 +1125,81 @@ class Tab:
...
+    def __init__(self, tab, parent_frame, frame_element):
+        self.scroll = 0
 ```
 
-Our browser supports browser-thread scrolling, but only for the root frame.
-To handle both cases, we'll need a new commit parameter:
+Now, as you might recall from [Chapter 13](animations.md), scrolling
+happens both inside `Browser` and inside `Tab`, to reduce latency.
+That was already quite complicated, so to keep things simple, we won't
+do this for non-root iframes. To support this, we'll need a new commit
```
chrishtr, line 1177:
```suggestion
support both for non-root iframes. For root frames, we'll need a new commit
```

PR #761, opened 2023-03-11
Pass on iframe scripts

pavpanchekha:
This PR does a pass on iframe scripting. It includes two major changes to the code:
- `JSContext`s are now stored on the `Tab` and requested from it by the `Frame` on load, as we discussed today
- The `with` hack is gone, replaced with admonitions to prefix everything with `window`. The nice thing about this is that when it fails, it's an error, so I think this is a big improvement.

There are some more minor code changes. I reorganized the RAF code into a `dispatch_RAF` method, though it's not called out in the text. I moved loading the runtime into `add_window`.

There are also various text changes, but the major one is to combine the two iframe scripting sections into one, while moving the `window.parent` text into the next section, which will now be both `postMessage` and `window.parent`, the two ways of having iframes communicate.

More minor changes include reordering the `window_id` refactors and dropping one.

### DIFF
--- a/book/embeds.md
+++ b/book/embeds.md
@@ -1669,7 +1669,7 @@ have modified the `runtime.js` file to store the `LISTENERS` on the
 `window` object, meaning each `Frame` will have its own set of event
 listeners to dispatch to:
 
-```
+``` {.js}
 window.LISTENERS = {}
 
 // ...
--- a/book/embeds.md
+++ b/book/embeds.md
@@ -1500,7 +1500,8 @@ implement that.
 
 For two frames' JavaScript environments to interact, we'll need to put
 them in the same `JSContext`. So, instead of each `Frame` having a
-`JSContext` of its own, we'll want to store `JSContext`s on the `Tab`:
+`JSContext` of its own, we'll want to store `JSContext`s on the `Tab`,
+in a dictionary that maps origins to JS contexts:
 
 ``` {.python}
 class Tab:
@@ -1524,7 +1525,7 @@ class Frame:
         # ...
 ```
 
-We've now got multiple page's scripts living inside one JavaScript
+We've now got multiple pages' scripts living inside one JavaScript
 context, so we've got to keep them separate somehow. The key is going
 to be the `window` global, of type `Window`. In the browser, this
 refers to the [global object][global-object], and instead of writing a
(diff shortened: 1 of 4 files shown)


pavpanchekha:
Unfortunately I think the last section, on `postMessage` and `window.parent` is sort of blocked on this PR (because I moved text between sections), so I'd appreciate it, @chrishtr, if you could prioritize the review here. If it's a matter of some typos, feel free to push a commit onto this branch and merge it.

book/embeds.md
```diff
@@ -1489,155 +1489,240 @@ effects.
...
+now, each `Frame` has its own `JSContext`, so these scripts can't
+really interact with each other. In reality, *same-origin* iframes run
+in the same JavaScript context and can access each other's globals,
+call each other's functions, and modify each other's DOMs. Let's
+implement that.
 
-[global-object]: https://developer.mozilla.org/en-US/docs/Glossary/Global_object
+For two frames' JavaScript environments to interact, we'll need to put
+them in the same `JSContext`. So, instead of each `Frame` having a
+`JSContext` of its own, we'll want to store `JSContext`s on the `Tab`:
```
chrishtr, line 1503:
```suggestion
`JSContext` of its own, we'll want to store `JSContext`s on the `Tab`, in a
dictionary that maps from origin to JS context:
```

book/embeds.md
```diff
@@ -1489,155 +1489,240 @@ effects.
...
+```
 
-Instead of switching to whole new JavaScript runtime, I'll just approximate the
-feature with two tricks: overwriting the `window` object and the `with`
-operator. The `with` operator is pretty obscure, but what it does is evaluate
-the content of a block by looking up objects on the given object first, and
-only after falling back to the global scope.^[It's important to reiterate that
-this is a hack and doesn't actually do things correctly, but it suffices to
-show the concept in our toy browser.] This example:
+We've now got multiple page's scripts living inside one JavaScript
```
chrishtr, line 1527:
```suggestion
We've now got multiple pages' scripts living inside one JavaScript
```

book/embeds.md
```diff
@@ -1489,155 +1489,240 @@ effects.
...
+:::
 
-class Frame:
-    def __init__(self, tab, parent_frame, frame_element):
-        self.parent_frame = parent_frame
+[global-object]: https://developer.mozilla.org/en-US/docs/Glossary/Global_object
+
+To get multiple frames' scripts to play nice inside one JavaScript
+context, we'll create multiple `Window` objects, so imagine having a
+`window1`, a `window2`, and so on. Before running a frame's scripts,
```
chrishtr, line 1566:
```suggestion
`window_1`, a `window_2`, and so on. Before running a frame's scripts,
```

chrishtr (approved):

PR #765, opened 2023-03-14
Fix scheduling chapter to agree with code for schedule_task

### LINKED ISSUE #748: Feedback on /scheduling.html
A reader points out that in `TaskRunner.schedule_task`, after we append a task to the task queue, we should also `notify_all` on the condition variable. This makes sense to me—I'm not sure if we don't do this now, or if there's just a mismatch between the book and the code.

chrishtr:
Fixes #748

### DIFF
--- a/book/scheduling.md
+++ b/book/scheduling.md
@@ -92,7 +92,7 @@ To run those tasks, we need to call the `run` method on our
 ``` {.python expected=False}
 class Tab:
     def __init__(self):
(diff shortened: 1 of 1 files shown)


pavpanchekha:
Would be good to figure out how to make that block not expected false, but that can be for later

chrishtr:
I tried that but it was complicated.

PR #769, opened 2023-03-14
Pass on communication section

pavpanchekha:
This PR does a pass on the `window.parent` / `window.postMessage` section. The changes are pretty minor, stuff like renaming `PostMessageEvent` to `MessageEvent` (that's what it's called in a real browser?) or adding a global window ID to Window map.

I have two questions—let's merge this and settle them in new PRs.

**Configurable**: Why did you make the `window.parent` property `configurable`? It's the first I've seen this property; I read the description but have no strong feelings whether or not it should be. It's one line of code, so not a big deal either way, but it would be good to explain it, and possibly other properties we define should also be configurable?

**postMessage**: Does postmessage to cross-origin frames actually work? I think looking at the current code, you'd get a bogus target window ID of `-1` and the message wouldn't go anywhere. I think I'm missing something, so please explain what I'm missing so we can make sure to explain it correctly.

### DIFF
--- a/book/embeds.md
+++ b/book/embeds.md
@@ -1863,7 +1863,7 @@ Window.prototype.dispatchEvent = function(evt) {
 ```
 
 That's everything on the receiver side; now let's do the sender side and implement
-The `postMessage` API itself. Note that `this._id` is the ID of the
+the `postMessage` API itself. Note that `this._id` is the ID of the
 receiver or target window:
 
 ``` {.javascript}
--- a/book/embeds.md
+++ b/book/embeds.md
@@ -1774,7 +1774,7 @@ code return a fresh `Window` object with a fake, negative id. This
 makes further accesses to that window's APIs to fail, making it
 impossible to modify that frame's state.^[Note however that in a real
 browser, this `Window` object is not fake, and some APIs can be called
-on it, most important its `parent` can also be retrieved. There is a
+on it, such as that its `parent` can also be retrieved. There is a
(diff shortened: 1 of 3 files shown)


book/embeds.md
```diff
@@ -1702,144 +1702,142 @@ mis-feature---what do you think?
...
-between two different event loops that doesn't require any shared variable
-state or locks.
+Note that it's possible for the lookup in `WINDOWS` to fail, if the
+parent frame is not in the same origin as the current one and
+therefore isn't running in the same `JSContext`. In that case, this
+code return a fresh `Window` object with a fake, negative id. This
+makes further accesses to that window's APIs to fail, making it
+impossible to modify that frame's state.^[Note however that in a real
+browser, this `Window` object is not fake, and some APIs can be called
+on it, most important its `parent` can also be retrieved. There is a
```
chrishtr, line 1777:
```suggestion
on it, such as that its `parent` can also be retrieved. There is a
```

book/embeds.md
```diff
@@ -1864,23 +1862,17 @@ Window.prototype.dispatchEvent = function(evt) {
 }
 ```
 
-And finally, there is the `postMessage` method itself. It has to pass `self._id`
-because the post message is broadcast to all windows *except* the current one:
+That's everything on the receiver side; now let's do the sender side.
```
chrishtr, line 1865:
```suggestion
That's everything on the receiver side; now let's do the sender side and implement
```

PR #774, opened 2023-03-17
Fix a11y hit test hover rectangles

chrishtr:


book/embeds.md
```diff
@@ -1341,6 +1341,57 @@ class FrameAccessibilityNode(AccessibilityNode):
         return node
 ```
 
+Hit testing should now work, but the bounds of the hovered node when drawn
+to the screen are still wrong. For that, we'll need a method that returns
+the absolute screen rect of an `AccessibilityNode`. And that method in turn
```
pavpanchekha, line 1346:
```suggestion
the absolute screen position of an `AccessibilityNode`. And that method in turn
```

book/embeds.md
```diff
@@ -1341,6 +1341,57 @@ class FrameAccessibilityNode(AccessibilityNode):
...
+needs parent pointers so we can walk up the tree, so let's add that first:
```
pavpanchekha, line 1347:
```suggestion
needs parent pointers to walk up the accessibility tree, so let's add that first:
```

book/embeds.md
```diff
@@ -1341,6 +1341,57 @@ class FrameAccessibilityNode(AccessibilityNode):
...
+        obj = self
+        while obj:
+            obj.map_to_parent(rect)
+            obj = obj.parent
+        return rect
+```
+
+This method does nothing until we've defined `map_to_parent` of course, and
+that does nothing for most accessibility nodes (because they are in the same
+coordinate space of their parent):
```
pavpanchekha, line 1379:
```suggestion
This method depends calls `map_to_parent` to adjust the bounds. For
 most accessibility nodes we don't need to do anything, because they are in the same
coordinate space as their parent:
```

book/embeds.md
```diff
@@ -1341,6 +1341,57 @@ class FrameAccessibilityNode(AccessibilityNode):
...
+
+``` {.python}
+class AccessibilityNode:
+    def map_to_parent(self, rect):
+        pass
+```
+
+But not for a `FrameAccessibilityNode`:
```
pavpanchekha, line 1387:
```suggestion
A `FrameAccessibilityNode`, on the other hand, adjusts for the iframe's position:
```

pavpanchekha (approved):

pavpanchekha:
Sorry—I got too involved debugging something. I have language suggestions but go ahead and merge after you go over them, no need to re-review.

book/embeds.md
chrishtr: It's actually a rect, not a position.

book/embeds.md
```diff
@@ -1341,6 +1341,57 @@ class FrameAccessibilityNode(AccessibilityNode):
...
+            self.bounds.x(), self.bounds.y(),
+            self.bounds.width(), self.bounds.height())
+        obj = self
+        while obj:
+            obj.map_to_parent(rect)
+            obj = obj.parent
+        return rect
+```
+
+This method depends calls `map_to_parent` to adjust the bounds. For
```
chrishtr, line 1377:
```suggestion
This method depends on calls `map_to_parent` to adjust the bounds. For
```

PR #776, opened 2023-03-18
Editorial fixes to chapter 15, plus one bugfix

chrishtr:
The bug was that scroll was incorrectly accessed in `FrameAccessibilityNode` (not allowed to access DOM from the
browser thread), and that it wasn't taken into account in `map_to_parent`.

### DIFF
--- a/book/embeds.md
+++ b/book/embeds.md
@@ -17,7 +17,7 @@ Images
 ======
 
 Images are certainly the most popular kind of embedded
-content[^img-late] on the web, dating back to [early
+content on the web,[^img-late] dating back to [early
 1993][img-email].[^img-history] They're included on web pages via the
 `<img>` tag, which looks like this:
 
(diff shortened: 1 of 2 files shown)


book/embeds.md
```diff
@@ -496,7 +517,7 @@ Videos are similar to images, but demand more bandwidth, time, and
 memory; they also have complications like [Digital Rights Management
 (DRM)][drm]. The `<video>` tag addresses some of that, with built-in
 support for advanced video [*codecs*][codec]^[In video, it's called a
-"codec", but in images it's called a "format"--go figure.], DRM, and
+"codec", but in images it's called a "format",--go figure.] DRM, and
```
pavpanchekha, line 520: This seems wrong. You meant to move it before the footnote?

pavpanchekha (commented):
Good catch!

book/embeds.md
chrishtr: Oops, fixed in new PR

PR #777, opened 2023-03-19
A few more editorial fixes to chapter 15, and a bugfix

chrishtr:
The bug was threefold regarding outlines:

* We weren't adjusting outlines for zoom
* Outlines were painted before clip (in some corner cases on the real web the new way is slightly wrong also I think, due to interaction with visual effects)
* We weren't respecting the `thickness` parameter

I also changed the default focus border and iframe border to 1px.

### DIFF
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -1145,8 +1145,10 @@ def paint_outline(node, cmds, rect, zoom):
         cmds.append(DrawOutline(rect, "black", 1))
 ```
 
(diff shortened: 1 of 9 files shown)


pavpanchekha (approved):

pavpanchekha:
Oof. I think this confirms that `zoom` should be a field of the layout object (not a parameter passed through everything), a refactoring we'll need for Ch16 anyway. It's also worth thinking about whether there's a less error-prone way we could organize the paint methods in general.

PR #781, opened 2023-03-21
Make toString async, and fix CSS for the generated browsers

chrishtr:
The makefile was missing lab1, and toString needs to be async because it calls async functions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,7 @@ examples: $(patsubst %,www/examples/example%.html,$(EXAMPLE_HTML)) \
 	$(patsubst %,www/examples/example%.css,$(EXAMPLE_CSS))
 
 widgets: \
+	www/widgets/lab1.js \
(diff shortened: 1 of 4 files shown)


pavpanchekha:
Hmm, I think we have toString as not-async because we want to add things to strings and have it call toString. But that implicit call isn't going to be async. I'm worried this breaks something in, say, one of the widgets (lab4-parser?)

chrishtr:
I looked at all of the generated js files and none call `toString`, because it's a debugging method not called by regular
browser code.

I also audited all of the browsers through chapter 8, and they all work fine. Chapter 9's browser is broken for a second
reason I haven't yet debugged (will fix that later).

chrishtr:
The CSS was also broken for the browsers, because the canvas was underneath the control at the top.

pavpanchekha (approved):

PR #783, opened 2023-03-23
Ensure all files are tested by runtests.py

pavpanchekha:
This PR adds a new check to runtests.py: every file (except a list of ignored files) is tested. This helped identify a couple of files that should be tested, and also a few files that seemed to have no meaningful content (for example, server11.py only different from server10.py in that it was missing a few bug fixes). I also fixed outlines to work in Chapter 15.


src/lab14.py
```diff
@@ -322,6 +322,7 @@ def cascade_priority(rule):
     media, selector, body = rule
     return selector.priority
 
+
```
chrishtr, line 325: supernit: remove

chrishtr (approved):

PR #786, opened 2023-03-31
Add chapter 11 browser widget

chrishtr:
* Uses CanvasKit (the entire Skia binary cross-compiled to WebAssembly) for Skia; license included.

Limitations:
* Non-default font styling (italics, bold) not yet supported


pavpanchekha (approved):
Not sure why the wrap_class constants can’t be static variables, but if it all works, that can be refactored later.

PR #789, opened 2023-03-31
Use `@wbetools.patch` on early chapters

pavpanchekha:
This PR deletes a bunch of duplicate code from earlier chapters using the `wbetools.patch` code. It has to delete one test which relies on running `lab6` and `lab7` in parallel to get two different layouts. It also pulls in an improved version of `wbetools.patch`, which was originally written for Chapter 16.

### DIFF
--- a/book/layout.md
+++ b/book/layout.md
@@ -258,8 +258,6 @@ class BlockLayout:
                 self.children.append(next)
                 previous = next
         else:
-            self.display_list = []
-
             self.cursor_x = 0
             self.cursor_y = 0
             self.weight = "normal"
--- a/infra/asttools.py
+++ b/infra/asttools.py
@@ -140,13 +140,14 @@ class ResolvePatches(ast.NodeTransformer):
                 for patch in patches:
                     for repl_name, repl_stmt in iter_methods(patch):
                         body2[repl_name] = repl_stmt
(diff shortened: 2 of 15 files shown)


pavpanchekha:
This can't be merged right now because the compiler doesn't support patches.

pavpanchekha:
Everything now works _except_ that while the compiler supports patching, it doesn't support patching functions. Which is frustrating!

pavpanchekha:
Ok, I changed Chapter 8 to not require patches. This does kind-of work, but it really only sets us up for more complications if we ever get labs 15 & 16 working, since they also patch functions. Still, even if eventually we find we cannot patch functions, only classes, I think it won't be so bad to inline all of their callers.

pavpanchekha:
Ok, @chrishtr—this uses `@wbetools.patch` in a normal, standard way to get rid of about 1k lines of duplicate code throughout Chapters 1–11. The hope is that this makes cross-chapter changes easier.

One challenge I did run into is that the compiler doesn't support `patch`ing functions. If it did, we could delete some more duplicate code. But I tried for a while and I don't think JS makes this possible, and even in Python, it's pretty shady. Anyway for now I didn't use it, just more basic patching of classes.

chrishtr (approved):

PR #792, opened 2023-04-01
Add zoom as a layout object constructor parameter

chrishtr:
I also fixed a bug in `handle_enter` in chapter 14 that I found while debugging.

### DIFF
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -113,8 +113,8 @@ and abilities.
 ::: {.further}
 In the United States, the European Union, and many other countries,
 website accessibility is in many cases legally required. For example,
-United States Government websites are required to be accessible under [Section
-508][sec508] of the [Rehabilitation Act Amendments of
+United States Government websites are required to be accessible under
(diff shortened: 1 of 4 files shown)


pavpanchekha:
Hi Chris, the `handle_enter` fix is good but the zoom change isn't what I was expecting. I was thinking that we'd have `DocumentLayout` still receive a `zoom` argument to its `layout` method, just like the `frame_width` is, but all other layout objects do not receive it as an argument, and instead compute it like this:

```
class BlockLayout:
    def layout(self):
        self.zoom = self.parent.zoom
```

This means we're computing zoom in basically the same way we compute `width` or `x` or any of the other layout fields, which will make Ch16 more uniform.

pavpanchekha (changes requested):
For Chapter 16, we want zoom to work like `width` or `x`; see comment on PR.

chrishtr:
Done, PR updated.

pavpanchekha (approved):

PR #793, opened 2023-04-04
Refactor to share main function code across chapters

chrishtr:


pavpanchekha (approved):
Eh, I'm happy to merge this—it doesn't make things worse—but I think we should go back into Chapter 11 and actually tell readers to make a `mainloop` function. I think it'd make the chapter better (it'd replace a `tkinter.mainloop` function) and then in later functions we can call out `main` explicitly.

I like the idea of moving the argument parsing logic out of `main`, since the argument parsing logic is really there for setting WBE's internal flags.

PR #800, opened 2023-04-20
Add chapter 12 browser widget

chrishtr:
Notes:

* Runs in single-threaded mode for now.
* I added `truthy` to while loops, because there is a while loop in the code with an array as a boolean argument.
* I renamed some methods to not conflict with library method names.


chrishtr:
Fixed the Task code to remove the extra argument, please re-review.

pavpanchekha (approved):

PR #810, opened 2023-05-04
Require animation frame when scrolling in lab15

pavpanchekha:
#808 didn't fix scrolling in `lab15.py`

I believe I tracked down why—we mis-fixed it in #808. @chrishtr please check that I'm right, but the change is needed for Chapter 16.


chrishtr:
Looks correct to me - I didn't fix it correctly for the root frame.

PR #811, opened 2023-05-05
Add a new "DependantField" abstraction

pavpanchekha:
This PR refactors the Chapter 16 invalidation code to use a new `DependantField` abstraction. A `DependantField` stores:

- A value
- A dirty bit
- A list of dependencies
- A list of things that depend on this value

A typical use is something like this:

```
class FooLayout:
    def __init__(self, ...):
        # ...
        self.foo_field = DependantField(self, "foo")

    def layout(self):
        parent_foo = self.foo_field.read(self.parent.foo_field)
        self.foo_field.write(parent_foo * 2)
```

Here the `read` instruction records the dependencies while the `write` sets all fields that depend on this one to be dirty.

I'm still working on:

- [ ] How to describe this in text, specifically how to build this up step by step
- [ ] How to incorporate things outside layout (`dirty_style`, `dirty_children`)
- [ ] How to incorporate `dirty_descendants`

I think these issues are resolvable.


src/lab16.py
```diff
@@ -221,28 +221,35 @@ def __init__(self, node, parent, previous):
...
+            self.y = self.y_field.set(parent_y)
 
         for word in self.children:
             word.layout()
 
         if not self.children:
-            self.height = 0
-            self.parent.dirty_height = True
-            mark_dirty(self.parent)
+            self.height = self.height_field.set(0)
```
chrishtr, line 252: This misses dirtying self.parent.

src/lab16.py
```diff
@@ -252,14 +259,37 @@ def layout(self):
...
                            for child in self.children])
-        self.height = max_ascent + max_descent
-        self.parent.dirty_height = True
-        mark_dirty(self.parent)
 
-        self.dirty_width = False
-        self.dirty_height = False
-        self.dirty_x = False
-        self.dirty_y = False
+        self.height = self.height_field.set(max_ascent + max_descent)
```
chrishtr, line 263: Doesn't this mean that height_field should depend on those children? How is that represented?

src/lab16.py
```diff
@@ -278,69 +308,49 @@ def __init__(self, node, parent, previous, frame):
...
-            mark_dirty(self)
-            self.dirty_width = True
-            self.dirty_zoom = False
+        parent_zoom = self.zoom_field.read(self.parent.zoom_field)
+        self.zoom = self.zoom_field.set(parent_zoom)
 
         if self.dirty_style:
             self.dirty_children = True
-            self.dirty_width = True
+            self.width_field.dirty = True
```
chrishtr, line 328: This is the only place you explicitly set dirty to true. Is there something special about it?

src/lab16.py
```diff
@@ -278,69 +308,49 @@ def __init__(self, node, parent, previous, frame):
...
-            assert not self.parent.dirty_width
-            if "width" in self.node.style:
-                self.width = device_px(float(self.node.style["width"][:-2]), zoom)
-            else:
-                self.width = self.parent.width
+        assert not self.dirty_style
+        old_w = self.width
+        if "width" in self.node.style:
+            zoom = self.width_field.read(self.zoom_field)
+            self.width = self.width_field.set(float(self.node.style["width"][:-2], zoom))
```
chrishtr, line 336: You missed device_px here, I think.

src/lab16.py
```diff
@@ -278,69 +308,49 @@ def __init__(self, node, parent, previous, frame):
...
+        else:
+            parent_width = self.width_field.read(self.parent.width_field)
+            self.width = self.width_field.set(parent_width)
+        if self.width != old_w:
             self.dirty_children = True
-            for child in self.children:
```
chrishtr, line 315: Am I right that we get this for free because the first time through we recurse into children, and they mark themselves as depending on their parent?

src/lab16.py
```diff
@@ -252,14 +259,37 @@ def layout(self):
...
-        self.parent.dirty_height = True
-        mark_dirty(self.parent)
 
-        self.dirty_width = False
-        self.dirty_height = False
-        self.dirty_x = False
-        self.dirty_y = False
+        self.height = self.height_field.set(max_ascent + max_descent)
+        
+class DependentField:
```
chrishtr, line 265: Two main comments about this abstraction:

* It's common to simply set a field based on the field of the thing it depends on. I suggest making a version of set that does this automatically, without first setting a local variable. The other common pattern is adding together two depended-on values (and maybe that's all the cases?). Also make a version for adding two things?

* It's weird that there are values and dependent fields that represent that value, and duplicate the value itself. 
e.g. there is `zoom` and `zoom_field`. Why do we need both?

src/lab16.py
pavpanchekha: The parent should depend on the child height and be dirtied by the set operation.

src/lab16.py
pavpanchekha: Great great question. Right now I don’t do invalidation for elements of lines. Not sure if we should. Probably? There may be a bug here, too.

src/lab16.py
pavpanchekha: Good catch.

src/lab16.py
pavpanchekha: Yes, exactly. Every field remembers who read it, and dirties those on change.

src/lab16.py
pavpanchekha: Point one makes sense. I can introduce those helpers.

pavpanchekha:
Ok, I pushed a bunch of new commits, which implement:

- A new "ChangeSource" class and a helper "FieldManager" class for passing along change notifications _without_ having a value or a dirty bit
- A new option to have a dependent field be "eager" or "lazy"
- Complete handling of the dirty descendants bit
- Minimal but working style invalidation

I don't think it has any new bugs the old one didn't have. However, there's now a true menagerie of different signaling mechanisms, which is both ugly and also hard to explain. I actually tried to make a simpler implementation first, but I couldn't get it working. So instead I made this complicated, working one.

My new plan is to take this working version and try to simplify it before writing about it. I hope to at least go from three to two classes, which ideally would be a class-and-subclass. I think the idea of "eager" fields is necessary and I think I have a way to explain that, but perhaps that can also be simplified.

pavpanchekha:
Ok @chrishtr, here's an update on the status of this branch and the "protected field" approach in general.

- The whole chapter has been rewritten, twice now, to follow the `ProtectedField` approach. It works and is reasonably fast (layout & style under a ms for typing into a `contenteditable` on a large page)
- The chapter and code are now synced up. The lints (almost, see below) pass and it runs and works on real-world examples

A few things _are_ missing, though some do not need to go into this PR:

- I haven't written the section on granular style invalidation, but I did test the code and it won't be too hard.
- I also haven't written a summary or go furthers or exercises
- There are underinvalidation bugs when you modify an inline element, because inline elements have no layout objects. I believe this can be fixed by traversing up from the inline element to the first ancestor with a `BlockLayout` layout object.[^1]
- There is an issue with changing `ascent` and `descent` to fields. I believe this can be fixed by having `EmbedLayout` provide `layout_before` and `layout_after` methods instead of just an overridable `layout` method.
- There is an issue with leading on text nodes. I believe this can be fixed by subtracting leading in `TextLayout` paint.

@chrishtr, if you could give this chapter a critical read soon that would be much appreciated. I believe it is in a relatively good state, but your first read will test that!

I plan to take a break from this chapter tomorrow to do other work tasks, and hopefully on Monday can respond to your comments. If you can't get to it then, I'll work on fixing the issues above and if I'm satisfied with everything and you don't object, I'd like to merge this, just so we don't have a mega-branch sitting off of `main`.

[^1]: There are alternatives (we could protect the `children` field on nodes; this is technically no problem, but leads to a bunch of annoying API changes in other parts of the code like the HTML parser)

pavpanchekha:
I fixed the three issues noted above. I think there may be an issue with changing styles on inline elements. I plan to fix that with a tweak to how eager propagation works. Other than that, I know of no current bugs or issues with the code, so my hope is we can merge this (once the bug is fixed) and then work on granular style invalidation and go furthers in a separate PR.

PR #812, opened 2023-05-12
Remove `example` class on homepage elements.

chrishtr:
Right now there are borders and the word "example" because of these errant classes.

### DIFF
--- a/src/lab16.py
+++ b/src/lab16.py
@@ -290,6 +290,12 @@ class BlockLayout:
         self.dirty_descendants = True
 
     def layout(self):
+        old_zoom = self.zoom
+        old_width = self.width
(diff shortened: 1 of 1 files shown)


chrishtr:
This will need to get pushed to the site asap.

pavpanchekha:
I'm confused why this PR is important? It seems to add some spare lines of code to ch16 that just define variables that don't get used. Did you push the right commit?

chrishtr:
Ok that's super weird. The stuff I uploaded is wrong I agree.

I meant to upload a PR that removes class="example" from some elements on the home page, which is broken right now (check out http://browser.engineering). But the ToT code looks fine it seems.

Issue #815, opened 2023-05-15
Example boxes look strange


wilcoxjay:
Check out this example from Chapter 8
![example from Chapter 8](https://github.com/browserengineering/book/assets/229924/3b7a66ea-fc32-4c32-988b-c111141e96a5)

There seems to be a missing newline or something. Perhaps [this CSS](https://github.com/browserengineering/book/blob/8e5a4ace74ea39fb44d57f83c7b9e3bdb4dda838/www/book.css#L139) should contain 
```
.example:before { content: "Example\0a\0a"; color: gray; }
```
which adds two newline characters.

PR #817, opened 2023-05-16
Add newlines after content:example

chrishtr:
The "Example" class is always applied to a `pre` element in the book, so we need to supply our own newlines.


chrishtr:
Fixes #815

pavpanchekha:
I think we just shouldn't use that name for a block. I will put out a better fix.

PR #818, opened 2023-05-16
Lab 13 browser

chrishtr:
Notable:

* Added a mechanism to patch modularized classes in JavaScript by copying prototypes
* Made small edits to some python code to make it easier to compile


infra/compile.py
```diff
@@ -72,8 +73,8 @@ def find_hint(t, key, error=None):
         break
     else:
         ln = t.lineno if hasattr(t, "lineno") else -1
-        hint = {"line": ln, "code": asttools.unparse(t, explain=True), key: "???"}
```
chrishtr, line 75: The parser of this output doesn't support "line".

Makefile
```diff
@@ -26,7 +26,8 @@ widgets: \
 	www/widgets/lab9-browser.html www/widgets/lab9.js www/widgets/server9.js \
 	www/widgets/lab10-browser.html www/widgets/lab10.js www/widgets/server10.js \
 	www/widgets/lab11-browser.html www/widgets/lab11.js \
-	www/widgets/lab12-browser.html www/widgets/lab12.js
+	www/widgets/lab12-browser.html www/widgets/lab12.js \
+	www/widgets/lab12-browser.html www/widgets/lab13.js
```
pavpanchekha, line 30: Uhh, is this a typo? `lab12-browser.html` appears twice!

infra/compile.py
```diff
@@ -157,9 +161,20 @@ def check_args(args, ctx):
...
+    "MakeXYWH",
+    "left",
+    "top",
+    "bottom",
+    "width",
+    "height",
+    "isEmpty",
+    "roundOut",
+    "intersects",
+    "Intersects",
```
pavpanchekha, line 176: Is there really an `Intersects` method?

infra/compile.py
```diff
@@ -272,6 +302,8 @@ def compile_method(base, name, args, ctx):
         if name in OUR_SYNC_METHODS:
             return base_js + "." + name + "(" + ", ".join(args_js) + ")"
         else:
+            if name == "__init__":
+                name = "init"
```
pavpanchekha, line 306: Good change

infra/compile.py
```diff
@@ -661,15 +708,25 @@ def compile(tree, ctx, indent=0):
             out += "\n" + " " * indent + "constants.{} = {}_constants.{};".format(const, tree.module, const)
         return out
     elif isinstance(tree, ast.ClassDef):
-        assert not tree.bases
-        assert not tree.keywords
-        assert not tree.decorator_list
+        if tree.decorator_list:
+            assert tree.decorator_list[0].func.value.id == "wbetools"
```
pavpanchekha, line 712: I'm a little worried about the lack of an `isinstance` assertion on the `tree.decorator_list[0].func`, and also the fact that it's not testing every decorator (just the first) though I'm sure it's fine.

infra/compile.py
```diff
@@ -111,8 +112,11 @@ def check_args(args, ctx):
     "int": "parseInt",
     "float": "parseFloat",
     "print": "console.log",
+    "AnimationClass": "AnimationClass"
 }
 
+SKIPPED_FNS = ["add_main_args"]
```
pavpanchekha, line 118: Are we skipping it because it's hard to compile? If not, why not just compile it and then just don't call it?

infra/compile.py
```diff
@@ -749,10 +806,17 @@ def compile(tree, ctx, indent=0):
         out += "\n" + " " * indent + "}"
         return out
     elif isinstance(tree, ast.For):
+        try:
+            t = find_hint(tree.iter, "type")
+        except MissingHint:
+            t = "list"
+        assert t in ["dict", "list"]
```
pavpanchekha, line 813: I'm fine with this code but if we ever do iterate over a `dict` I feel like it improves readability to write `for i in foo.keys()`

pavpanchekha (approved):

Makefile
chrishtr: Fixed

infra/compile.py
chrishtr: It's because it's work to make it compile. e.g. I'd have to mock the class that parses arguments.

infra/compile.py
chrishtr: Yes. Skia.Rect.Intersects (static function)

There is no "intersects" though, removed that.

PR #820, opened 2023-05-23
Prefetch web fonts

chrishtr:
This should make them load a bit faster.

(I'm also working on a Chrome feature that blocks rendering until such fonts have loaded, and want to see if it works
on our site.)

I found the URLs by looking at the Network in tab in Chrome devtools when loading the site.

### DIFF
--- a/infra/template.html
+++ b/infra/template.html
@@ -2,6 +2,11 @@
 <html lang="en-US" xml:lang="en-US">
 <head>
   <meta charset="utf-8" />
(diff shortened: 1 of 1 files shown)


chrishtr:
The URLs do change. We can't self-host without copying the entire style sheet (which we could do if we want...)

See comment above for how I got the URLs.

pavpanchekha:
Yeah, I meant leaving a comment in the template saying how we got the URLs—so that when they break it's obvious what to do. Not a big deal either way.

PR #823, opened 2023-05-27
Fix root canvas variable for lab2-render

chrishtr:
Also deletes labN-browser.html, which doesn't appear to be used.


pavpanchekha (commented):
Are you 100% sure that labN-browser.html isn't used? I think it was, at least initially, a template used by `Makefile` to create the browser HTML pages. Though I agree that it probably makes more sense to delete it and commit all the generated pages, which I think have ended up diverging a bit.

chrishtr:
I looked in the Makefile, and `grep -r labN` yields no hits, so I think so.

PR #825, opened 2023-05-29
Reuse the window object across resizes

chrishtr:
Previously, every embedded browser resize (resulting from resizing the outer browser window, via the `resize` event), would create a new `Browser` object *and `tkinter.Tk` object. After resize, the old window and browser would remain active and compete for the screen.

Now we keep around one `tkinter.Tk` object across resizes and rebind events to it. Old `Browser` objects are unref'd and therefore ready for garbage collection. `clearInterval` is also called when needed.

### DIFF
--- a/www/widgets/rt.js
+++ b/www/widgets/rt.js
@@ -22,6 +22,7 @@ function wrap_class(cls, fn) {
     return f;
 }
 
+
 function http_ok(body, headers) {
     let text = "HTTP/1.0 200 OK\r\n";
     for (let [k, v] of Object.entries(headers || {})) {
@@ -40,6 +41,7 @@ const rt_constants = {};
 rt_constants.ZOOM = 1.0;
 rt_constants.ROOT_CANVAS = null;
 rt_constants.URLS = {};
+rt_constants.WINDOW = null;
 
 let fontManager;
 let CanvasKit;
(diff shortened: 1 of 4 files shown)


www/widgets/rt.js
```diff
@@ -164,55 +166,90 @@ class ssl {
 }
 
 class tkinter { 
-    static Tk = wrap_class(class {
-        constructor() {
-            this.elt = rt_constants.ROOT_CANVAS;
+    static Tk = function(...args) {
+        if (rt_constants.WINDOW) {
+            console.log('reused')
```
pavpanchekha, line 171: Do we need to keep this debug print? (If we do, we should make it more informative at least.)

www/widgets/rt.js
```diff
@@ -164,55 +166,90 @@ class ssl {
...
+                    };
+                } else {
+                    console.error("Trying to bind unsupported event", key);
+                }
+                if (key in this.bindings) {
+                    this.bindings[key] = fn;
+                    return;
+                }
+                this.bindings[key] = fn;
+                if (!(event_name in this.event_names_to_key)) {
```
pavpanchekha, line 238: JavaScript has an `in` keyword now?

www/widgets/rt.js
```diff
@@ -164,55 +166,90 @@ class ssl {
...
-            } else if (key == '<Configure>') {
-                window.addEventListener("resize", function(e) {
-                    fn({}).catch(on_error);
-                });
-            } else if (key == '<Key>') {
-                window.addEventListener("keydown", function(e) {
-                    if (e.key.length == 1) {
-                        e.preventDefault();
-                        fn({ char: e.key }).catch(on_error);
+        return wrap_class(class {
```
pavpanchekha, line 175: I'm a little confused what was changed by the diff here.

pavpanchekha (commented):
Maybe I'm missing something, but where do we _unbind_ old `bind` handlers? Or do we avoid re-running `bind`?

pavpanchekha:
I gotta say, I didn't totally understand this PR. Perhaps you should merge it anyway, if you're confident it fixes a bug, but I'm a little lost as to the series of events that leads to us rendering to a canvas from two different browsers.

www/widgets/rt.js
chrishtr: yeah

www/widgets/rt.js
```diff
@@ -164,55 +166,90 @@ class ssl {
...
+                    fn = function(e) {
+                        if (e.key == key) {
+                            e.preventDefault();
+                            callback({ char: e.key }).catch(on_error);
+                        }
+                    };
+                } else {
+                    console.error("Trying to bind unsupported event", key);
+                }
+                if (key in this.bindings) {
```
chrishtr, line 232: Here is where I avoid re-binding. It changes the bound function and then returns.

www/widgets/rt.js
chrishtr: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in

www/widgets/rt.js
chrishtr: The Tk function now does this:

If `rt_constants.WINDOW` doesn't exist, make it in a wrap_class and return it. Otherwise return `rt_constants.WINDOW`. This implements a singleton pattern, which allows us to preserve the Tk object across resizes, and rebind it to new callback functions when the `Browser` object is re-created.

pavpanchekha:
Thanks for the explanations, I get it now. Cool, let's merge it.

PR #826, opened 2023-05-30
Don't check trivial dirty bits

pavpanchekha:
This PR changes Chapter 16 to only check dirty bits for:

- Rebuilding the layout tree
- Skipping `layout` entirely
- Recomputing `style`

The text changes corresponding to this are probably not the best I could do—I plan to leave that for another PR that does a full read of the chapter.

### DIFF
--- a/book/embeds.md
+++ b/book/embeds.md
@@ -957,11 +957,10 @@ class Frame:
 And we can set those when the parent frame is laid out:[^no-set-needs]
 
 [^no-set-needs]: You might be surprised that I'm not calling
-    `set_needs_render` on the child frame here. Our toy browser is so
-    limited that it's safe not to: the `width` and `height` attributes
-    can't change, and when the `zoom` changes we invalidate all of the
-    frames. But if our browser had, for example, the `setAttribute`
-    method, then we'd need to force that child frame to render.
+    `set_needs_render` on the child frame here. That's a shortcut: the
+    `width` and `height` attributes can only change through
+    `setAttribute`, while `zoom` can only change in `zoom_by` and
+    `reset_zoom`. All of those handlers already invalidate all frames.
 
 ``` {.python}
 class IframeLayout(EmbedLayout):
--- a/book/embeds.md
+++ b/book/embeds.md
@@ -954,7 +954,14 @@ class Frame:
         self.frame_height = 0
 ```
 
-And we can set those when the parent frame is laid out:
+And we can set those when the parent frame is laid out:[^no-set-needs]
+
+[^no-set-needs]: You might be surprised that I'm not calling
+    `set_needs_render` on the child frame here. Our toy browser is so
+    limited that it's safe not to: the `width` and `height` attributes
+    can't change, and when the `zoom` changes we invalidate all of the
+    frames. But if our browser had, for example, the `setAttribute`
+    method, then we'd need to force that child frame to render.
 
 ``` {.python}
 class IframeLayout(EmbedLayout):
--- a/book/invalidation.md
+++ b/book/invalidation.md
@@ -652,9 +652,8 @@ this:
 ``` {.python dropline=self.parent.zoom.get() replace=set(parent_zoom)/copy(self.parent.zoom)}
 class BlockLayout:
     def layout(self):
-        if self.zoom.dirty:
-            parent_zoom = self.parent.zoom.get()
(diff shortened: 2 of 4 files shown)


book/invalidation.md
```diff
@@ -446,7 +446,7 @@ array isn't dirty. Let's find the line at the top of `layout` that
 resets the `children` array, and move it into the two branches of the
 `if` statement:
 
-``` {.python expected=False}
+``` {.python}
```
pavpanchekha, line 449: I dunno why this changed but I don't want to mess with it at the moment.

pavpanchekha: (As in, I don't know why this is now matching, but I don't care.)

book/invalidation.md
```diff
@@ -947,12 +945,17 @@ class BlockLayout:
     def layout(self):
         # ...
         if mode == "block":
-            # ...
+            if self.children.dirty:
+                # ...
```
pavpanchekha, line 949: Also a random change to diffing that shouldn't be affected by this chapter but somehow is.

src/lab16.py
```diff
@@ -233,26 +233,34 @@ def __init__(self, node, frame):
...
+        if self.width.dirty: return True
+        if self.height.dirty: return True
+        if self.x.dirty: return True
+        if self.y.dirty: return True
+        if self.descendants.dirty: return True
+        return False
+
     def layout(self, width, zoom):
         self.zoom.set(zoom)
+        self.width.set(width - 2 * device_px(HSTEP, zoom))
```
chrishtr, line 247: Why put these two lines before the layout_needed check?

pavpanchekha: Well, if the `zoom` or `width` argument has changed, we need to update those fields. The `layout_needed` method will check their dirty bits, but not diff the current value with the argument. We could change how zooming / resizing works but I think this is easier.

chrishtr: If `layout_needed` checks `self.zoom.dirty` and `self.width.dirty`, then i'll always recompute layout if they change, right?

pavpanchekha: I think I figured out the confusion. `zoom` and `width` aren't modified or marked by the `Frame` when the zoom level or frame width changes. (So they're not dirty until the `set` operation.) Would you rather do that—mark those fields dirty when the inputs change?

chrishtr: That would be more consistent, so let's try that.

pavpanchekha: Done. Honestly, I think this approach is a little worse, but take a look and let me know. It's not a huge difference either way.

book/invalidation.md
```diff
@@ -1926,24 +1926,39 @@ class ProtectedField:
...
-    def layout(self, width, zoom):
-        self.zoom.set(zoom)
-        self.width.set(width - 2 * device_px(HSTEP, zoom))
-        if not self.layout_needed(): return
+class IframeLayout(EmbedLayout):
+    def layout(self):
+        if self.node.frame:
+            # ...
+            self.node.frame.document.height.mark()
+            self.node.frame.set_needs_render()
```
chrishtr, line 1990: why is set_needs_render needed here?

pavpanchekha: So, it at least needs to `set_needs_layout`, right?

pavpanchekha:
Decided: remove the `set_needs_render` line but add a sentence to Chapter 15 about why we get to skip it.

pavpanchekha:
Changes decided on in meeting have been done: no more `set_needs_render`, and two footnotes about:

1. Why not call it.
2. Why not protect `frame_width` instead.

book/embeds.md
```diff
@@ -954,7 +954,14 @@ class Frame:
         self.frame_height = 0
 ```
 
-And we can set those when the parent frame is laid out:
+And we can set those when the parent frame is laid out:[^no-set-needs]
+
+[^no-set-needs]: You might be surprised that I'm not calling
```
chrishtr, line 959: I just remembered that chapter 14 introduces `setAttribute`. So this is not true?

chrishtr: I just put up a PR to call set_needs_render from setAttribute. So just rephrase this footnote
to refer to the call there?

pavpanchekha:
Ok, I fixed the footnote. Let's test and, if necessary, fix Chapter 16 in another PR.

PR #829, opened 2023-05-31
Change how our browser computes style values

pavpanchekha:
This PR redoes a bit of Chapter 6 that 1) I was never happy with; 2) students never got; and 3) gets in the way of Chapter 16. The issue is computed styles.

"Computed styles" are a CSS concept that divides the style pass into first computing "specified styles" and then computing "computed styles". Computed styles differ from specified styles because they have resolved a variety of relative values and have clamped properties like paddings to positive values.

In our browser, the only computed value was font-size, which can be specified as a percentage value and which is resolved relative to the parent font size. (Note that _most_ percentage sizes are resolved during layout, but font size is resolved during computed styles because it is the basis of the `em` unit and friends.)

Our browser implemented this via a `compute_style` function which was called on all values before storing them in the style dictionary. This is different from real browsers—instead of being a separate pass, it is fused into prior passes. This also had other problems:

- First, the function was confusing and I saw in class that students didn't understand why you would do something in `compute_style` instead of in `style` itself. (This was hard to answer because this abstraction boundary was fake, not something in a real browser.)
- Second, it had a weird API. Apparently it could output `None` to mean not to store the value at all, but this was only implemented in one of the two cases where it was used. Due to the API boundary itself being unclear, it wasn't clear when you would use this.
- Third, it gets in the way of Chapter 16. In Chapter 16, we want to track dependencies between fields, which means using `ProtectedField`s instead of ordinary values. The `compute_style` API hindered that because it accessed style fields and returned style values. Updating it seemed hard because the API itself made no sense.

This PR goes back into Chapter 6 and redoes `compute_style`. Instead of computing style values whenever we set `node.style`, we just do a "post-pass" at the end of the `style` method to update each computed style value. This is more like a real browser, it drops the API entirely, and is easy for Chapter 16 to deal with.

### DIFF
--- a/book/styles.md
+++ b/book/styles.md
@@ -223,7 +223,7 @@ but before doing layout.
 This `style` function will also fill in the `style` field by parsing
 the element's `style` attribute:
     
-``` {.python replace=(node)/(node%2C%20rules),%3d%20value/%3d%20computed_value}
+``` {.python replace=(node)/(node%2C%20rules)}
 def style(node):
     # ...
     if isinstance(node, Element) and "style" in node.attributes:
@@ -490,7 +490,7 @@ the page, this will require looping over all elements *and* all rules.
 When a rule applies, its property/values pairs are copied to the
 element's style information:
 
-``` {.python replace=%3d%20value/%3d%20computed_value}
+``` {.python}
 def style(node, rules):
     # ...
     for selector, body in rules:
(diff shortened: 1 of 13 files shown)


pavpanchekha:
If we merge this, I'll make a blog post about it.

book/styles.md
```diff
@@ -792,59 +792,39 @@ implementing those other properties in this book.
...
 def style(node, rules):
-    # ...
-    for selector, body in rules:
-        if not selector.matches(node): continue
-        for property, value in body.items():
-            computed_value = compute_style(node, property, value)
-            if not computed_value: continue
-            node.style[property] = computed_value
-    # ...
+    if node.style["font-size"].endswith("%"):
```
chrishtr, line 814:
```suggestion
    # ...
    if node.style["font-size"].endswith("%"):
```

chrishtr (approved):

PR #837, opened 2023-06-03
Add a go-further about ProtectedField and real browsers

chrishtr:


book/invalidation.md
```diff
@@ -1749,8 +1749,8 @@ we wrapped onto more lines). As a bonus, editing should now also feel
 :::
 
 
-Avoiding redundant recursion
-============================
+Avoiding redundant work
+=======================
```
pavpanchekha, line 1753: I agree the original section title is bad, but I don't like the replacement either. Pretty much the whole chapter is redundant work! Maybe "Skipping traversals" or "Skipping unnecessary layouts" or something like that?

pavpanchekha (approved):

PR #838, opened 2023-06-04
Add set_needs_render to setAttribute, since it affects roles and layout

chrishtr:
This was a bug in chapter 14.

It also fixes an error in GL namespaces.


pavpanchekha (approved):

pavpanchekha:
Wait—we have `setAttribute`? Doesn't that mean we actually do need to deal with frame width and height changing?

pavpanchekha:
Definitely something to worry about in Chapter 16 if so.

chrishtr:
Yep we do need to deal with attributes changing iframe size. Need to test and make sure it works..

PR #840, opened 2023-06-07
Add resizing exercise

pavpanchekha:
This adds the resizable browser window exercise that we discussed. It prompts the user to think about both horizontal resizing (wherein invalidation should save style but not layout) and vertical resizing (wherein invalidation should save both).

### DIFF
--- a/book/invalidation.md
+++ b/book/invalidation.md
@@ -2361,10 +2361,10 @@ to be re-laid-out to match the new size.
 
 *Resizing the browser*: Perhaps, back in [Chapter
 2](graphics.md#exercises), you implemented support for resizing the
-browser. Most likely, you dropped support for it since then.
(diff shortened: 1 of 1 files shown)


book/invalidation.md
```diff
@@ -2359,6 +2359,15 @@ correctly when changing these values. For `iframe` elements, make sure
 adjusting these attributes causes both the parent and the child frame
 to be re-laid-out to match the new size.
 
+*Resizing the browser*: Perhaps, back in [Chapter
+2](graphics.md#exercises), you implemented support for resizing the
+browser. Most likely, you dropped support for it since then.
```
chrishtr, line 2364:
```suggestion
browser. (And most likely, you dropped support for it since then, when SDL was added.)
```

chrishtr (approved):

PR #847, opened 2023-06-13
Remove temp_children

chrishtr:


pavpanchekha:
Hmm, on the one hand this definitely works, but I'm a bit worried about it. In general, mutating the value in a `ProtectedField` should involve notifying any dependents. In fact, now that I think about this, I don't think this is safe. Imagine if an element *had* no children, but then some were added. In that case, when you `set` the empty array, it won't notify any dependents, because the value is unchanged. But then when you add stuff, nothing changes.

In general, you need to notify on any *modification* to a `ProtectedField`. I've been trying to treat the value in a `ProtectedField` as immutable for this reason; modifications are dangerous unless you also `notify` the field whenever you do that.

chrishtr:
Can't I fix it by just calling notify()?

pavpanchekha:
I think you merged in your descendants PR in. Yes, it could be fixed by calling `notify` manually, but why is that better than having `temp_children`? Now, instead of a strange protocol involving `temp_children`, we have to introduce and discuss a strange protocol of modifying protected values. (Whereas otherwise we would only ever change them via `set`).

PR #849, opened 2023-06-13
Remove the `descendants` ProtectedField

chrishtr:
And replace it with a dirty bit on each layout object.

`descendants` always depended on all of the `ProtectedField`s for each descendant anyway, so we can just
set the dirty bit directly.

As a result, we can get rid of the concept of an eager field which simplifies the code and the chapter.

### DIFF
--- a/book/invalidation.md
+++ b/book/invalidation.md
@@ -1820,7 +1820,11 @@ dependencies---though there are some differences.
 
 So let's add a new dirty flag, which I call `has_dirty_descendants`,
 to track whether anty descendants have a dirty
-`ProtectedField`.
+`ProtectedField`.[^ancestors]
+
+[^ancestors]: In some code bases, you will see these
+called *ancestor* dirty flags instead. It's the same thing, just
+following the flow of dirty bits instead of the flow of control.
 
 ``` {.python}
 class BlockLayout:
--- a/book/invalidation.md
+++ b/book/invalidation.md
@@ -1869,20 +1869,6 @@ Here's `BlockLayout`, for example:
 
 
 ``` {.python}
-class ProtectedField:
-    def set_ancestor_dirty_bits(self):
-        parent = self.parent
-        while parent:
-            parent.has_dirty_descendants = True
-            parent = parent.parent
-
-    def mark(self):
-        # ...
-        self.set_ancestor_dirty_bits()
-
-    def notify(self):
-        # ...
-
 class BlockLayout:
     def __init__(self, node, parent, previous, frame):
         # ...    
@@ -1895,9 +1881,6 @@ class BlockLayout:
 ```
 
 And then the bit needs to be cleared after `layout`:
-
-For each layout object type, pass the parameter for each `ProtectedField`.
-Here's `BlockLayout`, for example:
 
 ``` {.python}
 class BlockLayout:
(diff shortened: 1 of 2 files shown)


book/invalidation.md
```diff
@@ -1818,114 +1815,82 @@ on calling its parent's `layout` method, and so on. We can apply
 invalidation to control dependencies just like we do to data
 dependencies---though there are some differences.
 
-So let's add a new dirty flag, which I call `descendants`,[^ancestors]
-to track the control dependencies for a node's descendants:
-
-[^ancestors]: In some code bases, you will see these
-called *ancestor* dirty flags instead. It's the same thing, just
-following the flow of dirty bits instead of the flow of control.
```
pavpanchekha, line 1826: Not sure why you dropped this footnote.

book/invalidation.md
```diff
@@ -1818,114 +1815,82 @@ on calling its parent's `layout` method, and so on. We can apply
...
+So let's add a new dirty flag, which I call `has_dirty_descendants`,
+to track whether anty descendants have a dirty
```
pavpanchekha, line 1819:
```suggestion
to track whether any descendants have a dirty
```

book/invalidation.md
```diff
@@ -1075,13 +1075,12 @@ computation we are skipping here---line breaking and rebuilding the
...
-references to `children`. There are couple of possible fixes, but in
-the interests of expediency,[^perhaps-local] I'm going to use a
-second, unprotected field, `temp_children`:
-
-[^perhaps-local]: Perhaps the nicest design would thread a local
-    `children` variable through all of the methods involved in line
-    layout, similar to how we handle `paint`.
+references to `children`. I'll solve this by first setting
+`this.children` to an empty array, then filling it in. This is
+a bit of an expediency, because the `ProtectedValue` for `children`
```
pavpanchekha, line 1080:
```suggestion
a bit of an expediency, because the `ProtectedField` for `children`
```

pavpanchekha (approved):
Ok, you know, I was ready to be very unhappy with this PR but you're right, this is way simpler and less code. I originally used `ProtectedField` for this because I wanted to make it part of the same `mark`/`notify` machinery as the other invalidations, but at the time the structure was different and that would have been hard. But this looks good.

pavpanchekha:
Oh, one more thought about this PR—would this code "just work" if you added `has_dirty_descendant` bits to each `Element`/`Text` and then added `parent` links to style fields? That's an exercise and we want to make it reasonably easy.

Also, can you check that the text about how you modify `layout_needed` is sensible with the changes? It's not in the diff (or I missed it).

pavpanchekha:
Oh, uhh, also, this PR seems to have the `temp_children` one mixed in, which I don't feel so good about. Can we separate them and merge the `has_dirty_descendants` PR without merging the `temp_children` one?

chrishtr:
> Oh, uhh, also, this PR seems to have the `temp_children` one mixed in, which I don't feel so good about. Can we separate them and merge the `has_dirty_descendants` PR without merging the `temp_children` one?

Fixed, oops.

chrishtr:
> Oh, one more thought about this PR—would this code "just work" if you added `has_dirty_descendant` bits to each `Element`/`Text` and then added `parent` links to style fields? That's an exercise and we want to make it reasonably easy.

Yes I think it would work.

chrishtr:
> Also, can you check that the text about how you modify `layout_needed` is sensible with the changes? It's not in the diff (or I missed it).

Checked, and it looks ok to me.

PR #851, opened 2023-06-14
Rename depended_on to invalidations

chrishtr:
I found `depended_on` confusing and hard to say. This new name seems better.


pavpanchekha (approved):
Good name

Issue #850, opened 2023-06-14
Cross-frame DOM methods don't always invalidate correctly


pavpanchekha:
Suppose you have two same-origin frames on a page, A and B, and A runs a script that sets the `style` of an element in frame B. In this case, our browser will `set_needs_render` on A, not B.

PR #852, opened 2023-06-14
Test iframe resizing

pavpanchekha:
This PR adds a test to Chapter 16 that tests what happens when one frame's script resizes, using `setAttribute`, another frame. (It should force a re-layout in that frame.)

Funny enough, Chapter 14 added `setAttribute` but somehow Chapter 15 lost it. So this PR first adds `setAttribute` back into Chapter 15, describes it in the Chapter 15 text, and then adds it to Chapter 16, and describes it in the Chapter 16 text.

For some reason, the test passes without having to protect the `frame_width` field. Not sure why, will investigate tomorrow.

### DIFF
--- a/src/lab16-tests.md
+++ b/src/lab16-tests.md
@@ -206,7 +206,7 @@ The parent frame should now have resized the iframe:
            LineLayout(x=13.0, y=18.0, width=774.0, height=152.0, node=<body>)
              IframeLayout(src=http://test/2, x=13.0, y=18.0, width=102.0, height=152.0)
 
-But also the child frame should have resided as well:
+But also the child frame should have resized as well:
 
     >>> lab16.print_tree(frame2.document)
      DocumentLayout()
--- a/src/lab16-tests.md
+++ b/src/lab16-tests.md
@@ -149,6 +149,10 @@ Next, we can execute some JavaScript to change the page:
 Test iframe resizing
 ====================
 
+(This duplicates a test from [the Chapter 15 tests](lab15-tests.md),
+but modifies the styling because leading works differently in the two
+chapters.)
+
 Here's web page with an iframe inside of it. We make it narrow so that
 resizing is dramatic:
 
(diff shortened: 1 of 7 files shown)


pavpanchekha:
Ok: I found and fixed one important bug. Basically, in each layout object type, we need to clear the descendant dirty bits by calling `self.descendants.set(None)`. However, in `LineLayout`, we were doing this right after the recursive call to `layout`. That seems correct, but actually isn't, because after that recursive call, we change the child's `y` field, which re-sets the `descendant` field, and then that won't be cleared. Big mistake! (This will be difference once I merge with `main` but I think a similar bug will occur.)

pavpanchekha:
Ok, yes, this does work for the reason we thought it would: when you `setAttribute`, you invalidate every frame (this step is important, see #850) and also the `IframeLayout`. Then when you modify the `IframeLayout`, it modifies the `frame_width`, which in turn means that when you get to the other frame it'll invalidate the `DocumentLayout`'s `width`.

src/lab16-tests.md
```diff
@@ -145,3 +145,77 @@ Next, we can execute some JavaScript to change the page:
                  LineLayout(x=13.0, y=38.0, width=774.0, height=20.0, node=<div>)
                    TextLayout(x=13.0, y=38.0, width=80.0, height=20.0, node=..., word=Short)
                    TextLayout(x=109.0, y=38.0, width=64.0, height=20.0, node=..., word=body)
+
+Test iframe resizing
```
chrishtr, line 149: Could you copy these to lab15-tests also?

chrishtr (approved):

chrishtr:
Code and text look good, just needs rebasing on top of my PR to remove `descendants`, and a quick lab15 test.

pavpanchekha:
Done. Apologies, had an urgent thing to finish last week.

PR #853, opened 2023-06-18
Fix URL parsing for url_origin

chrishtr:
For lab15 and lab16, if you tried to run `lab15.py http://browser.engineering`, it would complain that the trailing slash was missing from the URL. This bug is not present in previous chapters.

I fixed it by factoring out URL parsing into a `parse_url` method and used in in all chapters, so that we don't have different code in some places than others (it also aligns better with the commentary in chapter 1 about not reinventing wheels).

I also added some code to avoid crashing on `calc(...` syntax for browser.engineering iframes in chapter 15
(copied from chapter 16).

### DIFF
--- a/book/http.md
+++ b/book/http.md
@@ -250,11 +250,16 @@ url = url[len("http://"):]
 ```
 
 Now we must separate the host from the path. The host comes before the
-first `/`, while the path is that slash and everything after it:
+first `/`, while the path is that slash and everything after it. Let's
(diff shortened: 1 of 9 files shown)


pavpanchekha (approved):
This is not my personally favorite way to factor the code, but we already have a pass on Chapter 1 scheduled for later and this moves us in the right direction.

PR #854, opened 2023-06-19
Analyze dependencies of protected fields

chrishtr:
* Adds a command-line flag to print them out
* Results are in `invalidation_dependencies.txt` (I created this by running with the command-line flag, removing some spurious output lines, and piping the result into `sort -u`
* Uses the result to pre-compute all dependencies in the code, except for `self.children` and style protected fields.

This doesn't remove the need for `read`, because that's important for correctness checking, but now calls to `read` are a form of assert rather than runtime dependency-building. And for each class, we have all of its dependencies listed out in exactly one place. (Except for the case of ascent and descent which are in a slightly different place in one instance,
and `self.children` for `BlockLayout`.)

Any missing dependencies will cause asserts at runtime and can be easily debugged and fixed when added (or the code that adds incorrect dependencies fixed).

Once this PR lands, we can add some text to the chapter about it,  and the corresponding figure derived from `invalidation_dependencies.txt`.


src/lab16.py
```diff
@@ -288,15 +315,29 @@ def __init__(self, node, parent, previous, frame):
...
+        self.height = ProtectedField(self, "height", self.parent)
+        self.x = ProtectedField(self, "x", self.parent)
+        self.y = ProtectedField(self, "y", self.parent)
 
         self.has_dirty_descendants = True
 
+        # self.children has dependencies on the zoom and font style of various
+        # inline children. But we don't know which until self.recurse is
+        # called, and building up the dependencies would amount to a duplicate
+        # implementation of self.recurse.
```
pavpanchekha, line 330: Then do I understand correctly that there are now two modes? Most fields have "frozen" dependencies, but a few don't?

src/lab16.py
```diff
@@ -119,14 +119,20 @@ def paint_visual_effects(node, cmds, rect):
...
+        self.obj = obj
         self.name = name
         self.parent = parent
 
         self.value = None
         self.dirty = True
         self.invalidations = set()
+        self.frozen_dependencies = False
+
+    def set_dependencies(self, dependencies):
```
pavpanchekha, line 132: I'm a little worried about the case where we call `set_dependencies` a second time on a given field. For example, suppose you have a node with `BlockLayout`, and you modify its children. Then won't we end up calling `set_dependencies` twice for the height? I agree this is something we need to do (now we depend on the new children!) but I want to make sure it'll work correctly. Have you thought carefully that all the invalidations will still be right?

pavpanchekha: To clarify: I think it's right, but I don't 100% trust myself for this kind of code.

src/lab16.py
```diff
@@ -340,6 +381,10 @@ def layout(self):
                 self.children.set(self.temp_children)
                 self.temp_children = None
 
+        height_dependencies = [child.height for child in self.children.get()]
```
pavpanchekha, line 384: Ok, I am on board with this general approach but I'm a little unhappy with the `get` call here. In my head, I was always thinking about `get` as an unsafe command and `read` as a safe one. (Because `get` allows you to not set up the dependency.) So I was always auditing my code by looking for each `read` call and making sure they are only in `paint` or similar phases. Probably not fixable.

src/lab16.py
```diff
@@ -288,15 +315,29 @@ def __init__(self, node, parent, previous, frame):
...
+        # called, and building up the dependencies would amount to a duplicate
+        # implementation of self.recurse.
+        self.zoom.set_dependencies([self.parent.zoom])
+        self.width.set_dependencies([self.parent.width])
+        self.x.set_dependencies([self.parent.x])
+
+        previous_dependencies = [self.parent.y]
+        if self.previous:
+            previous_dependencies.append(self.previous.y)
+            previous_dependencies.append(self.previous.height)
```
pavpanchekha, line 338: How about doing something like:
```
self.y.set_dependencies([
  self.parent.y,
  self.previous and self.previous.y,
  self.previous and self.previous.height,
])
```

This is a bit of type punning (we will need `set_dependencies` to skip `False`) but shorter and, I think, clearer. It also avoids naming the temporary.

pavpanchekha: Also, if there is a `previous`, you actually don't directly depend on the parent's `y` position. I don't think this is a _problem_, per se, since if the parent `y` position changes the previous element's `y` position probably will too, but you could imagine adding CSS features where this over-invalidates. Either way, I think we'll need to mention this in the text.

pavpanchekha:
Ok, having read this PR, here's how I understand it:

- This PR aims at explicitly listing all dependencies
- To solve the problem where you don't know all of your children yet, it puts the `set_dependencies` call into the `layout` method. To make sure it's safe, we keep the `read` call.
- To solve the problem that `recurse` is really complicated, `ProtectedField`s actually have two modes: frozen and unfrozen. Frozen fields have an explicit `set_dependencies` call, and then `read` functions as an `assert`. Unfrozen fields don't have an explicit `set_dependencies` call, and its `read` function adds dependencies.

On the one hand, modulo some minor tweaks that I put in the review, I think this PR basically works. It'll require some text changes but that's fine. I like the opportunity to talk about auditability. However, I have two questions:

- Also, if we still have unfrozen fields, what does the explicit dependency list get us? There might still be indirect dependencies that aren't obvious in the text because they flow through one of the unfrozen fields. I suppose the advantage is that we restrict the unknowns to a few fields, but grepping for those fields is hard. Maybe having separate `read_dependency` (frozen) and `read_and_depend` (unfrozen) methods would help, because you can then grep for `read_and_depend` to fill out your mental model of the dependency graph?
- On the topic of auditability, it's not easy to tell which fields are frozen and which are unfrozen. What if we make this two different classes, or a parameter you set in the constructor?

pavpanchekha:
Another question here is how to add this to the chapter. I can see two approaches.

First option: rewrite the whole chapter to talk about the new `ProtectedField`. The challenge here is spacing out the introduction of frozen and unfrozen so it doesn't look like the abstraction is immediately full of holes.

Second option: keep most of the chapter as is, but introduce frozen protected fields at the end, and talk about why you might want that.

Any thoughts? I'm leaning toward option 2, but I'm not really sure.

pavpanchekha:
Also—and maybe this PR isn't the place—given the changes, do we still want to rename `read`? What about `field.read_dependency(source)`? Or `source.read_from(field)` We had talked about `depend_on`, but I think it makes less sense in this PR.

src/lab16.py
chrishtr: Good idea, done.

src/lab16.py
chrishtr: The existing code has the same problem - read will keep adding things to the dependency set over and over. Is this new thing different?

src/lab16.py
pavpanchekha: The `style` fields only depend on their parents, right? (For resolving computed styles and for inheritance.)

src/lab16.py
chrishtr: Updated to not use `get()`

src/lab16.py
pavpanchekha: Absolutely no different. I just want to make sure we definitely have it right!

pavpanchekha: Ok, I guess here's a question I do have: if, instead of doing `self.dependencies.add(...)`, we instead did `self.dependencies = set(...)`, that is, if instead of adding the dependencies we replaced them, would it be correct? Does it matter which one we do?

chrishtr:
> Ok, having read this PR, here's how I understand it:
> 
> * This PR aims at explicitly listing all dependencies

Yes, but mainly as a software engineering pedagogy tool for the chapter, showing that these dependencies allow you to assert things about your code and reason about it as a whole. There's also a secondary performance benefit that we're not calling `add` to frozen sets all the time.

> * To solve the problem where you don't know all of your children yet, it puts the `set_dependencies` call into the `layout` method.

`set_dependencies` is always called in the constructor, except for `height` in one case.

> To make sure it's safe, we keep the `read` call.

Yes. This is also a software engineering lesson: use asserts in the code to ensure invalidation is correct, because it's teh second-hardest thing in computer science.

> * To solve the problem that `recurse` is really complicated, `ProtectedField`s actually have two modes: frozen and unfrozen. Frozen fields have an explicit `set_dependencies` call, and then `read` functions as an `assert`. Unfrozen fields don't have an explicit `set_dependencies` call, and its `read` function adds dependencies.

Yes, and the protected fields in `style`.
 
> On the one hand, modulo some minor tweaks that I put in the review, I think this PR basically works. It'll require some text changes but that's fine. I like the opportunity to talk about auditability. However, I have two questions:
> 
> * Also, if we still have unfrozen fields, what does the explicit dependency list get us?

i think it gets us an opportunity to show how you can write out the dependencies. Which will probably avoid bugs, since if you accidentally start depending on something new you are now forced to think about whether that is ok or not, or whether it's a performance or correctness bug.

There might still be indirect dependencies that aren't obvious in the text because they flow through one of the unfrozen fields. I suppose the advantage is that we restrict the unknowns to a few fields, but grepping for those fields is hard. Maybe having separate `read_dependency` (frozen) and `read_and_depend` (unfrozen) methods would help, because you can then grep for `read_and_depend` to fill out your mental model of the dependency graph?

I think either freezing or not a whole `ProtectedField` is easier to think about, as a form of more fully locking down the code.

> * On the topic of auditability, it's not easy to tell which fields are frozen and which are unfrozen. What if we make this two different classes, or a parameter you set in the constructor?

Sure, good idea!

chrishtr:
> Another question here is how to add this to the chapter. I can see two approaches.
> 
> First option: rewrite the whole chapter to talk about the new `ProtectedField`. The challenge here is spacing out the introduction of frozen and unfrozen so it doesn't look like the abstraction is immediately full of holes.

Right, we could try that.

> 
> Second option: keep most of the chapter as is, but introduce frozen protected fields at the end, and talk about why you might want that.

> Any thoughts? I'm leaning toward option 2, but I'm not really sure.

Option 2 sounds better to me. It makes for another step towards practical techniques for program correctness assertions, and makes sense as a thing to do on top of the other work. Understanding all this stuff is often a bootstrapping exercise, and jumping to the answer and showing them it's correct feels like too many steps at once.

chrishtr:
> Also—and maybe this PR isn't the place—given the changes, do we still want to rename `read`? What about `field.read_dependency(source)`? Or `source.read_from(field)` We had talked about `depend_on`, but I think it makes less sense in this PR.

I do want to rename `read`, but in a followup PR.

src/lab16.py
chrishtr: > Ok, I guess here's a question I do have: if, instead of doing `self.dependencies.add(...)`, we instead did `self.dependencies = set(...)`, that is, if instead of adding the dependencies we replaced them, would it be correct? Does it matter which one we do?

We could do that as long as all of the dependencies were known at the same time. The current code doesn't always do that.

chrishtr: FWIW I just changed the `height` dependencies to be inside the dirty bits, which at least avoids recomputing these dependencies on every layout.

pavpanchekha:
Ok, I think then we've basically converged on what to do here:

- Modify this so ProtectedField are frozen or not when they are constructed, then merge.
- Describe frozen fields in a final section in the chapter. (Boy am I glad we left a thousand words in the tank for this and the framing section!) Talk about why one would want to do this.
- Rename `read`, name still TBD, in another PR.

chrishtr:
> * Modify this so ProtectedField are frozen or not when they are constructed, then merge.

I changed to pass dependencies in the constructor, except for height, ascent and descent. We could either leave that or
delete the code that computes those dependencies in another PR...

PR #856, opened 2023-06-23
First pass on a new section about frozen dependencies

pavpanchekha:
This PR adds a section about "frozen" `ProtectedField`s. Honestly I was at a bit of a loss writing this section—I'm not sure it makes a good end to the chapter, because I don't make a strong enough case for why one would want this. Perhaps, @chrishtr, you should take over this PR and add some words from your experience in Blink.

As I was writing, I made a few changes:

- I removed the feature where a dependency can be `False` and then it is skipped. I just wrote a slightly nicer `if` statement and it looks better without that type punning
- I also removed `set_dependencies`, and just left `height`, `ascent`, and `descent` unfrozen. I could really find a good way to justify having three states (frozen in constructor, frozen during layout, and unfrozen), and it felt better to just weaken the guarantees about `height`, `ascent`, and `descent`.

I can go either way on either of these, I just tried to do what was natural given the text I was writing. Feedback and/or takeovers of the PR welcome.

### DIFF
--- a/book/invalidation.md
+++ b/book/invalidation.md
@@ -2323,6 +2323,124 @@ invalidate any layout fields (because none of those fields depend on
 these properties) and so animations will once again skip layout
 entirely.
 
+::: {.further}
+TODO
+:::
+
+
+Making dependencies visible
+===========================
+
+We now have dirty bits to skip unnecessarily rebuilding the layout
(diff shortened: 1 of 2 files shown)


chrishtr:
Going to merge this PR and then build the rest of what we agreed in a followon one.

PR #861, opened 2023-07-01
Don't use asserts in CSSParser

chrishtr:
Instead use exceptions. This allow us to use the `-O` commandline parameter to remove asserts and test 
performance of the browser.

Also adds a helper method to `MeasureTime` to print out a snapshot of time after particular points of interest
(e.g. it could be used for after layout)).

### DIFF
--- a/book/styles.md
+++ b/book/styles.md
@@ -78,8 +78,8 @@ it started and extracts the substring it moved through.
     and colors (which use the hash sign). Real CSS values have a more
     complex syntax but this is enough for our toy browser.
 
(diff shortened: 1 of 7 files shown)


chrishtr:
The browser renders about 10% faster in lab16.py when `-O` is used.

pavpanchekha (approved):

pavpanchekha:
Shouldn't we update the chapter?

chrishtr:
Updated Chapter 6's text, plus the code for lab16 and lab13.

PR #864, opened 2023-07-09
Fix several lab16 bugs, and two lab15 ones.

chrishtr:
lab16:

* Invalidate style on focus change
* null-check layout objects
* frame, not tab

lab15 + lab16:

* url is on `Frame`, not `Tab`
* scope some JS code to `window`


chrishtr:
(merging, can revise after if you see a problem)

pavpanchekha:
Lgtm

PR #868, opened 2023-07-12
More bug fixes for lab15 and lab15

chrishtr:
* lab15: wrong arguments when resetting zoom
* lab16: need `x.get()` and `y.get()` in `click`
* Height depends on width (because of aspect ratio for `LayoutEmbed` subclasses)


pavpanchekha (commented):
Looks good but we probably need text changes for the `height` depends on `width` thing.

chrishtr:
> Looks good but we probably need text changes for the `height` depends on `width` thing.

Done.

pavpanchekha (approved):

PR #871, opened 2023-07-22
Update "what wasn't covered" to include more topics

chrishtr:
Also, rewrote and updated some of the existing content.

### DIFF
--- a/book/skipped.md
+++ b/book/skipped.md
@@ -1,16 +1,16 @@
 ---
 title: What Wasn't Covered
 type: Conclusion
-prev: visual-effects
+prev: invalidation
 next: change 
 ...
 
-The last eleven chapters have, I hope, given you a solid understanding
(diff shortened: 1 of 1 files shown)


pavpanchekha (approved):
Wish we could actually suggest a book every time we say to read one. But for most of these topics I don’t have recommendations, though. Do you?

You figure we just shouldn’t order these?

chrishtr:
> Wish we could actually suggest a book every time we say to read one. But for most of these topics I don’t have recommendations, though. Do you?

I don't have any recommendations either. And prefer not to pick winners..
 
> You figure we just shouldn’t order these?

Yeah. My reasoning is that I don't want anyone who works on those topics to think they are being disrespected. Also I have no particular reason to think that any of them are more important than others -- they are all necessary for a real browser.

PR #876, opened 2023-07-31
Fix inlined versions of chapters 1-16

pavpanchekha:
This PR adds missing imports in a variety of chapters to make sure the `labXX.full.py` versions work correctly. I didn't test them super hard, but they do all render the book page roughly correctly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ widgets: \
 	www/widgets/lab12-browser.html www/widgets/lab12.js \
 	www/widgets/lab13-browser.html www/widgets/lab13.js
 
-src/lab%.full.py: src/lab%.py
+src/lab%.full.py: src/lab%.py infra/inline.py infra/asttools.py
 	python3 infra/inline.py $< > $@
 
 CHAPTER=all
--- a/infra/asttools.py
+++ b/infra/asttools.py
@@ -158,13 +158,6 @@ class ResolvePatches(ast.NodeTransformer):
         self.visit(tree)
         return (self.visit(tree), self.patches)
 
-def has_js_hide(decorator_list):
-    return any([
-        isinstance(dec, ast.Attribute) and dec.attr == "js_hide"
-        and isinstance(dec.value, ast.Name) and dec.value.id == "wbetools"
-        for dec in decorator_list
-    ])
-
 class ResolveJSHide(ast.NodeTransformer):
     def visit_FunctionDef(self, cmd):
         if any([
--- a/infra/asttools.py
+++ b/infra/asttools.py
@@ -113,8 +113,9 @@ class ResolvePatches(ast.NodeTransformer):
         if not cmd.decorator_list:
             patches = self.patches.get(cmd.name, [])
             if patches:
+                args2 = patches[-1].args
                 body2 = patches[-1].body
-                return ast.FunctionDef(cmd.name, cmd.args, body2, [])
+                return ast.FunctionDef(cmd.name, args2, body2, [])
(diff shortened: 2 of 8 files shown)


infra/asttools.py
```diff
@@ -152,6 +158,23 @@ def double_visit_and_patches(self, tree):
...
+def has_js_hide(decorator_list):
+    return any([
+        isinstance(dec, ast.Attribute) and dec.attr == "js_hide"
+        and isinstance(dec.value, ast.Name) and dec.value.id == "wbetools"
+        for dec in decorator_list
+    ])
+
+class ResolveJSHide(ast.NodeTransformer):
+    def visit_FunctionDef(self, cmd):
+        if any([
```
chrishtr, line 170: This code duplicates `has_js_hide`. Did you mean to call it?

Makefile
```diff
@@ -29,7 +29,7 @@ widgets: \
 	www/widgets/lab12-browser.html www/widgets/lab12.js \
 	www/widgets/lab13-browser.html www/widgets/lab13.js
 
-src/lab%.full.py: src/lab%.py
+src/lab%.full.py: src/lab%.py infra/inline.py infra/asttools.py
```
chrishtr, line 32: Does this change cause `make test` to fail if the behavior regresses in the future?

src/lab15.py
```diff
@@ -17,29 +17,35 @@
...
 import urllib.parse
 import wbetools
+import OpenGL.GL
 
 from lab1 import parse_url
+from lab2 import WIDTH, HEIGHT, HSTEP, VSTEP, SCROLL_STEP
 from lab4 import print_tree
-from lab13 import Text, Element
+from lab5 import BLOCK_ELEMENTS
+from lab14 import Text, Element
```
chrishtr, line 26: This is out of order, is it necessary for patching reasons?

infra/asttools.py
```diff
@@ -76,11 +76,16 @@ def iter_methods(cls):
...
 def load(filename):
     if filename not in inline_cache:
+        inline_cache[filename] = "in-progress"
         with open(filename) as f:
             ast = parse(f.read(), filename)
         ast = inline(ast)
         inline_cache[filename] = ast
-    return inline_cache[filename]
+        return ast
+    elif inline_cache[filename] == "in-progress":
```
chrishtr, line 85: Is it also easy to check for unused imports?

chrishtr (commented):
Can you check if the browser widgets still compile and work?

Makefile
pavpanchekha: No, just helps work on the inliner. It would be good to auto test the inlined version but I don’t do that here.

src/lab15.py
pavpanchekha: No, just matching the existing order.

infra/asttools.py
pavpanchekha: Huh. I don’t think so, since this is only looking at imports, not uses.

infra/asttools.py
pavpanchekha: Well, it would be circular. Maybe it would be good to call in the other direction but I thought that was maybe complicated. Can fix.

chrishtr: Nothing seems to call `has_js__hide`...

pavpanchekha: Oh, d'oh! Yes, I copied it over but then inlined it into `ResolveJSHide`. Removed the `has_js_hide` method..

PR #877, opened 2023-07-31
Paint the chrome before rastering it

pavpanchekha:
This PR changes how the browser chrome is drawn. Currently, the browser chrome is rastered directly, without an intermediate display list. This PR makes it create an intermediate display list first, made up of normal display commands, and then separately raster it.

The benefit of this is that the transition to Skia is much simpler, because we don't need separate `draw_line`, `draw_text`, and `draw_rect` commands, plus their (odd) calling conventions. (I think they also had a bug?) This makes the drawing command abstraction clearer and tighter.

The cost—the reason why we didn't do it earlier—is that we now need to define `DrawLine` and `DrawOutline` earlier. But we ended up needing them anyway, so I think it's not a big cost. I also changed the back button to draw a `<` character instead of drawing a polyline of that shape.

### DIFF
--- a/book/chrome.md
+++ b/book/chrome.md
@@ -685,8 +685,6 @@ class Tab:
 To better separate the chrome from the page, let's also add a border:
 
 ``` {.python}
-CHROME_PX = 100
-
 class Browser:
     def paint_chrome(self):
         # ...
--- a/book/chrome.md
+++ b/book/chrome.md
@@ -622,82 +622,28 @@ happen in the `Browser` class.
 [^chrome]: Yep, that predates and inspired the name of Google's Chrome
     browser.
 
-Since we're interested in multiple tabs, let's add some code to draw a
-tab bar at the top of the browser window---and let's keep it simple,
-because this is going to require some tedious and mildly tricky
-geometry.
-
-We'll draw the tabs in the `draw` method, after the page contents are
-drawn:
+Much like tabs, the browser chrome is going to generate a display list
+and then draw it to the canvas. However, unlike tabs, this display
+list will always be drawn at the top of the window and won't be
+scrolled:
 
 ``` {.python}
 class Browser:
     def draw(self):
         # ...
-        tabfont = get_font(20, "normal", "roman")
-        for i, tab in enumerate(self.tabs):
-            # ...
+        for cmd in self.paint_chrome():
+            cmd.execute(0, self.canvas)
 ```
 
-Python's `enumerate` function lets you iterate over both the indices
-and the contents of an array at the same time. Let's make each tab 80
-pixels wide and 40 pixels tall. We'll label each tab something like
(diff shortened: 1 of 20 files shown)


pavpanchekha:
It's done. (Working on final tests now.) In the process of doing this I made a few bigger changes:

- I ended up changing up the order of chrome drawing in Chapter 7 quite a bit, since we now need new drawing commands and also a new `paint_chrome` method. But I think the new text is still good.
- I changed the constructor for `DrawOutline` take x/y/x/y instead of a `Rect`, so that it is the same as `DrawRect`. It's a bit weird when we switch to Skia later but it seemed better not to mess with it.
- In Chapter 11, the section on using Skia as the canvas got a lot shorter, so I moved in the related text/font stuff into that section. That left very little in the next section, but I guess for now I kept it as a breather. I also switched around "go further" blocks to better fit the content.
- In Chapter 14, doing all-black outlines first and then adding colored outlines is a bit less impactful now, since the underlying `DrawOutline` supports them from the start (which I did so it is similar to the other `Draw` commands), but I think it basically still works.

book/chrome.md
```diff
@@ -622,29 +622,120 @@ happen in the `Browser` class.
...
+class Tab:
+    def scrolldown(self):
+        max_y = self.document.height - (HEIGHT - CHROME_PX)
+        self.scroll = min(self.scroll + SCROLL_STEP, max_y)
+```
+
+To better separate the chrome from the page, let's also add a border:
+
+``` {.python}
+CHROME_PX = 100
```
chrishtr, line 688: remove this as redundant given line 649 above

book/visual-effects.md
```diff
@@ -289,218 +289,92 @@ use; modern browsers support [quite a lot][css-colors].
...
 
-Next, the plus button for adding a new tab:[^move-plus]
-
-[^move-plus]: I also changed the *y* position of the plus sign.
-    Skia draws fonts a bit differently from Tkinter, and the new *y*
-    position keeps the plus centered in the box. Feel free to adjust
-    the positions of the UI elements to make everything look good on
-    your system.
+To draw just the outline, set the `Style` parameter of the `Paint` to
+`Stroke_Style`. Here "stroke" is a standard term refering to drawing
```
chrishtr, line 337:
```suggestion
`Stroke_Style`. Here "stroke" is a standard term referring to drawing
```

chrishtr (commented):
I assume you also tested all the chapters and they drew correctly? Do the widgets look ok?

pavpanchekha:
The chapters do all draw correctly.

pavpanchekha:
Ok, I checked the widgets and found two bugs. (Also, the fonts in the widgets are misaligned—I think that was already there—and there's some weird rendering behavior in lab11 which I also think was already there.

PR #879, opened 2023-08-01
Group URL functions into a URL class

pavpanchekha:
This PR organizes `parse_url`, `request`, `resolve_url`, and `url_origin` into a new class called `URL`. The constructor is basically `parse_url`, and `request`, `resolve`, and `origin` are methods on it. This is a little neater organization-wise. It's also more testable and more similar to most URL handling libraries.

The code, tests, and text for each chapter have all been updated.

The largest changes involved:
- I reorganized Chapter 1 a little bit to move the example of the URL up, to make it more clearly the focus of the chapter. I also had to make other changes to the text to better match the new organization of all of the code inside a `URL` class.
- Almost everywhere where we used to use a string URL—in the `Tab`'s `url` field, in the `load` arguments, inside the `history`—that has now become a `URL` object.
- When creating `iframe`s, we were resolving the iframe `src` attribute against the root frame's URL, which isn't right; it should be resolved against the containing frame. That's been fixed.
- I had to fix a quite subtle bug in the JS compiler's outline loading mechanism to get JS compilation to go through.


book/http.md
```diff
@@ -214,61 +232,75 @@ So far we've communicated with another computer using `telnet`. But it
...
+        if "/" not in url:
+            url = url + "/"
+        self.host, url = url.split("/", 1)
+        self.path = "/" + url
 ```
 
-The `split(s, n)` method splits a string at the first `n` copies of
-`s`. The path is supposed to include the separating slash, so I make
-sure to add it back after splitting on it.
+When you see a code block with a `# ...`, like this one, that you're
```
chrishtr, line 280:
```suggestion
(When you see a code block with a `# ...`, like this one, that you're
```

book/http.md
```diff
@@ -214,61 +232,75 @@ So far we've communicated with another computer using `telnet`. But it
...
+adding code to an existing method or block. The `split(s, n)` method
```
chrishtr, line 281:
```suggestion
adding code to an existing method or block.) The `split(s, n)` method
```

book/http.md
```diff
@@ -214,61 +232,75 @@ So far we've communicated with another computer using `telnet`. But it
...
+splits a string at the first `n` copies of `s`. Note that there's some
+tricky logic here for handling the slash between the host name and the
+path. That (optional) slash is part of the path.
 
-With the host and path identified, the next step is to connect to the
-host. The operating system provides a feature called "sockets" for
-this. When you want to talk to other computers (either to tell them
+Our browser will create a `URL` objects based on user input, and then
```
chrishtr, line 286:
```suggestion
Our browser will create a `URL` object based on user input, and then
```

book/http.md
```diff
@@ -332,27 +370,31 @@ Request and response
...
+               .encode("utf8"))
 ```
 
-There are a few things to be careful of here. First, it's very
-important to use `\r\n` instead of `\n` for newlines. It's also
-essential that you put *two* newlines `\r\n` at the end, so that you
-send that blank line at the end of the request. If you forget that,
-the other computer will keep waiting on you to send that newline, and
-you'll keep waiting on its response.[^literal]
+There are a few things to here that have to be exactly right. First,
```
chrishtr, line 382:
```suggestion
There are a few things to note here that have to be exactly right. First,
```

book/http.md
```diff
@@ -370,11 +412,11 @@ different types to text and to bytes:
 If you see an error about `str` versus `bytes`, it's because you
 forgot to call `encode` or `decode` somewhere.
 
-If you run this, you'll notice that the `send` call returns a number,
-in this case `47`. That tells you how many bytes of data you sent to
-the other computer; if, say, your network connection failed midway
-through sending the data, you might want to know how much you sent
-before the connection failed.
+If you run this in the REPL, you'll notice that the `send` call
```
chrishtr, line 415: REPL?

book/http.md
```diff
@@ -610,23 +667,50 @@ Let's try to take this code and add it to `request`. First, we need to
...
-    "Unknown scheme {}".format(scheme)
+class URL:
+    def __init__(self, url):
+        self.scheme, url = url.split("://", 1)
+        assert self.scheme in ["http", "https"], \
+            "Unknown scheme {}".format(self.scheme)
+        # ...
 ```
 
+Note that here you're supposed to replace the existing scheme parsing
```
chrishtr, line 678:
```suggestion
(Note that here you're supposed to replace the existing scheme parsing
```

book/http.md
```diff
@@ -610,23 +667,50 @@ Let's try to take this code and add it to `request`. First, we need to
...
+code with this new code. It's usually clear from context and the code
+itself what you need to replace.
```
chrishtr, line 680:
```suggestion
itself what you need to replace.)
```

book/styles.md
```diff
@@ -606,25 +606,22 @@ To download the style sheets, we'll need to convert each relative URL
...
-            if dir.count("/") == 2: continue
-            dir, _ = dir.rsplit("/", 1)
-        return dir + "/" + url
+class URL:
+    def resolve(self, url):
+        if "://" in url: return URL(url)
+        if not url.startswith("/"):
+            dir, _ = self.path.rsplit("/", 1)
+            while url.startswith("../"):
+                _, url = url.split("/", 1)
```
chrishtr, line 615: Why change the logic of the while loop?

book/http.md
pavpanchekha: Initially the conceit was you'd run all of this code in a Python REPL (that is, in an interpreter on the command line) but now that conceit is gone and this sentence is kind of homeless. I think probably this whole paragraph needs to become a footnote now.

book/styles.md
pavpanchekha: Oh yeah good question. Basically, because before the string contained the hostname too (because it started as the full URL, and then I removed the scheme) but now it is just the path (because it's already parsed). But I also changed some small stuff, like using `url.split("/", 1)` instead of `url[3:]` to strip off the `../`

book/http.md
chrishtr: I didn't see REPL defined earlier in the chapter, so I don' think readers will understand. Just say "run this via Python on the command line"?

chrishtr: Ok. Just make sure it's well-tested...

chrishtr (approved):

PR #883, opened 2023-08-03
`text` to `word`

pavpanchekha:
This PR changes how text layout works a little bit. Currently, a whole `Text` node is passed a method called `text`, which splits it into words and adds each of those words to the current line. This PR changes it to a method called `word`, and moves the splitting and looping into the caller. The main advantage is that `word` is now similar in scope to other methods like `input`, `iframe`, `image`, and so on. The changes required are quite modest.

### DIFF
--- a/book/embeds.md
+++ b/book/embeds.md
@@ -447,15 +447,14 @@ class BlockLayout:
         self.cursor_x += w + font(node.style, self.zoom).measureText(" ")
 ```
 
-We can redefine  `text` and `input` in a satisfying way now:
+We can redefine  `word` and `input` in a satisfying way now:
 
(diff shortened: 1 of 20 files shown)


book/embeds.md
```diff
@@ -447,15 +447,14 @@ class BlockLayout:
         self.cursor_x += w + font(node.style, self.zoom).measureText(" ")
 ```
 
-We can redefine  `text` and `input` in a satisfying way now:
+We can redefine  `word` and `input` in a satisfying way now:
```
chrishtr, line 450:
```suggestion
We can redefine `word` and `input` in a satisfying way now:
```

chrishtr (approved):

PR #884, opened 2023-08-05
Draw text cursor in `InputLayout.paint`, not `Tab.render`

pavpanchekha:
This PR makes cursor drawing the `InputLayout`'s responsibility. The downside of this approach is that it requires each node to know whether or not it is focused, but it turns out we need that later anyway, and the new approach avoids some graphics glitches when an `<input>` element is `transform`ed. In total I think it also makes the text a bit simpler.


pavpanchekha:
I will note that I tested this (hence the random bugfixes)

book/accessibility.md
```diff
@@ -1030,23 +1030,18 @@ element they want or if they need to keep hitting `Tab`. In most
...
-implement text cursors. Recall that, right now, text cursors are added
-by drawing a vertical line in the `Tab`'s `render` method. We could
-extend that code to draw a cursor or an outline, but before we make
-that method more complicated let's move it into the `InputLayout` so
-we have easier access to size and position information.[^effects] To
-do that, we'll need each `InputLayout` to know whether or not it is
-currently focused:
+To implement focus rings, we'll use the same mechanism we use to draw
+text cursors. Recall that, right now, text cursors are added by
+drawing a vertical line in the `InputLayout`'s `paint` method. We'll
```
chrishtr, line 1035:
```suggestion
drawing a vertical line in `InputLayout`'s `paint` method. We'll
```

book/accessibility.md
```diff
@@ -1030,23 +1030,18 @@ element they want or if they need to keep hitting `Tab`. In most
...
+add a call to `paint_outline` in that method:
```
chrishtr, line 1036:
```suggestion
add a call to `paint_outline` in that method, to draw a rectangle around the focused
element:
```

book/accessibility.md
```diff
@@ -1064,33 +1059,23 @@ class Tab:
 
 To draw an outline, we'll use `DrawOutline`:
```
chrishtr, line 1060: I don't see DrawOutline in the code below any more, this line needs to be adjusted.

chrishtr (commented):
Don't you need to edit lab15.py also?

book/accessibility.md
pavpanchekha: Yes, because we now introduce it in Ch 7 to draw the browser chrome!

pavpanchekha:
Kind of a funny story about lab15.py but ultimately you are right, it is better to update it.

chrishtr (approved):

PR #886, opened 2023-08-07
Create latex / pdf output for each chapter

chrishtr:
Chapter footnotes and code previews work.

Still TODO:

* Making some special characters (e.g. Chinese) render correctly
* Finding a way to render all chapters into one tex file, so we can automatically generate a table of contents and index

Debian Linux packages I've installed to make this work:
`texlive`
`texlive-base`
`texlive-latex-extra`
`texlive-lang-cjk`


pavpanchekha (approved):
That's a pretty gnarly template! But this is a great start, let's merge it!

PR #890, opened 2023-08-09
Generate all chapters into one book output

chrishtr:
This PR creates a latex file for the whole book. It works as follows:

1. Generates a simple sub-template for each chapter named `latex/<chapter name>-chapter.tex`
2. Generates a skeleton overall latex output for the book named `latex/book-skeleton.tex`. This is done by running all of the chapters together through pandoc with a latex output mode, and then throwing away the actual chapter content. This has the effect of including all definitions and packages from `template-book.tex` that are necessary to render any of the chapters
3. Inserts the sub-templates from the chapters into the output of step 2, resulting in `book.tex` for the entire book
4. Renders `book.tex` into a PDF

Also adds support for Chinese characters to the template, and removes one emoji character that is hard to render.

I installed the `texlive-lang-cjk` Debian package, which contains CJK fonts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,23 +41,23 @@ PANDOC=pandoc --from markdown --to html --lua-filter=infra/filter.lua --fail-if-
 
 PANDOC_LATEX=pandoc --standalone --from markdown --to latex --fail-if-warnings --metadata-file=config.json $(FLAGS)
 
+# Generates a simple chapter latex rendering meant to be inserted into the larger book skeleton.
 latex-chapters: $(patsubst %,latex/%-chapter.tex,$(CHAPTERS))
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -2,13 +2,15 @@
 
 FLAGS=
 
-CHAPTERS=$(patsubst book/%.md,%,$(wildcard book/*.md))
+CHAPTERS=about preface intro history http graphics text html \
```
chrishtr, line 5: This ensures the chapters are in the correct order, which is necessary for latex output.

chrishtr:
>  leave the main file not-generated, just a static file that lists all the chapters and everything else.

How would that work? A book needs a single tex file somehow, right?

I'll merge for now but we can always revisit later.

PR #893, opened 2023-08-10
Fix issues reported by readers at /feedback

chrishtr:


book/http.md
```diff
@@ -783,6 +783,10 @@ should look something like this:
 Exercises
 =========
 
+*Alternate encodings:* add support for a non-`utf8` value for
+`Content-Type`. Test it on a real site such as `google.com`,
+which doesn't use `utf8`).
```
pavpanchekha, line 788:
```suggestion
`Content-Type`. Test it on a real site such as `google.com`
(which doesn't use `utf8`).
```

book/styles.md
```diff
@@ -350,7 +350,8 @@ class DescendantSelector:
         self.descendant = descendant
 ```
 
-Then the `match` method is recursive:
+Then the `match` method is recursive (recursion bottoms out at a
+`TagSelector`):
```
pavpanchekha, line 354:
```suggestion
Then the `match` method is recursive (and bottoms out at, say, a
`TagSelector`):
```

pavpanchekha (approved):
Great, thanks! I archived the actual feedback items.

PR #895, opened 2023-08-12
Create an index and add some of the glossary entries to it

chrishtr:
The latex commands are ignored by pandoc when compiling to HTML.

I only looked at glossary entries referenced in chapter 1/history for now, the rest will be in subsequent PRs.


pavpanchekha (approved):
Later on we can write some pandoc scripts to make this easier.

PR #905, opened 2023-08-23
Fix clipping of composited outlines

chrishtr:


pavpanchekha:
Not the prettiest solution but I guess it's fine. It fixes the bug!

PR #906, opened 2023-08-23
Deduplicate and delete some methods

pavpanchekha:
This PR fixes some minor issues:

1. Sometimes we imported a function twice, and then it ended up in the outline twice
2. We had an unused method `announce_text`, which was an obsolete duplicate of `AccessibilityNode.text`
3. Removed the `is_focused` method, which I tested and determined was unneeded

### DIFF
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -1247,7 +1247,7 @@ class PseudoclassSelector:
         if not self.base.matches(node):
             return False
         if self.pseudoclass == "focus":
-            return is_focused(node)
+            return node.is_focused
         else:
             return False
 ```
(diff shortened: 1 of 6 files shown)


chrishtr (approved):

chrishtr:
Are you still working on this PR?

pavpanchekha:
Sorry for the confusion. Yes—in fact I think I will merge this in a second when the tests pass. Basically, there was an error about an "unnecessary hint" but it seemed strange to me that this hint was unnecessary. It should be used! So I dug into it and in fact I had made a fairly elementary error with the handling of patches in the compiler. The error lead to the compiler skipping patches which made it not use that hint.

I fixed the error and now all the unnecessary hints are necessary again.

PR #910, opened 2023-08-27
Fix widgets broken by recent refactoring

chrishtr:
* Need to pass a `URL` argument to `load`
* Special code needed rename `Layout` to `BlockLayout`
* Fix `toString` to work when writing the URL in the browser chrome
* Special-case `URL` as a class name that happens to be all caps
* Re-export patched classes
* Fix `outset` on rects
* Must import objects from the JS module that actually declared them (hence changes to `DocumentLayout` imports)


pavpanchekha (approved):
Great catches!

PR #915, opened 2023-09-13
Make `MeasureTime` output trace files

pavpanchekha:
This PR changes `MeasureTime` to create JSON trace files that can be used in chrome://tracing or Perfetto or the Firefox Profiler. It's been tested and works well enough—certainly better than the old print-only tracer.

For now, only Chapter 12 is updated; before merging, we need to:
- Update Chapters 13–15 to use the new `MeasureTime` API
- See how traces can be used in Chapter 16 to clarify performance impact


book/scheduling.md
```diff
@@ -974,100 +974,164 @@ this question, let's instrument the browser and measure how much time
...
+``` {.python}
 class Browser:
     def handle_quit(self):
-        print(self.measure_raster_and_draw.text())
+        self.measure.close()
 ```
 
-(Naturally you'll need to call these methods before quitting, from the main
-event loop, so it has a chance to print its timing data.)
+Naturally you'll need to call this methods before quitting, from the main
```
chrishtr, line 1111:
```suggestion
Naturally you'll need to call this method before quitting, from the main
```

book/scheduling.md
```diff
@@ -1428,6 +1492,71 @@ would improve as well.
...
+::: {.todo}
+Trace
+:::
+
+You can see how the render and raster tasks now happen on different
+threads, and how our multi-threaded architecture allows them to happen
+concurrently.
+
+::: {.further}
+
```
chrishtr, line 1555: I can write something for this if you want, about how important tracing has been for Chrome's performance optimization work over the years.

src/lab12.py
```diff
@@ -406,7 +421,7 @@ def __init__(self):
         self.url = None
         self.scroll = 0
 
-        self.measure_raster_and_draw = MeasureTime("raster-and-draw")
+        self.measure = MeasureTime("browser.trace")
```
chrishtr, line 424: Put this behind a flag, so it doesn't constantly create files unless we wan them.

book/scheduling.md
```diff
@@ -974,100 +974,164 @@ this question, let's instrument the browser and measure how much time
...
+trace event conveniently means every later trace event will need to be
+preceded by a comma. So let's add those later trace events. We
+specifically want `B` and `E` events, which mark the beginning and end
+of some interesting computation:
+
+``` {.python replace=1%7d/%27%20%2b%20str(tid)%20%2b%20%27%7d}
+class MeasureTime:
+    def start(self, name):
+        ts = time.time() * 1000000
+        self.file.write(
```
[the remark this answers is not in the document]
pavpanchekha: That's a good question. The benefit of doing it manually is we can stream out JSON, which chrome://tracing and Perfetto can handle (not Firefox). This also means that if you kill the browser you'll still get stuff written out.

book/scheduling.md
pavpanchekha: That would be really great.

src/lab12.py
pavpanchekha: You mean in our impl, not in the book, right? Can do, no problem.

pavpanchekha:
All chapters now use the new API. Still worth looking at whether we can use the trace infrastructure more in the book, but perhaps we should merge this PR first and then put that in a future PR.

pavpanchekha:
Ok, I'll need to rename `start` and `close` so those aren't also shared with library code.

book/scheduling.md
```diff
@@ -974,100 +974,167 @@ this question, let's instrument the browser and measure how much time
...
+Here, we write a trace event with a `ph` of `M` and a `name` of
+`process_name`, which means this record changes the process name
+displayed by the tracing tool (passed in the `args`). Since our
+browser only has one process, I just pass `1` for the process ID, and
+the `cat`egory has to be `__metadata` for metadata trace events. The
+`ts` field stores a timestamp; since this is the first event, it'll
+set the start time for the whole trace, so it's important to put in
+the actual current time.
+
+Updating the process name isn't that important, but writing an initial
```
chrishtr, line 1019:
```suggestion
Updating the process name isn't that important^[It would be important only for
a multi-process browser!], but writing an initial
```

book/scheduling.md
```diff
@@ -974,100 +974,167 @@ this question, let's instrument the browser and measure how much time
...
+        self.file.write(
+            ', { "ph": "E", "cat": "_",' +
+            '"name": "' + name + '",' +
+            '"ts": ' + str(ts) + ',' +
+            '"pid": 1, "tid": 1}')
+        self.file.flush()
 ```
 
-Let's measure the total time for render:
+Finally, when we shut down the browser, we want to leave the file a
```
chrishtr, line 1053:
```suggestion
Finally, when finishing tracing (e.g. when closing the browser), leave the file a
```

book/scheduling.md
```diff
@@ -974,100 +974,167 @@ this question, let's instrument the browser and measure how much time
...
+Note that I call `flush` every time I write to the file. This makes
+sure that if the browser crashes, all of the log events---which might
+help me debug---are already safely on disk.[^invalid-json]
 
+[^invalid-json]: Some of the tracing tools listed above actually
+accept invalid JSON files, in case the trace comes from a browser
+crash.
+
+Now we can use `MeasureTime` to measure how long various browser
+components take. We create it in the `Browser`:
```
chrishtr, line 1072:
```suggestion
components take. Create it in the `Browser`:
```

book/scheduling.md
```diff
@@ -974,100 +974,167 @@ this question, let's instrument the browser and measure how much time
...
-    def handle_quit(self):
-        print(self.tab.measure_render.text())
+    def render(self):
+        if not self.needs_render: return
+        self.browser.measure.start('render')
+        # ...
+        self.browser.measure.stop('render')
+```
+
+We close out the trace file when the browser stops:
```
chrishtr, line 1106:
```suggestion
And close out the trace file when the browser stops:
```

book/scheduling.md
```diff
@@ -974,100 +974,167 @@ this question, let's instrument the browser and measure how much time
...
 class Browser:
     def handle_quit(self):
-        print(self.measure_raster_and_draw.text())
+        self.measure.close()
 ```
 
-(Naturally you'll need to call these methods before quitting, from the main
-event loop, so it has a chance to print its timing data.)
+Naturally you'll need to call this method before quitting, from the
+main event loop, so it has a chance to print its timing data.
```
chrishtr, line 1115:
```suggestion
main event loop, so it has a chance to close out the JSON.
```

book/scheduling.md
```diff
@@ -974,100 +974,167 @@ this question, let's instrument the browser and measure how much time
...
-    Time in raster-and-draw on average: 66ms
-    Time in render on average: 20ms
+::: {.todo}
+Need screenshots
+:::
 
-On every animation frame, my browser spent about 20ms in `render` and
-about 66ms in `raster_and_draw`. That clearly blows through our 16ms
-budget. So, what can we do?
+In Chrome tracing, you can choosing the cursor icon from the toolbar
```
chrishtr, line 1126:
```suggestion
In Chrome tracing, you can choose the cursor icon from the toolbar
```

book/scheduling.md
```diff
@@ -1428,6 +1495,71 @@ would improve as well.
...
+Next, in every trace event, we'll want to provide a real thread ID in
+the `tid` field, which we can get by calling `get_ident` from the
+`threading` library:
+
+``` {.python}
+class MeasureTime:
+    def start(self, name):
+        # ...
+        tid = threading.get_ident()
+        self.file.write(
```
chrishtr, line 1537: You should be able to name the thread with one of the fields available.

book/scheduling.md
```diff
@@ -1428,6 +1495,71 @@ would improve as well.
...
+Trace
+:::
+
+You can see how the render and raster tasks now happen on different
+threads, and how our multi-threaded architecture allows them to happen
+concurrently.
+
+::: {.further}
+
+:::
```
chrishtr, line 1559:
```suggestion
The tracing system we introduced in this chapter comes directly from real browsers. And it's
used every day by browser engineers to understand the performance characteristics of
the browser in different situations, find bottlenecks, and fix them. Without these tools,
browsers would not have been able to make many of the performance leaps they did in
recent years. Good debugging tools are essential to software engineering!

:::
```

pavpanchekha:
Hi Chris, I fixed the things you pointed out, and the issue the CI didn't like, but now it's upset that it doesn't know how to write files in JS. I'll work on that when I can.

book/scheduling.md
chrishtr: ?

chrishtr (commented):
One more open comment thread still.

Also, did you test all of the lab*-browser.html pages?

pavpanchekha:
Good news: Python threads have `name` fields which we can set. I'll use that.

pavpanchekha:
Threads named, widgets tested, everything looks good!

chrishtr (approved):

PR #916, opened 2023-09-13
Split `DisplayItem` into two classes

pavpanchekha:
This PR splits the old `DisplayItem` superclass for display items into two: `DrawCommand` for paint commands and `VisualEffect` for visual effects. This has a few benefits:

- We no longer need to introduce `is_paint_command` initially. (Along the way, this fixes bugs like `DrawImage` technically claiming not to be a paint command.)
- We also do not need to pretend that drawing commands offer methods like `clone` and `map`
- Also drawing commands no longer need (dummy) fields like `node` and `needs_compositing`.

Also I changed how "composited bounds" work in this PR. Basically, the `rect` field always contains the union of all child `rect`s. This passes the tests and per my careful checks but equivalent to the existing source code. That said, I don't actually think it's correct with its handling of `Transform`s, I just don't think the existing code is, either.

### DIFF
--- a/book/animations.md
+++ b/book/animations.md
@@ -430,7 +430,9 @@ class ClipRRect:
             return "ClipRRect(<no-op>)"
 ```
 
-Now we can print out our browser's display list:
+You'll also need to add `children` fields to all of the paint
+commands, since `print_tree` relies on those. Now we can print out our
+browser's display list:
 
 ``` {.python expected=False}
 class Tab:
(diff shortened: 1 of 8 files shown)


book/animations.md
```diff
@@ -404,48 +404,12 @@ of multiple windows. "Compositing" can also refer to multi-threaded rendering.
...
-that's overridden for specific display commands. For example, we can make sure
-that every display command has a `children` field:
-
-``` {.python replace=children%3d[]/rect%2c%20children%3d[]%2c%20node%3dNone}
-class DisplayItem:
-    def __init__(self, children=[]):
-        self.children = children
-```
-
-We'll need that `children` field to use `print_tree` to print the
```
chrishtr, line 420: This was lost

book/animations.md
```diff
@@ -587,30 +551,60 @@ size.
 Compositing leaves
 ==================
 
-Let's implementing compositing. We'll need to traverse the display
-list, identify paint commands, and move them to composited layers.
-Then we'll need to create the draw display list that combines these
-composited layers. To keep things simple, we'll start by creating a
-composited layer for every paint command.
+Let's implementing compositing. We'll need identify paint commands and
```
chrishtr, line 554:
```suggestion
Let's implement compositing. We'll need identify paint commands and
```

book/animations.md
```diff
@@ -587,30 +551,60 @@ size.
...
+move them to composited layers. Then we'll need to create the draw
+display list that combines these composited layers. To keep things
+simple, we'll start by creating a composited layer for every paint
```
chrishtr, line 557: For every paint command or every visual effect?

book/animations.md
```diff
@@ -1672,16 +1653,18 @@ class Browser:
...
         ]
         # ...
         for cmd in non_composited_commands:
             # ...
 
 ```
 
+The multiple `if` statements inside the list comprehension are "and"ed
+together.
+=======
```
chrishtr, line 1667: conflict markers

chrishtr:
> Also I changed how "composited bounds" work in this PR. Basically, the `rect` field always contains the union of all child `rect`s. This passes the tests and per my careful checks but equivalent to the existing source code. That said, I don't actually think it's correct with its handling of `Transform`s, I just don't think the existing code is, either.

The old code (and the new code) should be correct, because transforms are always composited. I did that to simplify the code.

pavpanchekha:
Oh, yeah, I see, transforms are always composited so their `.rect` will never matter. Great.

chrishtr:
Can you resolve the conflicts? Then I can do some local testing to make sure nothing is broken, and complete the review.

PR #918, opened 2023-09-13
Simplify LaTeX build

pavpanchekha:
This PR redoes the LaTeX build to be based on `\include` instead of using shell scripts to concatenate files. The basic setup is now:

- `latex/book.tex` is a normal, version-controlled file that you can edit directly. It contains `\include` commands for all the chapters, in order. There is no template for this file, and it isn't generated by Pandoc, it's just hand-written.
- `latex/*-chapter.tex` are still generated by the makefile, and are included directly
- `latex/book.pdf` is still generated by LaTeX from `latex/book.tex` and now depends on all the chapters, too. I switched the compilation from `pdflatex` to `latexmk`, which I believe now comes with all major LaTeX distros. (If you don't have it, I'll switch back, but it is nice in a lot of ways, like rerunning `pdflatex` the right number of times.)


chrishtr:
> * `latex/book.tex` is a normal, version-controlled file that you can edit directly. It contains `\include` commands for all the chapters, in order. There is no template for this file, and it isn't generated by Pandoc, it's just hand-written.'

This is problematic, because pandoc dynamically includes all the necessary latex modules and features neccessary to render the pandoc you've written. If we make changes in the future it may not be in sync.

For this reason, I don't see the advantage of landing these aspects of your PR? That's why I decided not to do this PR myself a couple of weeks ago (forgot to mention it to you sorry).

> I switched the compilation from `pdflatex` to `latexmk`, which I believe now comes with all major LaTeX distros. (If you don't have it, I'll switch back, but it is nice in a lot of ways, like rerunning `pdflatex` the right number of times.)

If latexmk is a drop-in replacement for pdflatex, that seems fine.

pavpanchekha:
Ok, got it. By dynamically include, do you mean the $if$ in the book-template.tex file? I looked over those and most of those seem like cases where we'll want to take direction more from OUP's style than from Pandoc. (Like setting fonts, line spacing, microtype, etc.) The running headers & frontmatter/mainmatter a good examples of this, where Pandoc *can* insert the commands for this, but only for the default class file, but the OUP class file uses different commands. More generally, I at least feel a bit more confident editing the LaTeX directly rather than editing the LaTeX by modifying the Pandoc flags. The stuff we're going to need, like syntax highlighting stuff, I kept. We can discuss.

Latexmk is kind of drop-in. (It wraps pdflatex, or if we pass the right flags xelatex.)

chrishtr:
> Ok, got it. By dynamically include, do you mean the if in the book-template.tex file?

Yes.
 
> Latexmk is kind of drop-in. (It wraps pdflatex, or if we pass the right flags xelatex.)

PR #919, opened 2023-09-20
Don't dereference None in the click method

chrishtr:


pavpanchekha:
I'm confused, does this actually happen? As far as I can tell that whole `while` loop is guarded by `elt` being truthy...

chrishtr:
Yes it does happen - see the inner while loop just above the line I'm adding. It repros if you try to load lab13 with examples/example13-width-transition.html and click one of the `<button>` elements.

src/lab12.py
```diff
@@ -283,7 +283,8 @@ def click(self, x, y):
                     if elt.tag == "form" and "action" in elt.attributes:
```
pavpanchekha, line 283: Ok, now that I understand the issue, what if we instead change the `while elt` above this to be `while elt.parent`. My point here is that the `if` statement in the outer `while` loop is kind of weird given the condition of that loop. It's really a problem caused by this inner `while` loop so I'd rather fix it there.

pavpanchekha (approved):
Ok, I suggested what I think is a better fix but this is minor and not worth dragging out, so please pick a fix you prefer and merge.

src/lab12.py
chrishtr: Good idea.

PR #920, opened 2023-09-20
Remove the `has_outline` method

pavpanchekha:
This PR reduces our trio of outline helper functions (`has_outline`, `parse_outline`, `paint_outline`) to two. The basic insight is that `has_outline` just directly calls `parse_outline` so should be inlined, but that to make that palatable we need to make `parse_outline` not take a zoom parameter so it's easier to call.

Along the way I deduplicated a few double-calls to `parse_outline`.

### DIFF
--- a/src/lab15.py
+++ b/src/lab15.py
@@ -390,8 +390,7 @@ class InputLayout(EmbedLayout):
             cmds.append(DrawLine(cx, self.y, cx, self.y + self.height, color, 1))
 
         cmds = paint_visual_effects(self.node, cmds, rect)
-        outline = parse_outline(self.node.style.get("outline"))
-        paint_outline(outline, rect, cmds, self.zoom)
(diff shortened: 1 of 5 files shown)


src/lab15.py
```diff
@@ -435,14 +435,16 @@ def layout(self):
...
+        for child in self.children:
+            child.paint(display_list)
+
         outline_rect = skia.Rect.MakeEmpty()
         outline_node = None
         for child in self.children:
-            node = child.node
-            if isinstance(node, Text) and has_outline(node.parent):
-                outline_node = node.parent
+            child_outline = parse_outline(child.node.parent.style.get("outline"))
```
chrishtr, line 444: if parse_outline(...):
 ...

will avoid declaring an unused local variable.

chrishtr (approved):

PR #923, opened 2023-09-27
Adjust browser chrome to fit the size of the font

chrishtr:
* Adjust chrome.md accordingly
* Add some more color here and there
* Add one more exercise about HTML-in-browser-chrome

Testing:
* I tested lab{7-16}.py on sample pages, and exercised the browser UI functionality, plus scrolling.
* I tested lab{7-13}-browser.html and all of them work.

### DIFF
--- a/book/chrome.md
+++ b/book/chrome.md
@@ -484,10 +484,10 @@ little toy browser doesn't.[^ffx]
 [^ffx]: Back in the day, browser tabs were the feature that would
     convince friends and relatives to switch from IE 6 to Firefox.
 
-Fundamentally, tabbed browsing means distinguishing between the
-browser itself and tabs that show individual web pages. The canvas the
-browser draws to, for example, is shared by all web pages, but the
-layout tree and display list are specific to one page. We need to
+Fundamentally, implementing tabbed browsing requires us to distinguish
+between the browser itself and tabs that show individual web pages.
+The canvas the browser draws to, for example, is shared by all web pages,
+but the layout tree and display list are specific to one page. We need to
 tease these two types of things apart.
 
 Here's the plan: the `Browser` class will store the window and canvas,
@@ -642,16 +642,15 @@ storing it somewhere, because our browser will have pretty simple
 chrome, meaning `paint_chrome` will be fast. In a real browser, it
 might be saved and only updated when the chrome changes.
 
-First things first: we need to avoid drawing page contents to the part
-of the browser window where the tab bar goes. Let's reserve some space
-at the top for the browser chrome. But how much? One way could be to
-pick some arbitrary browser chrome height, then try to squeeze the chrome
-into it (making fonts smaller as necessary). Another, better way, is to
-pick a font size that is easy enough to read, then compute the chrome
-height accordingly.
+First things first: we need to avoid drawing page contents to the part of the
+browser window where the browser chrome is. Let's reserve some space for it at
+the top. But how much? One way could be to pick some arbitrary browser chrome
+height, then try to squeeze the chrome into it (and making its font smaller as
+necessary). Another, better way, is to pick a font size that is easy enough to
+read, then compute the chrome height accordingly.
 
 We don't know that height yet without computing it as part of designing
-the UI of the browser, chrome, but we do know it will be a value we can
+the UI of the browser, chrome, but we do know it will be a number we can
 pass to each `Tab`:
 
 ``` {.python}
@@ -667,7 +666,8 @@ Each tab needs to make sure not to draw to those pixels:
 class Tab:
     def draw(self, canvas):
         for cmd in self.display_list:
-            if cmd.top > self.scroll + HEIGHT - self.chrome_bottom: continue
+            if cmd.top > self.scroll + HEIGHT - self.chrome_bottom:
+                continue
             if cmd.bottom < self.scroll: continue
             cmd.execute(self.scroll - self.chrome_bottom, canvas)
 ```
@@ -683,7 +683,8 @@ come in a moment):
 class Browser:
     def paint_chrome(self):
         cmds = []
-        cmds.append(DrawRect(0, 0, WIDTH, self.chrome_bottom, "white"))
+        cmds.append(
+            DrawRect(0, 0, WIDTH, self.chrome_bottom, "white"))
         return cmds
 ```
 
@@ -693,7 +694,8 @@ the page content now being `HEIGHT - self.chrome_bottom`:
 ``` {.python}
 class Tab:
     def scrolldown(self):
-        max_y = max(self.document.height - (HEIGHT - self.chrome_bottom), 0)
+        max_y = max(
+            self.document.height - (HEIGHT - self.chrome_bottom), 0)
         self.scroll = min(self.scroll + SCROLL_STEP, max_y)
 ```
 
@@ -730,17 +732,21 @@ class DrawLine:
         )
 ```
 
-Let's start drawing the chrome. We'll have the following:
-* At the top, a list of tab names, separated by vertical lines, and a "+"
-  button ot add a new tab.
-* Underneath, the URL of of the current web page, plus a "<" to indicate the
-  browser back-button.
+Let's start drawing the chrome. We'll end up with the following:
+
+* At the top, a list of tab names, separated by vertical lines, and a "`+`"
+  button to add a new tab.
+
+* Underneath, the URL of of the current web page, and a "`<`" button to
+  represent the browser back-button.
 
 Given this design, we can now figure out `chrome_bottom`: it's the vertical
-height of those two lines plus some padding after them. For convenience, and
-to use them later for processing mouse clicks, I'll store these parameters
-on the `Browser` object:
-
+height of those two lines plus some padding between and after them. For
+convenience, and to use them later for processing mouse clicks, I'll store
+these parameters on the `Browser` object:^[I also chose `20px` as the
+font size. Depending on your computer, this may end up looking smaller,
+\index{device pixel ratio}
+because of the device pixel ratio of the screen.]
 
 ``` {.python}
 class Browser:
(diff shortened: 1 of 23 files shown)


chrishtr:
Please review carefully, it's a big change...

infra/compile.py
```diff
@@ -438,6 +438,8 @@ def compile_function(name, args, ctx):
         return "(new Error(" + args_js[0] + "))"
     elif name == "type":
         return args_js[0] + ".constructor"
+    elif name == "range":
+        return "[...Array(" + args_js[1] + ").keys()]"
```
pavpanchekha, line 442: Very clever, is this really the normal way to do this in JS?

src/lab11.py
```diff
@@ -417,28 +419,43 @@ def __init__(self):
...
+    def init_chrome(self):
+        self.chrome_font = get_font(20, "normal", "roman")
+        chrome_font_height = linespace(self.chrome_font)
+
+        self.padding = 5
+        self.tab_header_bottom = chrome_font_height + 2 * self.padding
+        self.addressbar_top = self.tab_header_bottom + self.padding
+        self.chrome_bottom = \
+            math.ceil(
+                self.addressbar_top + chrome_font_height + 2 * self.padding)
```
pavpanchekha, line 431: I assume this doesn't change over the course of a run? I'd rather these be locals, not fields. (And `self.padding` would ideally by a global, maybe `CHROME_PADDING_PX`.)

src/lab11.py
```diff
@@ -417,28 +419,43 @@ def __init__(self):
...
     def handle_click(self, e):
-        if e.y < CHROME_PX:
+        if e.y < self.chrome_bottom:
             self.focus = None
-            if 40 <= e.x < 40 + 80 * len(self.tabs) and 0 <= e.y < 40:
-                self.active_tab = int((e.x - 40) / 80)
-                self.tabs[self.active_tab].render()
-                self.raster_tab()
-            elif 10 <= e.x < 30 and 10 <= e.y < 30:
+            if intersects(e.x, e.y, self.plus_bounds()):
```
pavpanchekha, line 440: Oh man, this code is kind of ugly. I hadn't thought about hit testing.

src/lab11.py
```diff
@@ -417,28 +419,43 @@ def __init__(self):
...
                 self.tabs[self.active_tab].go_back()
-            elif 50 <= e.x < WIDTH - 10 and 50 <= e.y < 90:
+            elif intersects(e.x, e.y, self.addressbar_bounds()):
                 self.focus = "address bar"
                 self.address_bar = ""
+            else:
+                for i in range(0, len(self.tabs)):
+                    if intersects(e.x, e.y, self.tab_bounds(i)):
+                        self.active_tab = i
+                        break
```
pavpanchekha, line 451: I guess this code now allows variable-width tabs. Arguably this code is better—instead of an error-prone math computation this just checks each extant tab.

src/lab11.py
```diff
@@ -417,28 +419,43 @@ def __init__(self):
...
                 self.load(URL("https://browser.engineering/"))
-            elif 10 <= e.x < 35 and 50 <= e.y < 90:
+            elif intersects(e.x, e.y, self.backbutton_bounds()):
                 self.tabs[self.active_tab].go_back()
-            elif 50 <= e.x < WIDTH - 10 and 50 <= e.y < 90:
+            elif intersects(e.x, e.y, self.addressbar_bounds()):
                 self.focus = "address bar"
                 self.address_bar = ""
+            else:
+                for i in range(0, len(self.tabs)):
```
pavpanchekha, line 448: A more Python-y way to do it would be `for i, tab in enumerate(self.tabs)`. We do use this idiom already (actually only to iterate over tabs).

src/lab11.py
```diff
@@ -481,38 +498,92 @@ def raster_tab(self):
...
 
-    # We need to keep this because it calls get_font
+    def plus_bounds(self):
+        plus_width = self.chrome_font.measureText("+")
+        return (self.padding, self.padding,
+            plus_width + self.padding, self.tab_header_bottom - self.padding)
+
+    def tab_bounds(self, i):
+        (plus_left, plus_top, plus_right, plus_bottom) = self.plus_bounds()
+        tab_start_x = plus_right + self.padding
```
pavpanchekha, line 508: I know why this is here but it feels to me like inlining the relevant parts of `plus_bounds` and simplifying would be better.

src/lab11.py
```diff
@@ -481,38 +498,92 @@ def raster_tab(self):
         canvas.clear(skia.ColorWHITE)
         active_tab.raster(canvas)
 
-    # We need to keep this because it calls get_font
+    def plus_bounds(self):
+        plus_width = self.chrome_font.measureText("+")
+        return (self.padding, self.padding,
+            plus_width + self.padding, self.tab_header_bottom - self.padding)
```
pavpanchekha, line 504: I guess my main aesthetic objection is that we're "polluting" the `Browser` class with a bunch more methods so it's harder to glance at the code and say "what does `Browser`, fundamentally, do?" I'm not sure if it'd be better to move all of this to a `Chrome` class, but that seems like the other option.

pavpanchekha: Potentially the `Chrome` class could have a `layout` and `paint` method? (`layout` would compute the plus / tab / backbutton / urlbar bounds and store them in fields, `paint` would read those fields?) That might be a nice way to talk about the similarity between chrome and web content without actually doing HTML-y layout.

src/lab15-tests.md
```diff
@@ -157,11 +158,21 @@ Iframes can be sized too:
...
+     DrawText(text=.)
+     DrawText(text=.)
+     DrawText(text=.)
+     DrawText(text=.)
+     DrawText(text=.)
+     DrawText(text=.)
+     DrawText(text=.)
+     DrawText(text=.)
+     DrawText(text=.)
+     DrawText(text=.)
```
pavpanchekha, line 170: Hahaha

pavpanchekha (commented):
Ok, I've read through the code, but not yet the changes to the book text. Overall, I think the code is good but I'm thinking that perhaps we should actually go further here and move all of the chrome drawing to an independent `Chrome` class that has its own `layout` and `paint` phases: `layout` computes the position of all of the chrome elements and saves those as fields on `Chrome`, and `paint` outputs the drawing commands. This would highlight the fundamental similarity between chrome and web content.

Also now that I think about it, perhaps it should _also_ have a `hit_test` command that tells you which object was pressed when you click it. Because this code is primarily going to live in Chapter 7, which already does clicking—we could refactor layout trees to _also_ have `hit_test` commands and work in the same way? (Though that is tougher, because we need to think about how `transform`, which we add later, is supposed to work. The way it works now isn't that clean though, so the refactor is probably still good?)

Let's discuss this a bit more in meeting, especially the question of how hit testing is "supposed" to work.

infra/compile.py
chrishtr: I have no idea - it was the top suggestion on Stack Overflow :)

src/lab11.py
chrishtr: Padding could be a global. The rest can be local only if we are willing to recompute them in the *_bounds methods every time.

src/lab11.py
chrishtr: Really? It seems much more readable to me than the old code.

chrishtr:
Updated to use a new Chrome class. Please review

chrishtr:
I didn't pre-compute all of the *bounds methods, because:
* One of them is parameterized on i and I don't want to precompute for all tabs
* It didn't seem to make the book more readable

book/chrome.md
```diff
@@ -656,41 +681,49 @@ Each tab needs to make sure not to draw to those pixels:
...
 
-There are still sometimes going to be halves of letters that stick out
+Now let's turn our attention to designing the UI.
+
+First, there are still sometimes going to be halves of letters that stick out
 into the browser chrome, but we can hide them by just drawing over
-them:
+them (here I'm assuming we've already computed `self.bottom` (representing
+the bottom y coordinate of the browser chrome); that will
+come in a moment):
```
pavpanchekha, line 696: Let's just move the outer parenthetical to a sentence, I think you're not supposed to do nested parentheses in normal prose :)

book/chrome.md
```diff
@@ -656,41 +681,49 @@ Each tab needs to make sure not to draw to those pixels:
...
         cmds = []
-        cmds.append(DrawRect(0, 0, WIDTH, CHROME_PX, "white"))
+        cmds.append(
+            DrawRect(0, 0, WIDTH, self.bottom, "white"))
         return cmds
 ```
 
 You'll also need to adjust `scrolldown` to account for the height of
-the page content now being `HEIGHT - CHROME_PX`:
+the page content now being `HEIGHT - self.bottom`:
```
pavpanchekha, line 708: Hmm, we've been passing `chrome_bottom` to each `Tab`, but in Chapter 15 we instead seem to care about `frame_height`. Should we pass `HEIGHT - self.bottom` instead? It will align more with Chapter 15 and also the code will be a tiny bit nicer.

pavpanchekha, line 708: Also shouldn't this be `HEIGHT - self.chrome_bottom`?

book/chrome.md
```diff
@@ -715,70 +748,92 @@ class DrawLine:
         )
 ```
 
-Let's start drawing the chrome. First: the tab bar at the top of the
-browser window. I'll keep it simple, but this is still going to
-require some tedious and mildly tricky geometry.
+Now let's figure out the browser chrome size. We'll end up with the following:
```
pavpanchekha, line 751:
```suggestion
Now let's design the browser chrome. I think it should have the following:
```

book/chrome.md
```diff
@@ -715,70 +748,92 @@ class DrawLine:
...
-`120 + 80*i`:
+* Underneath, the URL of of the current web page, and a "`<`" button to
+  represent the browser back-button.
 
-``` {.python}
-for i, tab in enumerate(self.tabs):
-    name = "Tab {}".format(i)
-    x1, x2 = 40 + 80 * i, 120 + 80 * i
-```
+Let's reserve some space for all this at the top of the window. But how much?
```
pavpanchekha, line 759:
```suggestion
We'll need reserve some space for all this at the top of the window. But how much?
```

book/chrome.md
```diff
@@ -715,70 +748,92 @@ class DrawLine:
...
+One way could be to pick some arbitrary browser chrome height, then try to
+squeeze the chrome into it (and making its font smaller as necessary). Another,
+better way, is to pick a font size that is easy enough to read, then compute
+the chrome height accordingly.
 
-For each tab, we need to create a border on the left and right and
-then draw the tab name:
+Given this design, we can now figure out `self.bottom`: it's the vertical
```
pavpanchekha, line 765:
```suggestion
For our design, the chrome's height should be the vertical
```

book/chrome.md
```diff
@@ -715,70 +748,92 @@ class DrawLine:
...
+height of those two lines plus some padding between and after them. For
+convenience, and to use them later for processing mouse clicks, I'll store
+these parameters on the `Browser` object:^[I also chose `20px` as the
```
pavpanchekha, line 768:
```suggestion
these parameters on the `Chrome` object:^[I also chose `20px` as the
```

book/chrome.md
```diff
@@ -715,70 +748,92 @@ class DrawLine:
...
-        cmds.append(DrawLine(x2, 40, WIDTH, 40, "black", 1))
+class Chrome:
+    def plus_bounds(self):
+        plus_width = self.font.measure("+")
+        return (self.padding, self.padding,
+            plus_width + self.padding,
+            self.tab_header_bottom - self.padding)
+
+    def tab_bounds(self, i):
+        tab_start_x = self.font.measure("+") + \
```
pavpanchekha, line 809: Some of these `bounds` methods I'd like to be a little nicer to read. Some options:

- Minimally, can we do `self.padding + self.font.measure("+") + self.padding`, reflecting the actual left-to-right order?
- Or can we call `self.plus_bounds` and do offsets from there?
- I still think we could do better if it was one `layout` method, but I agree we shouldn't delay this PR

book/chrome.md
```diff
@@ -800,51 +855,97 @@ class DrawOutline:
...
+that end, let's add a quick method to test whether a point intersects one
+of them:
 
 ``` {.python}
-if e.y < CHROME_PX:
-    if 40 <= e.x < 40 + 80 * len(self.tabs) and 0 <= e.y < 40:
-        self.active_tab = int((e.x - 40) / 80)
+def intersects(x, y, rect):
+    (left, top, right, bottom) = rect
+    return x >= left and x < right and y >= top and y < bottom
```
pavpanchekha, line 933: This is sort of neither here nor there but this method (and some general chaos in the `DrawXXX` methods) makes me think that we'd be better off defining a simple `Rect` class for the Tk chapters that transitions into `skia.Rect` once we move to Skia.

book/chrome.md
```diff
@@ -869,12 +970,17 @@ get a little lost and forget what web page you're looking at. An
...
+        top_bar = addressbar_top + self.padding
         # ...
-        cmds.append(DrawOutline(40, 50, WIDTH - 10, 90, "black", 1))
-        url = str(self.tabs[self.active_tab].url)
-        cmds.append(DrawText(55, 55, url, buttonfont, "black"))
+            url = str(self.browser.tabs[self.browser.active_tab].url)
+            cmds.append(DrawText(
+                left_bar,
+                top_bar,
+                url, self.font, "black"))
```
pavpanchekha, line 983: What's with the indent? If we want to change stuff within a block I always put the first line of the block somewhere to notate that.

book/chrome.md
```diff
@@ -1204,8 +1349,26 @@ should delete the character before the cursor, and typing other keys
...
 
-*Multiple windows* Add support for multiple browser windows in
+*Multiple windows*: Add support for multiple browser windows in
 addition to tabs. This will require keeping track of multiple Tk
 windows and canvases and grouping tabs by their containing window.
 You'll also need some way to create a new window, perhaps with a
 keypress such as `Ctrl+N`.
+
+*Reusing HTML*: Browser chrome is quite complicated in real browsers, and even
+their layouts contain a lot of details, such as font sizes, padding, outlines,
```
pavpanchekha, line 1359:
```suggestion
*Reusing HTML*: Browser chrome is quite complicated in real browsers,
with tricky details such as font sizes, padding, outlines,
```

pavpanchekha (approved):
I have a lot of comments on the text but they are all minor—take them or leave them and then let's merge it!

book/chrome.md
```diff
@@ -715,70 +750,92 @@ class DrawLine:
...
-`120 + 80*i`:
+* Underneath, the URL of of the current web page, and a "`<`" button to
+  represent the browser back-button.
 
-``` {.python}
-for i, tab in enumerate(self.tabs):
-    name = "Tab {}".format(i)
-    x1, x2 = 40 + 80 * i, 120 + 80 * i
-```
+We'll need reserve some space for all this at the top of the window. But how much?
```
chrishtr, line 761:
```suggestion
We'll need to reserve some space for all this at the top of the window. But how much?
```

PR #925, opened 2023-09-27
Require crossOriginIsolated for lab9+-browser.html

chrishtr:
The JS error thrown when crossOriginIsolated is false causes a JS error in the browser that breaks the whole thing, and it's not worth trying to recover from it.

### DIFF
--- a/www/widgets/rt.js
+++ b/www/widgets/rt.js
@@ -411,8 +411,7 @@ class math {
 
 class JSInterpreterError extends ExpectedError {
     constructor() {
-        super("This widget cannot execute JavaScript due to sandboxing, " +
-             "but the book's Python code should work correctly.");
+        super("This widget cannot work unless crossOriginIsolated is true.");
         this.name = "JSEnvironmentError."
     }
 }


pavpanchekha:
I'm a little confused. This PR just changes the error message? I kind of think the old message was better, though I'm willing to go either way, but from the description it sounds like you meant a different change.

chrishtr:
The old message implies the browser still works. The new one indicates it's broken.

pavpanchekha:
Hmm, ok. The old message is trying to say that the JS version is broken but the Python version developed in the book, when run in Python, will work fine. Probably what we should actually do is redo the widget framework to have a "check" step before it actually renders that will say something like "This chapter doesn't work in your browser due to sandboxing. (Specifically, crossOriginIsolated needs to be true.)"

chrishtr:
> Hmm, ok. The old message is trying to say that the JS version is broken but the Python version developed in the book, when run in Python, will work fine. Probably what we should actually do is redo the widget framework to have a "check" step before it actually renders that will say something like "This chapter doesn't work in your browser due to sandboxing. (Specifically, crossOriginIsolated needs to be true.)"

Ok, but in the meantime, is making this error message not incorrect ok by you?

PR #926, opened 2023-09-28
Make tkinter fonts faster

chrishtr:
Credit to Lars Hamann for this hack.

### DIFF
--- a/book/text.md
+++ b/book/text.md
@@ -793,7 +793,7 @@ FONTS = {}
 
 The keys to this dictionary will be size/weight/style triples, and the
 values will be `Font` objects. We can put the caching logic itself in
-a new `get_font` function:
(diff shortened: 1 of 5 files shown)


h-man2:
I found the [source](https://github.com/python/cpython/blob/main/Lib/tkinter/font.py) (line 163ff.) dated back to 1998 an written by Fredrik Lundh:

```python
 def metrics(self, *options, **kw):
        """Return font metrics.

        For best performance, create a dummy widget
        using this font before calling this method."""
```

pavpanchekha (approved):
It would be nice to link to Fredrik Lundh in the footnote, and also make sure Lars shows up in our acknowledgements (which we can do by modifying `gh_names` in `infra/api.py`). But good change and crazy good find by Lars.

PR #930, opened 2023-10-03
Add hit testing exercises

chrishtr:


book/chrome.md
```diff
@@ -1209,3 +1209,23 @@ addition to tabs. This will require keeping track of multiple Tk
...
+logic for `Text` objects and `a` tags. Real browsers have many more special
+kinds of nodes, plus more complicated layout, so they tend to implement this
+logic directly on the layout tree.^[And they also use this logic for more than
+just clicking. The more general task of determining which objects on screen
+overlap a given point or region is called *hit testing*---by analogy with
+whether an arrow shot at that location would "hit" the object.] Implement
+`click` on the layout tree.
+
+*Hit testing on the display list*: Hit testing can be thought of as a "reversed"
+version of `paint`, since it looks at the elements front-to-back in paint order,
```
pavpanchekha, line 1224:
```suggestion
version of `paint`: `paint` turns elements into pixels,
while hit testing turns pixels into elements. Plus, it looks
at the elements front-to-back in paint order,
```

book/chrome.md
```diff
@@ -1209,3 +1209,23 @@ addition to tabs. This will require keeping track of multiple Tk
...
+as opposed to back-to-front. Building on this observation, we could also create
+a special display list that represents hit testing information, or build all
+of the necessary information for hit testing directly into the display list
```
pavpanchekha, line 1227:
```suggestion
as opposed to back-to-front. Building on this observation, we could build all
of the necessary information for hit testing directly into the display list
```

book/chrome.md
```diff
@@ -1209,3 +1209,23 @@ addition to tabs. This will require keeping track of multiple Tk
...
+instead of the layout tree. Implement one of these.^[You might want to implement
+hit testing in this way in a browser because display lists are pure data
+structurs and therefore easier to optimize or execute in different
+threads.]
```
pavpanchekha, line 1231:
```suggestion
instead of the layout tree. Implement one of these.^[This is more
challenging in real browsers, because you might want some kinds of
invisible elements to be clickable. But the advantage is that display
lists are pure data structures which are easy to pass between threads.]
```

pavpanchekha (approved):
I'm happy to merge as is, but to me the display list hit testing exercise would be better in Chapter 11 or 13, because that's the point where we add `transform` which ends up making hit testing kind of ugly. The display list hit testing would avoid that.

PR #939, opened 2023-10-07
Add mechanism for print-only, use it to omit last part of preface

chrishtr:


chrishtr:
With this new mechanism we can easily put stuff in `.web-only` or `.print-only` blocks, e.g. to distinguish figures from web or print mode.

pavpanchekha (approved):

PR #941, opened 2023-10-08
Resolve feedback about history, plus editorial improvements

chrishtr:
This fixes the one remaining item in our feedback document.

### DIFF
--- a/book/history.md
+++ b/book/history.md
@@ -5,10 +5,10 @@ next: http
 prev: intro
 ...
 
-This chapter digs into the web\index{web} itself: where it came from,
-and how the web and browsers have evolved to date. This history is not
-exhaustive;[^sgml] the focus is the key events and ideas that led to
-the web, and the goals and motivations of its inventors.
+This chapter dives into the history of the web\index{web} itself: where it came
+from, and how the web and browsers have evolved to date. This history is not
+exhaustive;[^sgml] the focus is the key events and ideas that led to the web,
+and the goals and motivations of its inventors.
 
 [^sgml]: For example, there is nothing much about
 [SGML](https://en.wikipedia.org/wiki/Standard_Generalized_Markup_Language) or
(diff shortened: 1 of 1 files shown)


book/history.md
```diff
@@ -85,7 +85,7 @@ text that is marked up with hyperlinks to other text.[^back-button]
 Sound familiar? A web page\index{web page} is hypertext, and links
 between web pages are hyperlinks. The format for writing web pages is
 HTML, which is short for HyperText Markup Language. The protocol for
-loading web pages is HTTP, which is short for HyperText Transport
+loading web pages is HTTP, which is short for HyperText Transfer
```
pavpanchekha, line 88: Wow great catch

book/history.md
```diff
@@ -214,22 +217,22 @@ worth quoting here (highlight is mine):
...
+   make a simple web page and publish it.
 
- - Once bootstrapped, it builds quickly upon itself via [network
+ - Once bootstrapped, it built quickly upon itself via [network
 effects](https://en.wikipedia.org/wiki/Network_effect) made possible by
 compatibility between sites and the power of the hyperlink to reinforce this
-compatibility. Hyperlinks drive traffic between sites, but also into the web
+compatibility. Hyperlinks drove traffic between sites, but also into the web
 _from the outside_, from sources such as email, social networking, and search
 engines.
```
pavpanchekha, line 233: There's still something off to me about these bullet points. I think part of it is the way that the first bullet point is benefit-then-reason, while the second does kind of both at once, while the third is reason-then-benefit. I would rephrase all of them to be reason-then-benefit:

> - Because there was no gatekeeper to doing anything, it was easy for novices to make simple web pages and publish them.
> - Because pages were identified simply by URLs, traffic could come to the web from outside sources like email, social networking, and search engines.
> - Because sites were compatible, the web built quickly upon itself via network effects.
> - Because the web was outside the control of any one entity---and kept that way via standards organizations---it avoided the problems of monopoly control and manipulation.

pavpanchekha (approved):

PR #942, opened 2023-10-08
Address feedback items for chapter 1

chrishtr:
* Exercise added to chapter 2 for about:blank
* Drop assert error messages
* Removed extra commentary about HTML syntax
* Get rid of 200 assert (I really like this change!)
* Use casefold throughout all chapters
* Don't return headers from `request` until chapter 10

Also:
* Changed numbered footnotes into named ones, and put them inline

### DIFF
--- a/book/graphics.md
+++ b/book/graphics.md
@@ -591,6 +591,13 @@ window's new width and height can be found in the `width` and `height`
 fields on the event object. Remember that when the window is resized,
 the line breaks must change, so you will need to call `layout` again.
 
+*about:blank:* Currently, a malformed URL causes the browser to crash.
+It would be much better to have error recovery for that, and instead
+show a blank page, so that the user can fix the error. To do this, add
+support for the special `about:blank` URL, which should just render
+a blank page, and cause malformed URLs to automatically render as if they
+were `about:blank`.
+
 [fill-expand]: https://web.archive.org/web/20201111222645id_/http://effbot.org/tkinterbook/pack.htm
 
 [^3]: On older systems, applications drew directly to the screen, and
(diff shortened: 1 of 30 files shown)


book/graphics.md
```diff
@@ -591,6 +591,13 @@ window's new width and height can be found in the `width` and `height`
 fields on the event object. Remember that when the window is resized,
 the line breaks must change, so you will need to call `layout` again.
 
+*about:blank:* Currently, a malformed URL causes the browser to crash.
```
pavpanchekha, line 594: Any reason this goes in Chapter 2 instead of 1? Just because we need more in Chapter 2?

book/security.md
```diff
@@ -792,9 +792,9 @@ def request(self, payload=None):
         params = {}
         if ";" in cookie:
             cookie, rest = cookie.split(";", 1)
-            for param_pair in rest.split(";"):
+        for param_pair in rest.split(";"):
```
pavpanchekha, line 795: Confused by this change. Either we need to initialize `rest` as the empty string, or the `for` loop needs to be inside the `if` statement. But maybe this code block was already misshapen? Or there's a spaces-vs-tabs issue?

pavpanchekha (approved):

book/graphics.md
chrishtr: Yes: chapter 1 doesn't have a browser, it's just a commandline tool. It makes more sense to me to justify this when there is a browser UI to think about.

book/security.md
chrishtr: This change was not on purpose, and seems to be a whitespace issue. Fixed.

PR #944, opened 2023-10-10
Add example output for chapters 1 and 2

chrishtr:
* Chapter 1: text output
* Chapter 2: one screenshot

* I also fixed the scrolling in the chapter 2 widget.


book/http.md
```diff
@@ -835,6 +835,21 @@ not recommended. It's very difficult to write a custom TLS
 implementation that is not only correct but secure.
 :::
 
+At this point you should be able to run your program on any web plage.
```
pavpanchekha, line 838:
```suggestion
At this point you should be able to run your program on any web page.
```

book/http.md
```diff
@@ -835,6 +835,21 @@ not recommended. It's very difficult to write a custom TLS
...
+Here is what it should output for [this example][example-simple]:
+
+[example-simple]: examples/example1-simple.html
```
pavpanchekha, line 841: I'm OK with this but why not use `example.com`? It's been a running example in most of the chapter.

pavpanchekha (approved):

book/http.md
chrishtr: I tried that, but it spams out a lot of text from the style sheet. Rather than have a wall of text that covers two pages, or omitting part of the actual output, I opted for a simpler web page.

PR #947, opened 2023-10-12
Minor changes to Chapter 5

pavpanchekha:
This chapter has a pretty big reordering of the layout paragraphs, but the code is kept mostly the same. Otherwise, all of the comments seem minor.

### DIFF
--- a/book/layout.md
+++ b/book/layout.md
@@ -9,9 +9,9 @@ So far, layout has been a linear process that handles open tags and
 close tags independently. But web pages are trees, and look like them:
 borders and backgrounds visually nest inside one another. To support
 that, this chapter switches to *tree-based layout*,\index{layout}
-where the tree of elements is transformed into a tree of *layout objects*
-for the visual elements of the page. In the process, we'll make our web
-pages more colorful with backgrounds.
+where the tree of elements is transformed into a tree of *layout
+objects* before drawing. In the process, we'll make web pages more
+colorful with backgrounds.
 
 The layout tree
 ===============
@@ -20,12 +20,13 @@ Right now, our browser lays out an element's open and close tags
 separately. Both tags modify global state, like the `cursor_x` and
 `cursor_y` variables, but they aren't otherwise connected, and
 information about the element as a whole, like its width and height,
-is never computed. That makes it pretty hard to draw a background color
-behind text. So web browsers structure layout differently.
+is never computed. That makes it pretty hard to draw a background
+behind an element, let alone more complicated visual effects. So web
+browsers structure layout differently.
 
 In a browser, layout is about producing a *layout tree*,\index{layout tree}
 whose nodes are *layout objects*, each associated with an HTML
-element,[^no-box] and each with a size and a position. The browser walks
+element[^no-box] and each with a size and a position. The browser walks
 the HTML tree to produce the layout tree, then computes the size and position
 for each layout object, and finally draws each layout object to the screen.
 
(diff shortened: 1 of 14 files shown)


book/layout.md
```diff
@@ -460,32 +455,23 @@ class BlockLayout:
...
-        self.x = HSTEP
-        self.y = VSTEP
-        child.layout()
-        self.height = child.height + 2*VSTEP
-```
-
-Note that there's some padding around the contents---`HSTEP` on the
-left and right, and `VSTEP` above and below. That's so the text won't
-run into the very edge of the window and get cut off.
+On the contrary, an element's `height` field
```
chrishtr, line 465:
```suggestion
On the other hand, an element's `height` field
```

book/layout.md
```diff
@@ -508,6 +494,24 @@ hasn't been computed yet, and the browser will have a bug. We'll come
...
+
+``` {.python}
+class DocumentLayout:
+    def layout(self):
+        # ...
+        self.width = WIDTH - 2*HSTEP
+        self.x = HSTEP
+        self.y = VSTEP
+        child.layout()
+        self.height = child.height + 2*VSTEP
```
chrishtr, line 508: why 2*VSTEP and not VSTEP here?

book/layout.md
```diff
@@ -646,10 +652,7 @@ class BlockLayout:
         # ...
 ```
 
-Note that we must add the block's `x` and `y`, since the positions in
```
chrishtr, line 649: Is this comment no longer relevant?

pavpanchekha: Correct. The block's `x` and `y` are added earlier, when adding the positions to the `self.display_list`.

book/layout.md
pavpanchekha: Ah, to put the padding on both sides.

book/layout.md
chrishtr: Is that just a "bug" in the text before your PR then?

book/layout.md
chrishtr: But self.y = VSTEP already puts a padding on the top?

book/layout.md
pavpanchekha: Yes. It seems to have been introduced when we switched `cursor_x` to be a relative measure from the start. (When published it was absolute and switched to relative in Ch7.)

book/layout.md
pavpanchekha: Right, but the `height` is needed to add the padding at the bottom too.

chrishtr: I still don't get it. The document's top is at VSTEP. It has height equal to content height plus 2*VSTEP. Doesn't this result in a VSTEP padding at the top and 2*VSTEP padding at the bottom?

pavpanchekha: Oh, yeah, I get the confusion. This is actually kind of weird!

So here's how these fields get used. The `document_height` field used for scrolling is just the `DocumentLayout`'s `height` field. We don't add on `y` or anything like that. In fact, that's the only use of this `height` field. So in that sense the correct answer is the current value.

That said it's kind of weird. We can say that the `height` field represents the padding box height, not just content height—but then it's weird that the `y` field represents the content position and not the padding box position. Unfortunately, we don't have a concept of padding, though, so the `y` position here really does have to be `VSTEP`.

So I think the best way to make this make sense is to set `self.height` to the content height and add on `2 * VSTEP` in in the scroll clamping computation. What do you think?

chrishtr: I like your suggestion in the last sentence.

pavpanchekha: Done.

chrishtr (approved):

PR #950, opened 2023-10-14
Address chapter 6 feedback

chrishtr:
* Adds two go-further blocks to other chapters based on chapter 6 feedback
* Renames `Layout.get_font` to `font` to avoid a conflict with the `get_font` global function

### DIFF
--- a/book/scripts.md
+++ b/book/scripts.md
@@ -1067,6 +1067,32 @@ def add_entry(params):
     return show_comments()
 ```
 
+::: {.further}
+
+Search engines are constantly [crawling] the web and [indexing] all of the web
+pages they can find. In the early days, indexing was just a matter of loading
+the HTML, parsing it and extracting the information. But these days, a lot of
+[single-page app][spa] sites use JavaScript to ["hydrate"][hydration]
+[^why-hydrate] their site into its full contents. On such sites, before
+hydration happens, the information in the site is hidden inside of JavaScript
+data structures. For this reason, search engines need to not just parse HTML,
+but also run JavaScript (and load style sheets) during indexing. In other
+words, the indexing systems use browsers (such as for example
+[headless Chrome][headless])---one more place browsers appear in the web
+ecosystem.
+
+[spa]: https://en.wikipedia.org/wiki/Single-page_application
+[hydration]: https://en.wikipedia.org/wiki/Hydration_(web_development)
+[headless]: https://chromium.googlesource.com/chromium/src/+/lkgr/headless/README.md
+
+[^why-hydrate]: This process is called "hydration" by analogy with how water
+is added to dehydrated food to make it edible again.
+
+:::
+
+[crawling]: https://en.wikipedia.org/wiki/Web_crawler
+[indexing]: https://en.wikipedia.org/wiki/Search_engine_indexing
+
 A closing thought
 =================
 
--- a/book/scripts.md
+++ b/book/scripts.md
@@ -1080,9 +1080,6 @@ def add_entry(params):
     return show_comments()
 ```
 
-[crawling]: https://en.wikipedia.org/wiki/Web_crawler
-[indexing]: https://en.wikipedia.org/wiki/Search_engine_indexing
-
 Note that we shouldn't---can't---rely on JavaScript being executed by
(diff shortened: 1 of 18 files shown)


book/scripts.md
```diff
@@ -1067,6 +1067,32 @@ def add_entry(params):
     return show_comments()
 ```
 
+::: {.further}
+
+Search engines are constantly [crawling] the web and [indexing] all of the web
+pages they can find. In the early days, indexing was just a matter of loading
+the HTML, parsing it and extracting the information. But these days, a lot of
+[single-page app][spa] sites use JavaScript to ["hydrate"][hydration]
+[^why-hydrate] their site into its full contents. On such sites, before
```
pavpanchekha, line 1076: I think the placement of `[^why-hydrate]` here adds an extra space between that and `"hydrate"`, which is not right.

src/lab6.py
```diff
@@ -11,7 +11,7 @@
 import tkinter.font
 from lab1 import URL
 from lab2 import WIDTH, HEIGHT
-from lab3 import FONTS, get_font
```
pavpanchekha, line 14: Wait, why? I think we want to import the global FONTS cache.

src/lab10.py
```diff
@@ -11,7 +11,6 @@
 import urllib.parse
 import dukpy
 from lab2 import WIDTH, HEIGHT, HSTEP, VSTEP, SCROLL_STEP
-from lab3 import FONTS, get_font
```
pavpanchekha, line 14: I think we should keep these imports around—one day I'll fix the inlining system so we don't need them but for now please keep them.

src/lab7.py
```diff
@@ -10,7 +10,7 @@
 import tkinter
 import tkinter.font
 from lab2 import WIDTH, HEIGHT, HSTEP, VSTEP, SCROLL_STEP
-from lab3 import FONTS, get_font
```
pavpanchekha, line 13: Same on keeping this.

src/lab8.py
```diff
@@ -11,7 +11,7 @@
 import tkinter.font
 import urllib.parse
 from lab2 import WIDTH, HEIGHT, HSTEP, VSTEP, SCROLL_STEP
-from lab3 import FONTS, get_font
```
pavpanchekha, line 14: Same on keeping this

src/lab9.py
```diff
@@ -12,7 +12,6 @@
 import urllib.parse
 import dukpy
 from lab2 import WIDTH, HEIGHT, HSTEP, VSTEP, SCROLL_STEP
-from lab3 import FONTS, get_font
```
pavpanchekha, line 15: Same on keeping this

pavpanchekha (changes requested):
Please undo the import stuff but other than that looks good. I think renaming `get_font` to just `font` is a great compromise.

src/lab10.py
chrishtr: Ok, but can you tell me why we need them for the inliner? I didn't know about this (or forgot).

chrishtr: I left FONTS but not get_font, which I don't think is needed?

chrishtr:
> Please undo the import stuff but other than that looks good. I think renaming `get_font` to just `font` is a great compromise.

Done

pavpanchekha:
Yeah, ok, so the summary of the annoying import rule here is that for the inliner to work properly, we need to import not just any functions / constants that we use (how importing normally works), but also any transitive dependencies. So, for example, the `Layout.font` function uses the `get_font` global function, which in turn uses the `FONTS` global constant, so we need to import them. (And in fact since we keep using the `font` method throughout, we need to keep using importing `get_font` throughout.)

src/lab10.py
pavpanchekha: I am guessing that `BlockLayout` still calls its `font` method which in turn still calls `get_font`, which we still need.

pavpanchekha (changes requested):

pavpanchekha:
Honestly it would be better to just fix the damn importer, but I guess we are prioritizing the book pass for now :)

chrishtr:
Updated.

pavpanchekha (approved):

PR #952, opened 2023-10-15
Factor out paint recursion into a paint_tree method

chrishtr:
It is a *lot* of changes through various chapters, but I hope things are clearer to the reader. Certainly there is less code
duplication, and likely the `paint_tree` method is easier for readers because it focuses on the recursion, while the
`paint` methods focus only on self-paint and self-paint-effect.

### DIFF
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -1092,7 +1092,7 @@ class LineLayout:
         outline_node = None
         for child in self.children:
             if child.node.parent.is_focused:
-                outline_rect.join(child.rect())
+                outline_rect.join(child.self_rect())
                 outline_node = child.node.parent
         if outline_node:
             paint_outline(
@@ -1356,7 +1356,7 @@ class LineLayout:
         for child in self.children:
             outline_str = child.node.parent.style.get("outline")
             if parse_outline(outline_str):
-                outline_rect.join(child.rect())
+                outline_rect.join(child.self_rect())
                 outline_node = child.node.parent
 ```
 
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -1060,20 +1060,15 @@ class Tab:
             node.is_focused = True
 ```
 
-To draw an outline, we'll use `DrawOutline`:
+To draw an outline, we'll use `DrawOutline`. This will have to happen in
+`paint_effects`, because it paints on top of the painted subtree.
 
 ``` {.python}
 class InputLayout:
-	def paint(self, display_list):
-		# ...
-        if self.node.is_focused and self.node.tag == "input":
-            cx = self.x + self.font.measureText(text)
-            cmds.append(DrawLine(cx, self.y, cx, self.y + self.height,
-                                 "black", 1))
-
+	def paint_effects(self, cmds):
         cmds = paint_visual_effects(self.node, cmds, rect)
         paint_outline(self.node, cmds, rect, self.zoom)
-        display_list.extend(cmds)
+        return cmds
 ```
 
 I also changed the cursor drawing to only happen if the node is
@@ -1087,12 +1082,12 @@ code. Moreover, those `TextLayout`s could be split across several
 lines, so we might want to draw more than one focus ring. To work
 around this, let's draw the focus ring in `LineLayout`. Each
 `LineLayout` finds all of its child `TextLayout`s that are focused,
-and draws a rectangle around them all:
+and draws a rectangle around them all. This will need to happen in
+`paint_effects` because it paints on top of the painted subtree.
 
 ``` {.python replace=child.node.parent.is_focused/parse_outline(outline_str)}
 class LineLayout:
-    def paint(self, display_list):
-        # ...
+    def paint_effects(self, cmds):
         outline_rect = skia.Rect.MakeEmpty()
         outline_node = None
         for child in self.children:
@@ -1101,7 +1096,8 @@ class LineLayout:
                 outline_node = child.node.parent
         if outline_node:
             paint_outline(
-                outline_node, display_list, outline_rect, self.zoom)
+                outline_node, cmds, outline_rect, self.zoom)
+        return cmds
 ```
 
 You should also add a `paint_outline` call to `BlockLayout`, since
@@ -1355,7 +1351,8 @@ to the browser style sheet above:
 
 ``` {.python}
 class LineLayout:
-    def paint(self, display_list):
+    def paint_effects(self, cmds):
+        # ...
         for child in self.children:
             outline_str = child.node.parent.style.get("outline")
             if parse_outline(outline_str):
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -1066,8 +1066,8 @@ To draw an outline, we'll use `DrawOutline`. This will have to happen in
 ``` {.python}
 class InputLayout:
 	def paint_effects(self, cmds):
-        cmds = paint_visual_effects(self.node, cmds, rect)
-        paint_outline(self.node, cmds, rect, self.zoom)
+        cmds = paint_visual_effects(self.node, cmds, self.rect())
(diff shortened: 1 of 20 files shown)


book/layout.md
```diff
@@ -654,29 +651,32 @@ commands in one place.
 
 ``` {.python}
 class BlockLayout:
-    def paint(self, display_list):
-        for x, y, word, font in self.display_list:
-            display_list.append(DrawText(x, y, word, font))
-        # ...
+    def paint(self):
+        cmds = []
+        if self.layout_mode() == "inline":
```
pavpanchekha, line 656: Why do we need the test here? If it's not in inline mode, the `display_list` should be empty.

book/chrome.md
```diff
@@ -309,23 +309,25 @@ class TextLayout:
...
 
 ``` {.python}
 class TextLayout:
-    def paint(self, display_list):
+    def paint(self):
+        cmds = []
         color = self.node.style["color"]
-        display_list.append(
+        cmds.append(
             DrawText(self.x, self.y, self.word, self.font, color))
```
pavpanchekha, line 329: Probably better to not even construct `cmds` to begin with, just return a one-element list.

book/forms.md
```diff
@@ -171,10 +174,12 @@ Finally, we draw that text:
 
 ``` {.python}
 class InputLayout:
-    def paint(self, display_list):
+    def paint(self):
+        cmds = []
```
pavpanchekha, line 178: Does this block need `# ...` somewhere?

book/forms.md
```diff
@@ -275,7 +280,7 @@ class BlockLayout:
             if bgcolor != "transparent":
                 x2, y2 = self.x + self.width, self.y + self.height
                 rect = DrawRect(self.x, self.y, x2, y2, bgcolor)
-                display_list.append(rect)
+                cmds.append(rect)
```
pavpanchekha, line 284: Do we need `# ...` here?

book/forms.md
```diff
@@ -504,11 +509,11 @@ an `input` element is focused:
...
 ``` {.python}
 class InputLayout:
-    def paint(self, display_list):
+    def paint(self):
         # ...
         if self.node.is_focused:
             cx = self.x + self.font.measure(text)
-            display_list.append(DrawLine(
+            cmds.append(DrawLine(
                 cx, self.y, cx, self.y + self.height, "black", 1))
```
pavpanchekha, line 517: Again a question about `# ...`

book/visual-effects.md
```diff
@@ -803,39 +803,45 @@ class SaveLayer:
...
+display list, we can use `SaveLayer` to add transparency to the whole element.
+I'm going to do this in a new `paint_effects` method, which will wrap `cmds`
+in a `SaveLayer`. The actual `SaveLayer` will be computed in a new
+global `paint_visual_effects` method (because other object types will need it
+also).^[As part of adding this method, I've factored out a `self_rect` and
+`is_atomic` method from `paint`, you'll need to update that method also to call
+these helpers.]
 
 ``` {.python}
 class BlockLayout:
```
pavpanchekha, line 817: Ok, I'm a bit confused. The reader will need to add `paint_effects` to every single layout class? Do all of them just call `paint_visual_effects`? I think we need to think this through carefully.

book/accessibility.md
```diff
@@ -1355,11 +1351,12 @@ to the browser style sheet above:
...
 ``` {.python}
 class LineLayout:
-    def paint(self, display_list):
+    def paint_effects(self, cmds):
+        # ...
         for child in self.children:
             outline_str = child.node.parent.style.get("outline")
             if parse_outline(outline_str):
-                outline_rect.join(child.rect())
+                outline_rect.join(child.self_rect())
```
pavpanchekha, line 1359: So it looks like we added `self_rect` to every inline object?

book/embeds.md
```diff
@@ -1021,9 +1021,30 @@ Most of the layout tree's `paint` methods don't need to change, but to
 paint an `IframeLayout`, we'll need to paint the child frame:
 
 ``` {.python}
+def paint_tree(layout_object, display_list):
+    cmds = layout_object.paint()
+
+    if isinstance(layout_object, IframeLayout) and \
+        layout_object.node.frame:
+        paint_tree(layout_object.node.frame.document, cmds)
```
pavpanchekha, line 1029: Why do this in `paint_tree` instead of in `IframeLayout`?

book/embeds.md
```diff
@@ -1021,9 +1021,30 @@ Most of the layout tree's `paint` methods don't need to change, but to
...
+    cmds = layout_object.paint_effects(cmds)
+    display_list.extend(cmds)
+
+```
+
+Then painting of an `IframeLayout` is just drawing a rectangle and background:
+
+::: {.todo}
+No idea why the expecation doesn't pass.
+:::
```
pavpanchekha, line 1043: I can look into it, but sometimes it's mysterious...

book/layout.md
chrishtr: This code was there before my CL. If it needs to change it should be in a new PR.

book/forms.md
chrishtr: Not this one.

book/accessibility.md
chrishtr: Yes. This was the code before?

book/embeds.md
chrishtr: Because I wanted there to be exactly three concepts:

* `paint`, which paints the current layout object
* `paint_effect`, which paints effects applying to the whole subtree (including subframes)
* `paint_tree`, which walks the whole tree across frames and patches together calls to `paint` and `paint_effect` in the right order

book/visual-effects.md
chrishtr: They don't just blindly call it. For `BlockLayout` it conditions it on being atomic.

chrishtr:
I answered all of the questions, can you re-review?

book/visual-effects.md
pavpanchekha: I thought about this a lot on the drive home last night and I think it's better to correct for the atomic stuff inside `paint_tree` rather than in `BlockLayout`. For example, suppose you have an `<input>` element with `opacity`. We only want that `opacity` applied once, right? So the atomic inline thing actually applies, I think, to any `BlockLayout` whose root element is not a block element.

Maybe the right move here is to merge this and think about another pass later. But the reason I'm so worried about this PR is that I am worried that we have actual bugs here, and like usual I'm trying to achieve "obviously no bugs" and not yet succeeding.

book/accessibility.md
pavpanchekha: I think it's fine for the PR, but I will say that I'm becoming more and more convinced that we need to add a `Rect` class in Chapter 7 and then set up the expectation of each layout object having a `rect` method earlier.

book/embeds.md
pavpanchekha: I guess that's fine. I don't have a strong theory of where exactly per-layout-object and paint-generic operations go.

pavpanchekha (approved):

pavpanchekha:
Ok, I'm approving this for merge. I continue to worry about atomic inlines—in fact, maybe we have bugs here—but the changes are clearly a net positive for the book.

book/accessibility.md
chrishtr: We did agree last week to add the rect class. :)

book/visual-effects.md
chrishtr: Ok I'll land it now and we can test & make another PR to try out moving is_atomic.

p.s. not sure what you mean by "whose root element is not a block element". Are you referring to it being for an inline element with some blocks inside it?

PR #960, opened 2023-10-21
Major changes to Chapter 7

pavpanchekha:
Here's the scope of the changes:

- Reordered both browser chrome sections quite dramatically. Do note that I used a lot of `newtab_rect[3]` and `back_rect[2]` but the intention is to turn those into `Rect` fields in a later PR.
- Reorder `LineLayout` and `TextLayout`'s `layout` methods so they flow a bit better.
- Explicitly bring up the "reverse pipeline" analogy for click handling
- Make `Browser.active_tab` a `Tab`, not an index. Both are a little dangerous but this is way shorter.
- Moved the layout tree hit testing exercise to a later chapter, where it will do more good.
- Move the HTML chrome exercise to the chapter on inputs and buttons, where it's at least hypothetically possible.
- Renamed `Browser.load` to `Browser.new_tab`, I think the old name was more confusing
- Don't store `previous_word` in `BlockLayout`, just compute it when needed

The tests don't pass yet and I need to do the forward pass through later chapters to make sympathetic changes, but it should be ready for a review of the chapter itself.

### DIFF
--- a/book/embeds.md
+++ b/book/embeds.md
@@ -834,7 +834,7 @@ class Tab:
         if self.needs_accessibility:
             # ...
 
-        if self.pending_hover:
+        if self.needs_paint:
             # ...
 
         # ...
@@ -1258,9 +1258,8 @@ class Browser:
         self.lock.acquire(blocking=True)
         if self.root_frame_focused:
             # ...
-        active_tab = self.tabs[self.active_tab]
-        task = Task(active_tab.scrolldown)
-        active_tab.task_runner.schedule_task(task)
+        task = Task(self.active_tab.scrolldown)
+        self.active_tab.task_runner.schedule_task(task)
         self.lock.release()
 ```
 
@@ -1346,7 +1345,7 @@ in the browser thread, which has limited information:
 
 We'll make a subclass of `AccessibilityNode` to store this information:
 
-``` {.python}
+``` {.python dropline=pass}
 class FrameAccessibilityNode(AccessibilityNode):
     pass
 ```
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -195,14 +195,12 @@ The `Browser` code just delegates to the `Tab`, via a main thread task:
 class Browser:
 	# ...
     def increment_zoom(self, increment):
-        active_tab = self.tabs[self.active_tab]
-        task = Task(active_tab.zoom_by, increment)
-        active_tab.task_runner.schedule_task(task)
+        task = Task(self.active_tab.zoom_by, increment)
+        self.active_tab.task_runner.schedule_task(task)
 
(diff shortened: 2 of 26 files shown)


chrishtr (commented):
These are nice improvements.

pavpanchekha:
In retrospect, this should have been, like, five PRs. I'm still working on getting tests to pass and forward-propagating all changes. Yeesh!

pavpanchekha:
Ok, I think it is done. The main things there were hard were:
- Renaming `Browser.load` to `Browser.new_tab` required changes in every tests file
- Changing `active_tab` to be the `Tab`, not the index, required complex and far-reaching changes in many chapters
- Some of the later chapters haven't yet been updated to use `wbetools.patch` and there are nontrivial differences between chapters where we fixed a bug in one chapter but not the next.
Anyway, I think it's ready for re-review and merging.

book/embeds.md
```diff
@@ -834,7 +834,7 @@ class Tab:
         if self.needs_accessibility:
             # ...
 
-        if self.pending_hover:
+        if self.needs_paint:
```
chrishtr, line 837: Why this change?

chrishtr (approved):

book/embeds.md
pavpanchekha: Yeah, this is a good question. As far as I can tell, this code block was only passing before by accident; `pending_hover` doesn't actually live here.

pavpanchekha: Oh, yeah, right, and it stopped passing because I deleted a bunch of unchanged methods and so it stopped passing by accidentally picking up part of them.

chrishtr: Was one of them actually not changed?

pavpanchekha: Ok, so here's the full story. In Chapter 14, we add handling for `pending_hover`. That code lives in `Browser`. So there's code somewhere in `Browser` (specifically, its `paint_draw_list` method) with the code `if self.pending_hover`. That code works, and it works in the same way in Chapter 15.

At some point in the past, things were different, and `pending_hover` was handled in `Tab`. I have no idea how far back this was—it may even be pre-publication of that chapter. Anyway, at that point, the line `if self.pending_hover` made it into that block in `embeds.md`. Later the pending hover code was moved to `Browser`, but the block in `embeds.md` was never updated and it accidentally kept passing, because it was _somewhere_ in the file, namely inside `Browser.paint_draw_list`.

In this PR, I removed `paint_draw_list` because the patch mechanism is now used. That made the block fail. I fixed it up to match the current actual contents of that method.

In short, this change fixes a bug that was already in `main` but got in the way of actually pushing through this PR.

PR #965, opened 2023-10-23
Apply clips to composited bounds

chrishtr:
I originally avoided this in chapter 13 to reduce complexity, but it turns out to be necessary to get acceptable scrolling
performance for browser.engineering in chapters 15 and 16, due to a huge offscreen piece of an iframe's composited layer.

### DIFF
--- a/book/animations.md
+++ b/book/animations.md
@@ -2010,9 +2010,15 @@ class CompositedLayer:
 The blue square should now be underneath the green square, so overlap
 testing is now complete.[^not-really]
 
-However, this now makes `absolute_bounds`'s size a bit different than
-`composited_bounds`, in situations where there are clips in the visual effect
-path to the root. Let's fix that by also applying those clips to
+There's one more situation worth thinking about, though. Suppose we have
+a huge composited layer, containing a lot of text, except that most of that
+layer is clipped out by a `ClipRRect` somewhere above it. Then the
+`absolute_bounds` consider the clip operations and the `composited_bounds`
+don't, meaning that we'll make a much larger composited layer than necessary
+and waste a lot of time raster pixels that the user will never see. This situation
+actually happens a lot when scrollable elements are drawn to the screen. 
+
+Let's fix that by also applying those clips to
 `composited_bounds`, so that we can properly minimize the amount of raster
 happening in each `CompositedLayer`.[^clipping-notes] We'll do it by first
 computing the absolute bounds for each item, then mapping them back to local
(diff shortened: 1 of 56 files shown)


book/animations.md
```diff
@@ -1974,8 +1974,10 @@ class Transform(VisualEffect):
 
 class ClipRRect(VisualEffect):
     def map(self, rect):
-        bounds = self.rrect.rect()
-        bounds.intersect(rect)
+        bounds = skia.Rect.MakeXYWH(
+            rect.x(), rect.y(),
+            rect.width(), rect.height())
+        bounds.intersect(self.rrect.rect())
```
pavpanchekha, line 1980: I think the first three lines here are just copying `rect`, right? We usually use `.makeOffset(0.0, 0.0)` for that.

book/animations.md
```diff
@@ -2004,13 +2005,73 @@ class CompositedLayer:
...
+space, which will have the effect of computing the "clipped local rect" for
+each display item:
+
+[^clipping-notes]: This is very important, because otherwise some
+composited layers can end up huge despite not drawing much to the screen.
+A good example of this optimization making a big difference is loading the
+browser from [Chapter 15](embeds.md) for the `browser.engineering` homepage,
+where otherwise we would end up with an enormous composited layer for an
+iframe.
+]
```
pavpanchekha, line 2030: Close paren unneeded, I think?

infra/compile.py
```diff
@@ -181,6 +181,8 @@ def check_args(args, ctx):
     "Intersects",
     "join",
     "makeOffset",
+    "x",
+    "y",
```
pavpanchekha, line 185: If we use `makeOffset` we can drop these two

book/animations.md
```diff
@@ -2004,13 +2005,73 @@ class CompositedLayer:
...
+            rect.join(local_to_absolute(item, item.rect))
         return rect
 ```
 
 The blue square should now be underneath the green square, so overlap
 testing is now complete.[^not-really]
 
+However, this now makes `absolute_bounds`'s size a bit different than
+`composited_bounds`, in situations where there are clips in the visual effect
+path to the root. Let's fix that by also applying those clips to
```
pavpanchekha, line 2017:
```suggestion
There's one more situation worth thinking about, though. Suppose we have
a huge composited layer, containing a lot of text, except that most of that
layer is clipped out by a `ClipRRect` somewhere above it. Then the
`absolute_bounds` consider the clip operations and the `composited_bounds`
don't, meaning that we'll make a much larger composited layer than necessary
and waste a lot of time raster pixels that the user will never see. This situation
actually happens a lot when scrollable elements are drawn to the screen. 

Let's fix that by also applying those clips to
```

pavpanchekha (approved):
This is a great change. I suggest adding a bit more explanation, I gave an example but perhaps you'd want to reword it. It might even be worth adding a figure here.

PR #970, opened 2023-10-26
Fix bugs in chapters 11-16

chrishtr:
* Interaction with the address bar was pretty broken, because the code in Browser and Chrome was out of date
* URLs were not being displayed correctly for chapters 15 and 16 after the Chrome class refactor.
* Renamed some wrong JS variable names, and renamed some more for consistency.


pavpanchekha (approved):
Excellent

PR #973, opened 2023-10-27
Multiple changes to Chapter 11

pavpanchekha:
I moved the Z-index exercise to the chapter with `transform`

### DIFF
--- a/book/visual-effects.md
+++ b/book/visual-effects.md
@@ -17,13 +17,15 @@ Installing Skia and SDL
 =======================
 
 Before we get any further, we'll need to upgrade our graphics system.
-While Tkinter is great for basic shapes and handling input, it lacks
+While Tkinter is great for basic shapes and input handling, it lacks
 built-in support for many visual effects.[^tkinter-before-gpu]
 Implementing all details of the web's many visual effects is fun, but
 it's outside the scope of this book, so we need a new graphics library.
 Let's use [Skia][skia],\index{Skia} the library that Chromium uses.
 Unlike Tkinter, Skia doesn't handle inputs or create graphical windows,
-so we'll pair it with the [SDL][sdl] GUI library.\index{SDL}
+so we'll pair it with the [SDL][sdl] GUI library.\index{SDL} Beyond
+new capabilities, switching to Skia will give me an opportunity to
+show you how graphics work at a lower level.
 
 [skia]: https://skia.org
 [sdl]: https://www.libsdl.org/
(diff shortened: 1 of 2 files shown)


book/visual-effects.md
```diff
@@ -274,15 +284,17 @@ documentation for more on the Skia API.
 def parse_color(color):
     if color == "white":
         return skia.ColorWHITE
-    elif color == "lightblue":
-        return skia.ColorSetARGB(0xFF, 0xAD, 0xD8, 0xE6)
-    # ...
+    elif color == "red":
```
chrishtr, line 287: Why?

book/visual-effects.md
```diff
@@ -1149,21 +1171,19 @@ large enough that the border radius is obvious.
...
-boxes, having to do with the possibility of the child content being
-scrolled---`hidden` means "clipped, but might be scrolled by
-JavaScript". If the blue box had not been taller, than it would have
-been impossible to see the text, which is really bad if it's intended
-that there should be a way to scroll it on-screen.
+sharp rounded edge.[^hidden] That's clipping; without the `overflow:
+clip` property these letters would instead be fully drawn, like we saw
+earlier in this chapter.
+
+[^hidden]: The `overflow: hidden` propertyis somewhat similar, but in
```
chrishtr, line 1178:
```suggestion
[^hidden]: The `overflow: hidden` property is somewhat similar, but in
```

book/visual-effects.md
pavpanchekha: Ok, right. We _need_ to add `red` because we use that for the address bar cursor. I guess we don't need to remove `lightblue` but I'm not sure if we use it either (though—we do use it in an example, but that example _also_ uses `lightgreen`...)

pavpanchekha: I'm leaning toward just implementing hex color parsing. That way exercises can use hex codes and we don't need any other keyword colors.

It should be really easy; colors in Skia are literally 32-bit integers so it should be something like:

```
if color.startswith("#") and len(color) == 7:
    r = int(color[1:3], 16)
    g = int(color[3:5], 16)
    b = int(color[5:7], 16)
    return skia.Color(r, g, b)
```

But I think that would be a separate PR, during or after the big reorg.

chrishtr: Ok. for now can you put lightblue back?

pavpanchekha: Done. Can we merge this now?

chrishtr (approved):

PR #974, opened 2023-10-28
Create a new Rect class in lab7

chrishtr:
This avoids passing around 4-tuples.

I also fixed the bugs in Chapters 12-16 which had a thread-unsafe browser url


book/chrome.md
```diff
@@ -930,13 +934,14 @@ class Browser:
...
-rectangle:
+let's add a quick method to test whether a point is cotained in a
+`Rect`:
 
 ``` {.python}
-def intersects(x, y, rect):
-    (left, top, right, bottom) = rect
-    return x >= left and x < right and y >= top and y < bottom
+class Rect:
+    def containsPoint(self, x, y):
```
chrishtr, line 942: I chose a different name from skia.Rect to make the JS easy to compile.

book/chrome.md
```diff
@@ -683,6 +683,18 @@ class Chrome:
...
+browser chrome. For that, a new `Rect` class will be convenient:
+
+``` {.python}
+class Rect:
+    def __init__(self, left, top, right, bottom):
+        self.left = left
+        self.top = top
+        self.right = right
+        self.bottom = bottom
+```
```
pavpanchekha, line 696: Do you want to also ask people to change `DrawRect` to take in a `Rect` argument? We could even ask the reader to replace the `top`/`left`/`right`/`bottom` fields on drawing commands like `DrawText` and `DrawLine`, which we use for clipping, with a per-drawing-command `rect` field. I think this chapter has typically been easy (and the hardest part was visualizing coordinnates, which we've eliminated with auto-sizing) so I'd rather add here to remove from Ch11.

book/visual-effects.md
```diff
@@ -333,6 +333,13 @@ class DrawRect:
         # ...
 ```
 
+Speaking of rects, let's now get rid of the old `Rect` class that was
+introduced in [Chapter 7](chrome.md) in favor of `skia.Rect`. Everywhere
+that a `Rect` was constructed, instead put `skia.Rect.MakeLTRB`, and
+everywhere that the sides of the rectangle (e.g. `left`) where checked,
+replace them with the corresponding function on a Skia `Rect` (e.g. `left()`).
+And replace calls to `containsPoint` with `contains`.
```
pavpanchekha, line 341:
```suggestion
Also replace calls to `containsPoint` with Skia's `contains`.
```

book/visual-effects.md
```diff
@@ -1472,7 +1480,7 @@ Browser compositing
 Optimizing away surfaces is great when they're not needed, but
 sometimes having more surfaces allows faster scrolling and
 animations. (In this section we'll optimize scrolling; animations
-will have to wait for [Chapter 13](animations.md).)
+will have to wait for [Chapter 13](animations.md)).
```
pavpanchekha, line 1483: I don't think this is right, if it's a whole sentence the parens go around it.

src/lab15.py
```diff
@@ -1648,82 +1647,6 @@ def post_message(self, message, target_window_id):
         frame.js.dispatch_post_message(
             message, target_window_id)
 
-@wbetools.patch(Chrome)
-class Chrome:
```
pavpanchekha, line 1652: Was there no change at all in this code? I can't think of any...

pavpanchekha (approved):

PR #976, opened 2023-10-29
Fix PDF build on macOS

pavpanchekha:
This is the BSD `sed` invocation fix from the other branch.


chrishtr:
Still works on linux, so LGTM

PR #977, opened 2023-10-29
Add a discarded field to JSContext

chrishtr:
This is used to avoid async callbacks from running on a detached JSContext.

### DIFF
--- a/book/scheduling.md
+++ b/book/scheduling.md
@@ -338,6 +338,10 @@ class JSContext:
     def __init__(self, tab):
         # ...
         self.discarded = False
+
+    def dispatch_settimeout(self, handle):
+        if self.discarded: return
+        self.interp.evaljs(SETTIMEOUT_CODE, handle=handle)
 ```
 
 ``` {.python}
--- a/book/scheduling.md
+++ b/book/scheduling.md
@@ -236,18 +236,7 @@ all run independently and communicate only via special message-passing APIs.
 
 [workers]: https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API
 
-Let's implement that:[^later-bug]
-
-[^later-bug]: This code has a *very* subtle bug, wherein a page might
-    create a `setTimeout`, and then have that timer trigger later, when
-    a user is visiting another web page. In our browser, that would
-    allow one page to run JavaScript that modifies a different
-    page---a huge security vulnerability! I *think* you can avoid this
-    by resetting `self.js.tab` when you navigate to a new page, but
-    ideally you'd do something more careful, like keeping track of all
-    the child threads spawned by a `JSContext` and ending all of them
-    before navigating. As our browser gets more complex, our bugs, and
-    their associated fixes, get more complex too!
+Let's implement that:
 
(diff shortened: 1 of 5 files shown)


book/scheduling.md
```diff
@@ -339,6 +328,27 @@ That's because after the task has been removed from the queue, it
 can't be accessed by another thread, so the lock does not need to be
 held while the task is running.
 
+The `setTimeout` code is now thread-safe, but still has yet another bug: if we
+navigate from one page to another, `setTimeout` callbacks still pending on the
+previous page eshouldn't do anything. That is easily fixed by adding a
```
pavpanchekha, line 333:
```suggestion
previous page might still try to execute. That is easily prevented by adding a
```

book/scheduling.md
```diff
@@ -339,6 +328,27 @@ That's because after the task has been removed from the queue, it
...
+        self.discarded = False
+```
+
+``` {.python}
+class Tab:
+    def load(self, url, payload=None):
+        # ...
+        if self.js:
+            self.js.discarded = True
+        self.js = JSContext(self)
```
pavpanchekha, line 349:
```suggestion
        self.js = JSContext(self)
        # ...
```

book/scheduling.md
```diff
@@ -339,6 +328,27 @@ That's because after the task has been removed from the queue, it
...
+        # ...
+        self.discarded = False
+```
+
+``` {.python}
+class Tab:
+    def load(self, url, payload=None):
+        # ...
+        if self.js:
+            self.js.discarded = True
```
pavpanchekha, line 348: I would do this on one line but whatever, up to you

book/scheduling.md
```diff
@@ -443,6 +453,7 @@ class JSContext:
         self, method, url, body, isasync, handle):
         # ...
         def run_load():
+            if self.discarded: return
```
pavpanchekha, line 456: I don't think we need / want this here. This is possibly run on another thread, so ideally shouldn't read `self.discarded`. (By the way, pre-existing bug! We should save `self.tab.url` outside `run_load` to avoid cross-thread reads / to avoid a security bug if you navigate before the request thread runs?) And actually issuing the request is probably safe, since the `send` was in fact executed...

book/scheduling.md
```diff
@@ -339,6 +328,27 @@ That's because after the task has been removed from the queue, it
...
+The `setTimeout` code is now thread-safe, but still has yet another bug: if we
+navigate from one page to another, `setTimeout` callbacks still pending on the
+previous page eshouldn't do anything. That is easily fixed by adding a
+`discarded` field on `JSContext` and setting it when loading a new page:
+
+``` {.python}
+class JSContext:
+    def __init__(self, tab):
+        # ...
+        self.discarded = False
```
pavpanchekha, line 340: Wait, don't we also need to check `discarded` in `dispatch_setTimeout`? I think that's the whole point!

pavpanchekha (approved):

pavpanchekha:
Please resolve comments but then go ahead & merge.

PR #978, opened 2023-10-29
Change is_atomic to should_paint

chrishtr:


src/lab10.py
```diff
@@ -13,13 +13,14 @@
...
-from lab5 import BLOCK_ELEMENTS, DrawRect, DocumentLayout
+from lab5 import BLOCK_ELEMENTS, DrawRect
 from lab6 import CSSParser, TagSelector, DescendantSelector
 from lab6 import INHERITED_PROPERTIES, style, cascade_priority
 from lab6 import DrawText, tree_to_list
-from lab7 import DrawLine, DrawOutline, LineLayout, TextLayout
+from lab7 import DrawLine, DrawOutline
 from lab8 import URL, Browser, Text, Element
 from lab8 import BlockLayout, InputLayout, INPUT_WIDTH_PX
+from lab8 import DocumentLayout, LineLayout, TextLayout
```
pavpanchekha, line 23: Why remove these? Is this something in your editor that removes unused imports?

pavpanchekha (approved):

pavpanchekha:
OK, so this is a net improvement to the current code, so I'm willing to merge, but in my dream version of this code, there's no method in `BlockLayout` at all, because I think the code should be something like:

```
def paint_tree(layout_object, display_list):
    should_paint = \
      not isinstance(layout_object, BlockLayout) or
      isinstance(layout_object.node, Element) and
      layout_object.node.tag in BLOCK_ELEMENTS
    if should_paint: # ...
```

Here's my logic. The issue only applies to BlockLayout. The issue is that some inline element may have a BlockLayout created pointing to it, instead of an anonymous block box pointing to some fake wrapper element that then contains the inline element. So let's just skip any BlockLayout that points to a non-block element. This avoids messing with the list of tag names, and I think handling this in `paint_tree` is cleaner than adding a method to the layout object API.

src/lab10.py
chrishtr: I removed them because they are not used. Is this another case where you think something or other will be broken - the outline maybe?

pavpanchekha: Yeah. I promise I will fix this to not be necessary once we ship the book text!

chrishtr:
> Here's my logic. The issue only applies to BlockLayout. The issue is that some inline element may have a BlockLayout created pointing to it, instead of an anonymous block box pointing to some fake wrapper element that then contains the inline element. So let's just skip any BlockLayout that points to a non-block element. This avoids messing with the list of tag names, and I think handling this in `paint_tree` is cleaner than adding a method to the layout object API.

Isn't it an anti-pattern to do type inspection unless you have to? Also:
* I'm not sure your proposed change makes a difference to readability of the book
* Putting the code close to `BlockFlow` seems logical to me pedagogically

pavpanchekha:
Ok, let's split my critique into three parts:

- The existing code and also your version allow painting a `BlockLayout` that points at a `Text` node. I think this is harmless but not painting it would also be harmless. I'm not sure why it's explicitly mentioned.
- The explicit list of tag names makes me worry that this is some sort of super weird hack. In fact the important property is `not in BLOCK_ELEMENTS`, not `in ["input", "button", "iframe", "img"]`, even if in practice both of those have similar results.[1] Moreover, I think my replacement property highlights the relationship to `layout_mode`, which also tests against `BLOCK_ELEMENTS`. That's correct, because this code is really a hack to work around a bug in `layout_mode`.
- Whether it's a method of all layout modes or a line in `paint_tree` is, I think, a matter of taste, and I'm willing to go your way. I agree that putting `should_paint` in `BlockLayout` is pedagogically nice, but it also requires adding `should_paint` in every other `XXXLayout`, and that's not so nice, so I personally am willing to take the type test. But this is subjective, happy to go either way.

[1] Not the same, because our current codebase doesn't even attempt to handle effects, opacity, or backgrounds for non-atomic inline elements. Like, not at all. Hard to say, given that, whether a non-atomic inline element inside an anonymous block box which has an effect or a background should or should not be painted; my proposed rewrite differs from current in that case.

src/lab10.py
chrishtr: Is it easy to thrown an error if this doesn't work in the inliner? How do we know we haven't missed other cases?

chrishtr: oh now I remember. I removed some from lab5 because lab8 now overrides them. I
did not in fact remove anyy imports.

PR #984, opened 2023-10-31
Address some feedback from chapter 12

chrishtr:
* Added text about `render` at the start of `click`, and about the difference between `render` and `run_animation_frame`
* Clarified text about starting the event loop
* Fix lint (still not working correctly on github actions...)

### DIFF
--- a/book/scheduling.md
+++ b/book/scheduling.md
@@ -689,10 +689,22 @@ class JSContext:
         self.tab.set_needs_render()
 ```
 
-There are more calls to `render`; you should find and fix all of them.
-However, there is one call to `render` at the beginning of `click` that
-should remain, because it's necessary to make sure that layout is up-to-date
-before finding which layout object is under the click.
+There are more calls to `render`; you should find and fix all of them...except
+`click`. Here's why:
+
+Now we don't immediately render when something changes. That means that the
+layout tree (and style) could be out of date when a method is called. Normally,
+this isn't a problem, but in one important case it is: click handling. That's
+because we need to read the layout tree to figure out what object was clicked
+on, which means the layout tree needs to be up to date. Add a call to render at
+the top of click:
+
+``` {.python}
+class Tab:
+    def click(self, x, y):
+        self.render()
+        # ...
+```
 
 Another problem with our implementation is that the browser is now
 doing `raster_and_draw` every time the active tab runs a task.
--- a/book/scheduling.md
+++ b/book/scheduling.md
@@ -1400,11 +1400,11 @@ class CommitData:
 When running an animation frame, the `Tab` should construct one of
 these objects and pass it to `commit`. To keep `render` from getting
 too confusing, let's put this in a new `run_animation_frame` method,
-and move `__runRAFHandlers` there too.^[I'm not just adding all this
-to `render`, because `render` is just about updating style, layout and paint
-when needed. That's why it is called from the beginning `click` instead of
-`run_animation_frame`. The latter also has script callbacks and also commits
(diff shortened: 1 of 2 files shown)


book/scheduling.md
```diff
@@ -698,6 +690,9 @@ class JSContext:
 ```
 
 There are more calls to `render`; you should find and fix all of them.
+However, there is one call to `render` at the beginning of `click` that
+should remain, because it's necessary to make sure that layout is up-to-date
+before finding which layout object is under the click.
```
pavpanchekha, line 695: I think this still isn't right. I went back through and Ch 9 is the last one that touches `click` (it adds JS event dispatching). But Ch9 doesn't render at the top of `click`—because in Chapter 9 we eagerly render any time we change, so there's no need to do it at the top of `click`. I think this paragraph should be something like:

> Now we don't immediately render on change. That means that the layout tree (and style) could be out of date when a method is called. Normally, this isn't a problem, but in one important case it is: click handling. That's because we need to *read* the layout tree to figure out what object was clicked on, which means the layout tree needs to be up to date. Add a call to `render` at the top of `click`:
> ... code block here

book/scheduling.md
```diff
@@ -1393,7 +1388,11 @@ class CommitData:
 When running an animation frame, the `Tab` should construct one of
 these objects and pass it to `commit`. To keep `render` from getting
 too confusing, let's put this in a new `run_animation_frame` method,
-and move `__runRAFHandlers` there too:
+and move `__runRAFHandlers` there too.^[I'm not just adding all this
+to `render`, because `render` is just about updating style, layout and paint
+when needed. That's why it is called from the beginning `click` instead of
+`run_animation_frame`. The latter also has script callbacks and also commits
+to the browser, which don't make sense to have happen from `click`.]
```
pavpanchekha, line 1395: This is a fantastic reason.[1]

[1] Unrelated, but I found it funny that we made the right architecture decision to account for a control flow path that our browser, due to a bug, actually didn't have!

pavpanchekha:
```suggestion
and move `__runRAFHandlers` there too.^[The `render` method is just about
updating style, layout and paint when needed; it's called for every frame,
but it's also called from `click`, and in real browsers from many other places too.
Meanwhile, `run_animation_frame` is only called for frames, and therefore
it, not `render`, runs RAF handlers and calls `commit`.]
```

Rephrase.

pavpanchekha (changes requested):
Please see note about `render`. It's not just that we *keep* a call; we never had it and never add it. (Eager rendering is such a _crazy_ way to write a browser that it's hard to believe `click` didn't trigger `render` in prior chapters...)

chrishtr:
Done, please take another look.

book/scheduling.md
```diff
@@ -697,7 +689,22 @@ class JSContext:
         self.tab.set_needs_render()
 ```
 
-There are more calls to `render`; you should find and fix all of them.
+There are more calls to `render`; you should find and fix all of them...except
+`click`. Here's why:
```
pavpanchekha, line 693:
```suggestion
There are more calls to `render`; you should find and fix all of them...
Except, let's take a closer look at `click`.
```

book/scheduling.md
```diff
@@ -697,7 +689,22 @@ class JSContext:
...
+
+Now we don't immediately render when something changes. That means that the
```
pavpanchekha, line 695:
```suggestion
We now don't immediately render when something changes. That means that the
```

pavpanchekha (approved):

PR #985, opened 2023-10-31
Add a paragraph to the preface calling out that software has bugs

chrishtr:


pavpanchekha:
You know, the text is fine, but I'm kind of against adding this. I think it's fair to say there might be bugs, where to report them, and where to get an up-to-date copy of the book/code, but I think we should stand behind our code more than this paragraph does! And, like, set a high bar for ourselves.

pavpanchekha (commented):
(see comment above)

pavpanchekha (changes requested):
(see comment two above—GH interface is confusing me.)

chrishtr:
ok. Closing then.

PR #987, opened 2023-11-01
Chapter 7 small editorial changes

chrishtr:


book/chrome.md
```diff
@@ -122,9 +122,15 @@ text = TextLayout(node, word, line, previous_word)
...
-later.
+Note that I needed a new parameter here, `previous_word`, to keep track of
+the previous word in the current line. So we'll need to initialize it:
+
+``` {.python}
+class TextLayout:
+    def __init__(self, node, word, parent, previous):
+        # ...
+        self.previous = previous
+```
```
pavpanchekha, line 133: Oh no! This change is bad but so is the original. That whole paragraph about `previous_word` is now defunct and we should just delete the whole thing. That's why you probably found that there was no code block initializing `previous_word`, but the code-block you added isn't related!

Longer story: we used to store a `previous_word` field, but then I changed it to just be a `previous_word` local variable. Now that it's not a field, you don't need to initialize it in the constructor. Originally I did a field because I thought those links had to go across lines. But that's wrong; it should be the previous word in the line. So we can just compute `previous_word` locally by looking at `line` instead of storing a field.

pavpanchekha (changes requested):
You found an important bug! (But it needs to be fixed a different way.)

chrishtr:
I deleted that whole paragraph, LGTY now?

pavpanchekha (approved):

PR #988, opened 2023-11-01
Editorial edits for chapter 8, and focus cleanup

chrishtr:
* There is a focus field on both Browser and Chrome. The former can be set only to "content" and the latter only to "address book".
* Schedule render on keypress and enter for chapters 12+ (bugfix I found while fixing focus).
* Fixed one broken `self_rect` method.

### DIFF
--- a/book/forms.md
+++ b/book/forms.md
@@ -398,33 +398,35 @@ remember that in its own `focus` field!
 
 In other words, when you click on the web page, the `Browser` updates
 its `focus` field to remember that the user is interacting with the
-page, not the browser interface:
+page, not the browser chrome:
 
 ``` {.python}
 class Browser:
     def handle_click(self, e):
         if e.y < self.chrome.bottom:
-            self.focus = "chrome"
+            self.focus = None
             # ...
         else:
             self.focus = "content"
+            self.chrome.focus = None
             # ...
         self.draw()
 ```
 
-The `if` branch that corresponds to clicks in the browser interface
-unsets `focus` by default, but some existing code in that branch will
-set `focus` to `"address bar"` if the user actually clicked in the
-address bar.
+The `if` branch that corresponds to clicks in the browser chrome
+unsets `focus`, but some existing code in that branch will
+set the `Chrome` `focus` to `"address bar"` if the user actually clicked
+in the address bar.
 
 When a key press happens, the `Browser` sends it either to the address
-bar or calls the active tab's `keypress` method:
+bar or calls the active tab's `keypress` method (or neither, if nothing is
+focused):
 
 ``` {.python}
 class Browser:
     def handle_key(self, e):
         # ...
-        if self.focus == "chrome":
+        if self.chrome.focus:
             self.chrome.keypress(e.char)
             self.draw()
         elif self.focus == "content":
--- a/book/forms.md
+++ b/book/forms.md
@@ -783,8 +783,7 @@ import socket
 s = socket.socket(
     family=socket.AF_INET,
     type=socket.SOCK_STREAM,
-    proto=socket.IPPROTO_TCP,
-)
+    proto=socket.IPPROTO_TCP)
 s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
 ```
 
--- a/book/forms.md
+++ b/book/forms.md
@@ -414,9 +414,8 @@ class Browser:
 ```
 
 The `if` branch that corresponds to clicks in the browser chrome
-unsets `focus`, but some existing code in that branch will
-set the `Chrome` `focus` to `"address bar"` if the user actually clicked
-in the address bar.
+unsets `focus`, meaning focus is no longer on the page contents,
+and key presses will thus be sent to the `Chrome`.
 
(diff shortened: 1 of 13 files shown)


book/forms.md
```diff
@@ -398,33 +398,35 @@ remember that in its own `focus` field!
...
 ```
 
-The `if` branch that corresponds to clicks in the browser interface
-unsets `focus` by default, but some existing code in that branch will
-set `focus` to `"address bar"` if the user actually clicked in the
-address bar.
+The `if` branch that corresponds to clicks in the browser chrome
+unsets `focus`, but some existing code in that branch will
+set the `Chrome` `focus` to `"address bar"` if the user actually clicked
+in the address bar.
```
pavpanchekha, line 419:
```suggestion
The `if` branch that corresponds to clicks in the browser chrome
unsets `focus`, meaning focus is no longer on the page contents,
and key presses will thus be sent to the `Chrome`.
```

book/forms.md
```diff
@@ -398,33 +398,35 @@ remember that in its own `focus` field!
...
-bar or calls the active tab's `keypress` method:
+bar or calls the active tab's `keypress` method (or neither, if nothing is
+focused):
 
 ``` {.python}
 class Browser:
     def handle_key(self, e):
         # ...
-        if self.focus == "chrome":
+        if self.chrome.focus:
```
pavpanchekha, line 429: I say this should instead be `if self.focus == None`, and then let the `Chrome` sort out whether anything inside the chrome is focused. I'm trying to preserve some notion of hierarchy here.

book/scheduling.md
```diff
@@ -741,12 +741,12 @@ class Browser:
             self.set_needs_raster_and_draw()
 
     def handle_key(self, char):
-        if self.focus == "chrome":
+        if self.chrome.focus:
```
pavpanchekha, line 744: Same `self.focus == None` here.

book/scheduling.md
```diff
@@ -741,12 +741,12 @@ class Browser:
...
             # ...
             self.set_needs_raster_and_draw()
 
     def handle_enter(self):
-        if self.focus == "chrome":
+        if self.chrome.focus:
```
pavpanchekha, line 749: Same `self.focus == None` here.

book/scheduling.md
```diff
@@ -1340,7 +1340,7 @@ The same logic holds for `keypress`:
 class Browser:
     def handle_key(self, char):
         if not (0x20 <= ord(char) < 0x7f): return
-        if self.focus == "chrome":
+        if self.chrome.focus:
```
pavpanchekha, line 1343: Same `self.focus == None` here.

src/lab11.py
```diff
@@ -539,6 +539,7 @@ def click(self, x, y):
             self.browser.active_tab.go_back()
         elif self.address_rect.contains(x, y):
             self.focus = "address bar"
+            print('set focus')
```
pavpanchekha, line 542: Woops!

src/lab11.py
```diff
@@ -604,7 +606,7 @@ def handle_click(self, e):
 
     def handle_key(self, char):
         if not (0x20 <= ord(char) < 0x7f): return
-        if self.focus == "chrome":
+        if self.chrome.focus:
```
pavpanchekha, line 609: Same `self.focus == None` here.

src/lab11.py
```diff
@@ -587,12 +588,13 @@ def __init__(self):
...
     def handle_click(self, e):
         if e.y < self.chrome.bottom:
-            self.focus = "chrome"
+            self.focus = None
             self.chrome.click(e.x, e.y)
             self.raster_chrome()
         else:
             if self.focus != "content":
                 self.focus = "content"
+                self.chrome.focus = None
```
pavpanchekha, line 597: Ooh. Does this actually mean we need to add blur methods to Chrome? Do we correctly handle clicking on the address bar while editing an input element? Or the reverse?

src/lab11.py
```diff
@@ -614,7 +616,7 @@ def handle_key(self, char):
             self.draw()
 
     def handle_enter(self):
-        if self.focus == "chrome":
+        if self.chrome.focus:
```
pavpanchekha, line 619: Same `self.focus == None` here.

pavpanchekha (commented):
Ok, I think the simplification is good, there's no reason from browser focus to have three states. But I'm still not sure we're handling focus correctly, especially in cases where you are editing the address bar / an input and then click in the other. Can we add tests for this? Also I think it would be neat if we could maintain hierarchy, that is, `Browser` never looks at `Chrome`'s `focus` field. I think that should be possible. It might require adding `blur` methods (maybe we call it `unfocus`) but those might be necessary anyway because I think there might be bugs.

If you'd like, I'm fine merging this and doing the above in a separate PR, or just doing it in this PR.

book/forms.md
chrishtr: I'll fix this in a followup.

src/lab11.py
chrishtr: I tested it pretty carefully and it works. I added a unittest though just now to further ensure correctness.

PR #991, opened 2023-11-03
Address feedback from previous PR

chrishtr:
Don't access `self.chrome.focus`, instead add `blur()` and return values from event handlers

### DIFF
--- a/book/forms.md
+++ b/book/forms.md
@@ -396,9 +396,15 @@ handled by the `Browser`. So how does the `Browser` even know when
 keyboard events should be sent to the `Tab`? The `Browser` has to
 remember that in its own `focus` field!
 
-In other words, when you click on the web page, the `Browser` updates
-its `focus` field to remember that the user is interacting with the
-page, not the browser chrome:
+In other words, when you click on the web page, the `Browser` updates its
+`focus` field to remember that the user is interacting with the page, not the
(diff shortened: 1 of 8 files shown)


book/forms.md
```diff
@@ -425,14 +433,24 @@ focused):
...
 
+Here I've changed `keypress` to return true if the browser chrome consumed the
+key:
+
+``` {.python}
+class Chrome:
+    def keypress(self, char):
+        if self.focus == "address bar":
+            self.address_bar += char
+        return self.focus == "address bar"
```
pavpanchekha, line 451: I know it's a line longer but I think this would be cleaner if we just did `return True` inside the `if` statement and `return False` outside it.

book/scheduling.md
```diff
@@ -741,16 +741,27 @@ class Browser:
...
             self.set_needs_raster_and_draw()
 
     def handle_enter(self):
-        if self.chrome.focus:
+        if self.chrome.enter():
             # ...
             self.set_needs_raster_and_draw()
 ```
 
+Here I need a small change to make `enter` return whether something was done:
```
pavpanchekha, line 754: I don't think I see any return value here...

pavpanchekha: Though the lab is in fact correctly updated.

pavpanchekha (approved):
This isn't quite what I expected, but it's a good change.

PR #994, opened 2023-11-03
Address feedback in one exercise in chapter 7

chrishtr:


pavpanchekha (approved):
This could also be main text.

PR #998, opened 2023-11-05
Chapter 11 reorganization, phase 1

pavpanchekha:
This "Phase 1" focuses the "port to Skia" sections around surfaces and moves composited scrolling way earlier to reinforce the focus on surfaces.

The best way to review this diff might be to read the actual new text, which I've pushed here: https://browser.engineering/draft/visual-effects.html

### DIFF
--- a/book/visual-effects.md
+++ b/book/visual-effects.md
@@ -551,11 +551,236 @@ high-density enough to retire these techniques.
 
 [font-hinting]: https://en.wikipedia.org/wiki/Font_hinting
 
+
+
+Browser compositing
+===================
+
+So far, Skia and SDL have just made our browser more complex, but the
+low-level control offered by these libraries is important because it
+allows us to optimize commonly-used operations like scrolling.
+
+So far, any time the user scrolled a web page, we had to clear
+the canvas and re-raster everything on it from scratch. This is
+inefficient---we're drawing the same pixels, just in a different
+place. When the context is complex or the screen is large,
+rastering too often produces a visible slowdown and drains laptop and mobile
+batteries. Real browsers optimize scrolling
+using a technique I'll call *browser compositing*:
+drawing the whole web page to a hidden surface, and only copying the
+relevant pixels to the window itself.
+
+To implement this, we'll need two new Skia surfaces:
+a surface for browser chrome and a surface
+for the current `Tab`'s contents. We'll only need to
+re-raster the `Tab` surface if page contents change, but not when
+(say) the user types into the address bar. And we can scroll the `Tab`
+without any raster at all---we just copy a different part of the
+current `Tab` surface to the screen. Let's call those surfaces
+`chrome_surface` and `tab_surface`:[^multiple-tabs]
+
+[^multiple-tabs]: We could even use a different surface for each `Tab`,
+but real browsers don't do this, since each surface uses up a lot of
+memory, and typically users don't notice the small raster delay when
+switching tabs.
+
+``` {.python}
+class Browser:
+    def __init__(self):
+        # ...
+        self.chrome_surface = skia.Surface(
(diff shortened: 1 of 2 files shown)


pavpanchekha:
I anticipate also doing a "Phase 2" Chapter 11 reorganization, which will focus on breaking up the theory sections, but I think this "Phase 1" is worth doing standalone.

book/visual-effects.md
```diff
@@ -519,11 +551,236 @@ high-density enough to retire these techniques.
...
+
+
+Browser compositing
+===================
+
+So far, Skia and SDL have just made our browser more complex, but the
+low-level control offered by these libraries is important because it
+allows us to optimize commonly-used operations like scrolling.
+
+So far, any time the user scrolled a web page, we had to clear
```
chrishtr, line 563:
```suggestion
Any time the user scrolled a web page, we had to clear
```

book/visual-effects.md
```diff
@@ -519,11 +551,236 @@ high-density enough to retire these techniques.
...
+```
+
+I'm not explicitly creating `tab_surface` right away, because we need
+to lay out the page contents to know how tall the surface needs to be.
+
+We'll also need to split the browser's `draw` method into three parts:
+
+- `draw` will composite the chrome and tab surfaces and copy the
+  result from Skia to SDL;[^why-two-steps]
+- `raster_tab` will draw the page to the `tab_surface`; and
```
chrishtr, line 603:
```suggestion
- `raster_tab` will raster the page to the `tab_surface`; and
```

book/visual-effects.md
```diff
@@ -519,11 +551,236 @@ high-density enough to retire these techniques.
...
+- `raster_chrome` will draw the browser chrome to the `chrome_surface`.
```
chrishtr, line 604:
```suggestion
- `raster_chrome` will raster the browser chrome to the `chrome_surface`.
```

book/visual-effects.md
```diff
@@ -519,11 +551,236 @@ high-density enough to retire these techniques.
...
+a very large amount of memory. We'll ignore that, but a real browser
+would only paint and raster surface content up to a certain distance
+from the visible region, and re-paint/raster as the user scrolls.
+
+``` {.python}
+import math
+
+class Browser:
+    def raster_tab(self):
+        tab_height = math.ceil(self.active_tab.document.height + 2*VSTEP)
```
chrishtr, line 646: Add newlines to avoid a horizontal scrollbar here.

book/visual-effects.md
```diff
@@ -519,11 +551,236 @@ high-density enough to retire these techniques.
...
+            0, 0, WIDTH, self.chrome.bottom)
+        canvas.save()
+        canvas.clipRect(chrome_rect)
+        self.chrome_surface.draw(canvas, 0, 0)
+        canvas.restore()
+
+        # ...
+```
+
+Note the `draw` calls: we are drawing the `tab_surface` and
```
chrishtr, line 690: Some of this is redundant with the previous paragraph.

chrishtr (approved):
Thank you for breaking this into multiple PRs.

book/visual-effects.md
pavpanchekha: I think the "so far" is important because the rest of the sentence is in past tense.

chrishtr: Fair. I just didn't like how it repeated the exact same starting words as the previous paragraph. Is there a way to rephrase?

pavpanchekha: Ah! Previous paragraph can lose its "so far" instead.

book/visual-effects.md
pavpanchekha: Shortened.

PR #1000, opened 2023-11-05
Remove some redundant text from chapter 11

chrishtr:


pavpanchekha:
Thanks for finding this. I have a slightly different fix for this, moving the longer explanation of stroke down after the rect stuff.

PR #1003, opened 2023-11-06
Move layout-inducing animation parts to commentary on the width exercise

chrishtr:


book/animations.md
```diff
@@ -2188,7 +2164,30 @@ the string). Since `width` and `height` are layout-inducing, make sure that
...
+compositing, they're usually not a good idea on the web. Not only are they
+slower, but they also cause page elements to jump around suddenly and
+don't create that illusion of continuous movement. An
+exception is resizing the browser window with your mouse. That's
+layout-inducing, but it's very useful for the user to see the new layout
+as the window size changes. Modern browsers are fast enough to do this, but it
+used to be that they'd only redraw the screen every couple of frames,
+leaving a visual *gutter* between content and the edge of the window.
+
+[width-transition]: examples/example13-width-transition.html
```
pavpanchekha, line 2189:
```suggestion
[^note-layout-animations]: Width animations can't be composited because
width affects the layout tree, not just different display lists,
which in turn means that draw commands, not just visual effects,
change. Such animations are called *layout-inducing*, and
they are therefore slower and typically not a good idea.[^resize-window]
[Chapter 16](invalidation.md) will look at one way to speed them up somewhat.

[^resize-window]: One exception is resizing the browser window with your mouse. That's
layout-inducing, but it's very useful for the user to see the new layout
as the window size changes. Modern browsers are fast enough to do this, but it
used to be that they'd only redraw the screen every couple of frames,
leaving a visual *gutter* between content and the edge of the window.
```

pavpanchekha: Rewrote this to integrate with the exercise that came before and also to do two shorter footnotes instead of one longer one.

PR #1004, opened 2023-11-06
Add a scroll up listener alongside the scroll down key listener

Paganiniana:
Salutations! 

I was just playing with the chapter 5 code and noticed that there was no way to "scroll" back up through a page. I'm not sure if this is a desirable feature or not, but here it is!

### DIFF
--- a/src/lab2.py
+++ b/src/lab2.py
@@ -53,6 +53,7 @@ class Browser:
 
         self.scroll = 0
         self.window.bind("<Down>", self.scrolldown)
+        self.window.bind("<Up>", self.scrollup)
 
     def load(self, url):
         body = url.request()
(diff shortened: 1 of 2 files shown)


pavpanchekha:
Thanks, @Paganiniana! This is actually one of the exercises in that chapter, so it is desirable but intended for the reader to write on their own, so I'm going to have to deny the pull request, but you did implement it correctly!

Paganiniana:
🤦‍♂️ Yep, there it is, clear as day, in the _MouseWheel_ exercise. My bad! I was trying to compare rendering between the `src/lab5.py` file and my TypeScript repo and was confused at why the up arrow key wasn't working. 😅 I'll do my best to check that section before future PRs!

PR #1006, opened 2023-11-07
De-duplicate code from lab13, lab14 and lab15

chrishtr:
In the process I found a few discrepancies that are now fixed, and also back-ported one improvement to lab9,
and found and fixed bugs in lab11 and lab12.


chrishtr:
In the interests of time, going to go ahead and land this.

PR #1007, opened 2023-11-07
Improve naming and code structure around clone_latest

chrishtr:


src/lab13.py
```diff
@@ -1339,14 +1339,15 @@ def composite(self):
                 max(self.active_tab_height,
                     layer.absolute_bounds().bottom())
 
-    def clone_latest(self, visual_effect, current_effect):
-        node = visual_effect.node
+    def clone_latest(self, parent_effect, child_effect):
```
pavpanchekha, line 1342: Should we maybe name it `old_parent_effect`? Willing to merge as is too.

pavpanchekha (approved):
I'm not sure changing the `children` parameter to `child` is good, we might end up changing it back if we do the atomic draw-display-list, but the cleanup to names is good.

src/lab13.py
chrishtr: I like parent_effect. "old" people might think of as "from previous frame"

chrishtr:
FYI I tried the atomic thing, and I found:
* it was harder to understand IMO than anticipated
* It also had visual artifacts in the text I couldn't immediately figure out - something to do with drawing two textures on top of each other. I now recall I tried it way back when and found the same result. I'd rather not spend time on it at this point, so recommend we do not do it.

PR #1011, opened 2023-11-09
Fix contrast and a bug in chapter 14

chrishtr:
* parse_color needs to be patched so that it always gets called, even in code not imported into lab14.
* Adjust the background color buttons to ensure a WCAG AAA contrast rating.


pavpanchekha (commented):
Uhh, ok, I think the different color is obviously an improvement, but changing the definition of `orangered` is crazy! I am basically convinced that we need to implement proper hex colors and then we won't need to do this.

chrishtr:
Fixed. Please re-review.

pavpanchekha (approved):
Ok, let's merge this, I'll fix up the text if hex colors are done in Chapter 11 instead.

PR #1014, opened 2023-11-09
Editorial fixes in chapter 12

chrishtr:


book/scheduling.md
```diff
@@ -510,9 +510,6 @@ function __runXHROnload(body, handle) {
 }
 ```
 
-As you can see, tasks allow not only the browser but also applications
-running in the browser to delay tasks until later.
-
```
pavpanchekha, line 515: Ah, the reason for this paragraph is that, I think everywhere, we never end a section with a code block. Not actually sure that this rule matters but that's what sometimes creates these sort-of useless end section paragraphs. Often I fix these by pulling forward the first paragraph of the next section, though that doesn't always work.

book/scheduling.md
```diff
@@ -689,15 +686,15 @@ class JSContext:
         self.tab.set_needs_render()
 ```
 
-There are more calls to `render`; you should find and fix all of them...
-Except, let's take a closer look at `click`.
+There are more calls to `render`; you should find and fix all of
+them...except, let's take a closer look at `click`.
```
pavpanchekha, line 690: Haha, I think I just changed this the other way. That's fine, let's do it your way and see if the copyeditor complains.

book/scheduling.md
```diff
@@ -985,11 +982,11 @@ if our browser consistently runs slower than 60 frames per second, we
...
 ::: {.further}
-Before the `requestAnimationFrame` API, developers used a self-scheduling
-timer instead, via `setTimeout`. This did run animations at a
+Before the `requestAnimationFrame` API, developers approximated it with
+`setTimeout`. This did run animations at a
 (roughly) fixed cadence, but because it didn't line up with the
-browser's rendering loop, events would sometimes be handled between the callback and
-rendering, which might force an extra, unnecessary rendering step.
+browser's rendering loop, events would sometimes be handled between the
+callback an rendering, which might force an extra, unnecessary rendering step.
```
pavpanchekha, line 989:
```suggestion
callback and rendering, which might force an extra, unnecessary rendering step.
```

book/scheduling.md
```diff
@@ -1180,7 +1177,7 @@ Two threads
 ===========
 
 Well, one option, of course, is optimizing raster-and-draw, or even
-render. And we'll do that in the [next chapter][animations.md] But
+render. And we'll do that in the [next chapter](animations.md) But
```
pavpanchekha, line 1180: lol

pavpanchekha (approved):

PR #1015, opened 2023-11-10
Editorial changes from a read-through of chapter 13

chrishtr:
* Fixed a number of typos
* Improved some awkward sentences
* Removed some obsolete text and footnotes
* Simplified the animation code because we now only have opacity
* Removed a non-animations exercise, and a redundant one
* Renamed DrawCommand to PaintCommand
* Fixed inline examples for transform to skip width and height, which we don't support


pavpanchekha (approved):
Very nice simplification!

PR #1019, opened 2023-11-11
Fix bugs in `source_over` and `destination_in` definitions

pavpanchekha:
This PR just fixes those methods and the associated text & tests.

### DIFF
--- a/book/visual-effects.md
+++ b/book/visual-effects.md
@@ -1248,7 +1248,7 @@ however.)
 ``` {.python file=examples}
 class Pixel:
     def source_over(self, source):
-        new_a = 1 - (1 - source.a) * (1 - self.a)
+        new_a = source.a + self.a * (1 - source.a)
         if new_a == 0: return self
         self.r = \
             (self.r * (1 - source.a) * self.a + \
--- a/book/visual-effects.md
+++ b/book/visual-effects.md
@@ -1245,7 +1245,6 @@ generally does not use premultiplied representations, and the code below
 doesn't either. (Skia does represent colors internally in a premultiplied form,
 however.)
 
-
 ``` {.python file=examples}
 class Pixel:
     def source_over(self, source):
--- a/book/visual-effects.md
+++ b/book/visual-effects.md
@@ -1249,17 +1249,18 @@ however.)
 ``` {.python file=examples}
 class Pixel:
     def source_over(self, source):
-        self.a = 1 - (1 - source.a) * (1 - self.a)
-        if self.a == 0: return self
+        new_a = 1 - (1 - source.a) * (1 - self.a)
+        if new_a == 0: return self
         self.r = \
             (self.r * (1 - source.a) * self.a + \
-                source.r * source.a) / self.a
+                source.r * source.a) / new_a
         self.g = \
             (self.g * (1 - source.a) * self.a + \
-                source.g * source.a) / self.a
+                source.g * source.a) / new_a
         self.b = \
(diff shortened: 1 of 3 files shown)


book/visual-effects.md
```diff
@@ -1493,10 +1493,6 @@ fits perfectly. In code, destination-in looks like this:
 class Pixel:
     def destination_in(self, source):
         self.a = self.a * source.a
-        if self.a == 0: return self
-        self.r = (self.r * self.a * source.a) / self.a
```
chrishtr, line 1497: This line simplifies to self.r = self.r * source.a, not a no-op...

src/lab11-tests.md
```diff
@@ -154,7 +157,7 @@ destination-in compositing yields the backdrop.
 But transparency multiplies.
 
     >>> red_opaque.copy().destination_in(blue_semitransparent)
-    Pixel(0.5, 0.0, 0.0, 0.5)
+    Pixel(1.0, 0.0, 0.0, 0.5)
```
chrishtr, line 160: And my comment above leads me to conclude the red channel should be 0.5.

book/visual-effects.md
chrishtr: the link here seems to agree with what I said:

https://drafts.fxtf.org/compositing-1/#porterduffcompositingoperators_dstin

chrishtr: Note that this only makes a difference if the alpha of the source is something other than 1 or 0, which we don't use in this chapter. But still.

src/lab11-tests.md
pavpanchekha: How does that make sense? Destination in blending doesn’t make the color darker!

book/visual-effects.md
pavpanchekha: On line 1495 that should be new_a, not self.a, then it does cancel. Standard maybe different because premultiplied.

src/lab11-tests.md
chrishtr: I guess it does if the destination has opacity?

We could test this with a page that uses opacity on the white rounded clipping rectangle and see what it does in SKia.

pavpanchekha: Ok, so here's my test Skia program: https://fiddle.skia.org/c/2da37d5e990775229a208b41bdace46f

It draws an opaque red rectangle and then draws a semitransparent blue circle over it with DstIn. The resulting overlapped pixels are `rgb(1, 0, 0, .5)`. So I think the PR is correct, the original test was wrong.

pavpanchekha:
Hi Chris, current status of the conversation is that I am still convinced that the changes in this PR are correct and the current code + tests are incorrect. Not really a "re-review" situation but please examine arguments above and see if we can agree. I do think it's important to get this right, and it doesn't really conflict with any other code so we can take our time.

book/visual-effects.md
```diff
@@ -1245,21 +1245,21 @@ generally does not use premultiplied representations, and the code below
 doesn't either. (Skia does represent colors internally in a premultiplied form,
 however.)
 
-
 ``` {.python file=examples}
 class Pixel:
     def source_over(self, source):
-        self.a = 1 - (1 - source.a) * (1 - self.a)
-        if self.a == 0: return self
+        new_a = 1 - (1 - source.a) * (1 - self.a)
```
chrishtr, line 1251: Clean this up to not have so much subtraction

chrishtr (approved):

PR #1020, opened 2023-11-11
Fix stylesheets for Chapters 8–

pavpanchekha:
When we moved to a `DEFAULT_STYLE_SHEET` global we didn't update later chapters to import that constant (bad for outlines) or, more importantly, didn't update later chapters that modify the stylesheet to load the new one.


src/lab15.py
```diff
@@ -108,6 +108,8 @@ def request(self, top_level_url, payload=None):
         body = response.read()
         s.close()
         return response_headers, body
+      
+DEFAULT_STYLE_SHEET = CSSParser(open("browser15.css").read()).parse()
```
chrishtr, line 112: Also delete line 802 of this file.

chrishtr (approved):

PR #1022, opened 2023-11-11
Add `AlphaAndBlend` operation to all chapters

pavpanchekha:
In Chapter 11, we first add `Alpha` and `Blend` separately, and then merge them later to create `AlphaAndBlend`. The `AlphaAndBlend` operation gains an extra `isolate` parameter so that a layer is created when clipping with destination-in. Other chapters are fixed up to use `AlphaAndBlend` instead of `SaveLayer`.


book/visual-effects.md
```diff
@@ -1079,31 +1079,32 @@ which describes the opacity as a floating-point number from 0 to 1.
...
+in our display list, let's handle `opacity` with an `Alpha`
 command that takes a sequence of other drawing commands as an
 argument:
 
-``` {.python expected=False}
-class SaveLayer:
-    def __init__(self, sk_paint, children):
-        self.sk_paint = sk_paint
+``` {.python}
+class Alpha:
```
chrishtr, line 1087: Sorry for the late idea but should we call it Opacity instead of Alpha? Alpha is not the same as opacity..

book/visual-effects.md
```diff
@@ -1202,8 +1203,8 @@ class Pixel:
 I want to emphasize that this code is not a part of our browser---I'm
 simply using Python code to illustrate what Skia is doing internally.
 
-That `Alphaf` operation applies to pixels in one surface. But with
-`SaveLayer` we will end up with two surfaces, with all of their pixels
+That `alphaf` operation applies to pixels in one surface. But with
```
chrishtr, line 1206:
```suggestion
That `Alphaf` operation applies to pixels in one surface. But with
```

book/visual-effects.md
```diff
@@ -1558,56 +1591,83 @@ the page, so that clipping only applies to that element.
...
-    needs_blend_isolation = blend_mode != skia.BlendMode.kSrcOver or \
-        needs_clip
-    needs_opacity = opacity != 1.0
-
    return [
-        SaveLayer(skia.Paint(BlendMode=blend_mode), [
-            SaveLayer(skia.Paint(Alphaf=opacity), cmds,
-                should_save=needs_opacity),
-            SaveLayer(skia.Paint(BlendMode=skia.kDstIn), [
+        Blend(blend_mode, clip_radius > 0, [
```
chrishtr, line 1658: We could avoid the isolation parameter by this; WDYT?

if not blend_mode and clip_radius > 0:
  blend_mode = "source-over"
Blend(blend-mode, ...)

src/lab11.py
```diff
@@ -70,30 +70,73 @@ def parse_blend_mode(blend_mode_str):
...
     metrics = font.getMetrics()
     return metrics.fDescent - metrics.fAscent
 
-class SaveLayer:
-    def __init__(self, sk_paint, children,
-            should_save=True, should_paint_cmds=True):
-        self.should_save = should_save
-        self.should_paint_cmds = should_paint_cmds
-        self.sk_paint = sk_paint
+class Alpha:
```
chrishtr, line 82: I guess these classes are only there to make it easier to refer to them in the chapter. But it's dead code, which is something we don't have anywhere else...

src/lab13.py
```diff
@@ -760,15 +777,13 @@ def paint_visual_effects(node, cmds, rect):
...
-        SaveLayer(skia.Paint(BlendMode=blend_mode, Alphaf=opacity), node, [
-            ClipRRect(rect, clip_radius, cmds,
-                should_clip=needs_clip),
-        ], should_save=needs_blend_isolation)
-
-    transform = Transform(translation, rect, node, [save_layer])
-
-    node.save_layer = save_layer
+    blend_op = AlphaAndBlend(opacity, blend_mode, False, node, [
+        ClipRRect(rect, clip_radius,
```
chrishtr, line 781: This needs to be AlphaOrBlend("destination-in", ...

chrishtr: Can we delete ClipRRect also?

book/visual-effects.md
pavpanchekha: Sure, but then OpacityAndBlend sounds bad, can I just call the final thing Blend?

book/visual-effects.md
pavpanchekha: Ah, I see, you meant the Skia parameter name. Got it.

book/visual-effects.md
pavpanchekha: Oh yeah, we can! I initially had the parse elsewhere and changed it for unrelated reasons but now this works great!

src/lab11.py
pavpanchekha: Yeah, I can put them in the example file for compare to do it’s thing.

src/lab13.py
pavpanchekha: Will do but I think separate PR, it is a little complicated due to the local absolute thing in a later chapter. (I think we want a ClipRRect operation even if it works by destination in blending.

chrishtr: sg

book/visual-effects.md
chrishtr: Sounds good!

pavpanchekha:
This blocks a PR about removing ClipRRect, which is going to be moderately nasty, so I'd appreciate quick review. I think I've addressed all of the issues noted previously.

chrishtr (approved):

PR #1031, opened 2023-11-14
Get rid of the font method from chapter 6

chrishtr:
Inline the code instead. 

The global font method introduced in chapter 15 will stay, because it needs to be used by 3 different classes that don't
inherit from each other.


pavpanchekha (approved):
I'm surprised there are no changes to later chapters? I think in Chapter 16 there is some similar method that was a big pain...

chrishtr:
> I'm surprised there are no changes to later chapters? I think in Chapter 16 there is some similar method that was a big pain...

lab16.py patches the lab15 font method, which is not deleted.

PR #1032, opened 2023-11-14
Fix incorrect layout mode for images and iframes

chrishtr:


book/embeds.md
```diff
@@ -902,6 +917,12 @@ a "seamless" iframe whose layout coordinates with its parent frame.
 ``` {.python}
 class BlockLayout:
     # ...
+
+    def layout_mode(self):
+        # ...
+        elif self.node.tag in {"input", "img", "iframe"}:
```
chrishtr, line 923: Should this include "button"?

pavpanchekha: Uhh, I think so, though it should only matter in pretty weird cases since `button` elements typically have children.

book/embeds.md
```diff
@@ -484,6 +484,21 @@ class BlockLayout:
...
+``` {.python replace=%22img%22/%22img%22%2c%20%22iframe%22}
+class BlockLayout:
+    def layout_mode(self):
+        # ...
+        elif self.node.tag in {"input", "img"}:
+            return "inline"
+
+    def should_paint(self):
+        return isinstance(self.node, Text) or \
+            (self.node.tag not in {"input", "button", "img"})
```
pavpanchekha, line 498: Can we use a list `["input", "button", "img"]` instead of a set? We don't use sets anywhere else and this way we avoid forcing a tiny bit of Python syntax on the reader. With 3 elements it's plausible that lists are also faster (linear scan on 3 elements vs hash lookup) though in Python everything has such enormous overheads that that kind of logic is never decisive.

pavpanchekha (approved):
I am OK to merge this, and some kind of change like this is necessary, though I continue to think that `should_paint` is pretty ugly and we should find another way. Not urgent though!

PR #1034, opened 2023-11-15
Add RGBA color support

pavpanchekha:
This PR only adds RGBA _support_, and doesn't yet use this support to introduce opacity & isolation. That's for a later PR


src/lab11.py
```diff
@@ -65,6 +65,12 @@ def parse_color(color):
         g = int(color[3:5], 16)
         b = int(color[5:7], 16)
         return skia.Color(r, g, b)
+    elif color.startswith("#") and len(color) == 9:
```
chrishtr, line 68: Add a unittest for this

PR #1039, opened 2023-11-16
Only add `ClipRRect` if necessary, not always

pavpanchekha:
This PR changes the clip handling code in `paint_visual_effects` to only create a `ClipRRect` command if necessary. Previously we would always create the command, and pass it a `should_clip` parameter to turn it into a no-op. That approach also works (it's maybe a bit more verbose) but more importantly in Chapter 11 that `should_clip` flag is only added to `ClipRRect`, not to the earlier (and eventually only) destination-in-blend clip method. Which means that earlier on, when we're doing destination-in-blending, we're clipping off very long words even when overflow isn't set.

This PR also aligns the various cross-chapter copies of `paint_visual_effects`, which had drifted a bit and had dead code.


pavpanchekha:
Chris, I know you really want to just delete `ClipRRect` entirely, but there are some challenges there related to the absolute-to-local optimization in Chapter 13. Hence I'm trying to split off independent changes as micro-PRs.

chrishtr (approved):

PR #1043, opened 2023-11-18
Fix bugs breaking wikipedia cats in chapter 15

chrishtr:
1. Protocol-relative URLs were not supported, and that's most of the images on this page.
2. The `AttributeParser` got confused in some situations where a style sheet ended up being parsed as an attribute. I fixed it by eating the rest of the string if `word` fails to advance.
3. A typo in the exception handler for failed image loads.

Now this URL can be loaded, and the images show: https://en.wikipedia.org/wiki/Cat

### DIFF
--- a/book/styles.md
+++ b/book/styles.md
@@ -599,8 +599,7 @@ including query-relative and scheme-relative URLs, that I'm skipping.]
 -   A path-relative URL, which doesn't start with a slash and is
     resolved like a file name would be.
 -   A protocol-relative URL that starts with "//" followed by a full URL,
-    which should be completed to a full url using the same protocol
-    (`http` or` https`) as the current URL.
+    which should use the existing scheme.
 
 To download the style sheets, we'll need to convert each relative URL
 into a full URL:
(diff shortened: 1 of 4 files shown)


book/styles.md
```diff
@@ -598,6 +598,9 @@ including query-relative and scheme-relative URLs, that I'm skipping.]
     existing scheme and host; or
 -   A path-relative URL, which doesn't start with a slash and is
     resolved like a file name would be.
+-   A protocol-relative URL that starts with "//" followed by a full URL,
+    which should be completed to a full url using the same protocol
+    (`http` or` https`) as the current URL.
```
pavpanchekha, line 603:
```suggestion
    which should use the existing scheme.
```

pavpanchekha: We call it a "scheme" throughout.

pavpanchekha (approved):
Great! I'm personally still a bit torn on `AttributeParser`. To my knowledge it does not appear in the book, so I think we should drop it entirely from the code. But either way it should work!

book/styles.md
```diff
@@ -598,6 +598,8 @@ including query-relative and scheme-relative URLs, that I'm skipping.]
     existing scheme and host; or
 -   A path-relative URL, which doesn't start with a slash and is
     resolved like a file name would be.
+-   A protocol-relative URL that starts with "//" followed by a full URL,
```
chrishtr, line 601:
```suggestion
-   A scheme-relative URL that starts with "//" followed by a full URL,
```

PR #1044, opened 2023-11-18
Pass over Chapter 16

pavpanchekha:
The main thing this pass had to do is align with the new `ascent` and `descent` fields on `EmbedLayout`.

Note: I switched from `super().layout()` to `EmbedLayout.layout(self)`. This has something complicated to do with how `super()` works, which I don't understand, and our patch system. But it should be invisible to the reader and won't be a big problem to maintain.


src/lab16.py
```diff
@@ -776,24 +776,20 @@ def layout_before(self):
         else:
             self.x.copy(self.parent.x)
 
-    def layout_after(self):
-        height = self.height.read(notify=self.ascent)
-        self.ascent.set(-height)
-
-        self.descent.set(0)
-
-        self.has_dirty_descendants = False
```
chrishtr, line 785: Did you forget to add this line in the refactored code?

PR #1045, opened 2023-11-19
Fix three startup rendering issues

chrishtr:
* The screen shows black on my computer. "Fixed" by calling `draw` twice, which somehow causes more state flushing. (I looked though the SDL documentation and couldn't find any simple fix.)
* Fixed lab13 mainloop, which needs to call `composite_raster_and_draw`.
* Initialized `root_frame_focused` in chapter 15


pavpanchekha (approved):
The lab15 stuff is for root frame focused?

chrishtr:
> The lab15 stuff is for root frame focused?

Yes. Forgot to mention that.

PR #1047, opened 2023-11-19
Remove `ClipRRect`

pavpanchekha:
This PR removes `ClipRRect` entirely. In its place:

- All clipping now uses `destination-in` blending instead
- `Blend` has an unusual `map` operation that looks for clipping among its children


book/visual-effects.md
```diff
@@ -1603,7 +1603,16 @@ of two sheets of paper and destination-in compositing playing the role
 of the scissors. This implementation technique for clipping is called
 *masking*, and it is very general---you can use it with arbitrarily
 complex mask shapes, like text, bitmap images, or anything else you
-can imagine.
+can imagine.[^cliprrect]
+
+[^cliprrect]: If all our browser wanted to clip were rounded
```
chrishtr, line 1608: Can you check that this PR does not slow down rendering due to lack of the skia-internal
clipping optimization? If it is lost I'd expect raster to get slower.

pavpanchekha: In chapter 11 & 12, we’re not doing a lot of clipping anyway. By 15 where we add iframes, it’s not a big deal because we draw the iframe contents to a small surface, which is implicitly clipped.

chrishtr (approved):

PR #1048, opened 2023-11-19
Fix slow compositing in chapter 16

chrishtr:
The root cause was that we were calling `parse_blend_mode` in chapter 16 inside of `paint_visual_effects`, which was:
* A mismatch from what the `Blend` class needed (it expected to parse it itself)
* Causing the `blend_mode` variable to be not `None` (because it was `kSrcOver` due to the `normal` value for `mix-blend-mode`, which in turn made each such `Blend`'s `should_save` true, which in turn triggers compositing.

The fix is:
* Always expected `skia.BlendMode` parameter to `Blend`
* Define `should_save` to look for non-srcover blending

### DIFF
--- a/src/lab11.py
+++ b/src/lab11.py
@@ -84,7 +84,7 @@ def parse_blend_mode(blend_mode_str):
         return skia.BlendMode.kDifference
     elif blend_mode_str == "destination-in":
         return skia.BlendMode.kDstIn
-    elif blend_mode_str == "source-over":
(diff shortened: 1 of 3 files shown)


pavpanchekha:
Ok, I got the bug but I’m not a huge fan of the solution. Can we not just special-case “normal”? And in prior chapters default to “normal”? I think the code would be nicer.

PR #1055, opened 2023-11-21
Add screenshots and better tracing for chapter 12 concepts

chrishtr:
* Screenshots added for single-thread, multi-thread and scrolling
* Links to sample traces that are easy to load
* tracing for JS tasks
* Commentary
* Fixed up stale references to 16ms


book/scheduling.md
```diff
@@ -557,7 +557,7 @@ tasks at a fixed cadence. But besides JavaScript the browser also has
 to render the page, and as you may recall from [Chapter
 2](graphics.md#framebudget), we'd like the browser to render the page
 exactly as fast as the display hardware can refresh. On most
-computers, this is 60 times per second, or 16ms per frame. However, even
+computers, this is 60 times per second, or 33ms per frame. However, even
```
pavpanchekha, line 560:
```suggestion
computers, this is 30 times per second, or 33ms per frame. However, even
```

src/lab12.py
```diff
@@ -108,14 +108,18 @@ def __init__(self, tab):
...
+        self.tab.browser.measure.time('script-runtime')
         self.interp.evaljs(RUNTIME_JS)
+        self.tab.browser.measure.stop('script-runtime')
 
         self.node_to_handle = {}
         self.handle_to_node = {}
 
     def run(self, script, code):
         try:
+            self.tab.browser.measure.time('script-load')
```
pavpanchekha, line 120: I say we don't do this, but in principle we could extend `time` to also take an argument, and then these could be `script` blocks with different arguments. I think the chrome://tracing tools then let you aggregate all `script`s and look at the various arguments. Anyway, I don't think we should actually do it.

pavpanchekha (approved):

PR #1060, opened 2023-11-22
Fix composited clipping

chrishtr:
The current solution doesn't work, because the blend visual effect for a destination-in clip needs to be in the same composited layer as the visual and paint effects it's clipping. Therefore we need to make sure that such clipping is applied atomically along with the containing isolation blend.

Solution: put the clipping mask directly on the isolation `Blend` object as a new parameter. This actually simplifies some of the code (e.g. `map`), and centralizes bespoke code to apply the `destination-in` blend mode.


pavpanchekha:
Ok, so I understand the bug and also agree that this would fix it, but the way I see it, the issue isn't about what goes in which composited layer but rather the fact that we don't execute effects in the paint draw list atomically. That is, we incorrectly get this paint draw list:

    Blend(source-over)
        DrawCompositedLayer()
    Blend(source-over)
        Blend(destination-in)
            DrawCompositedLayer()

but if we did it right we'd have this:

    Blend(source-over)
        DrawCompositedLayer()
        Blend(destination-in)
            DrawCompositedLayer()

That would work, right?

chrishtr:
> That would work, right?

Yes, it would would. And yes you could describe the bug as you're phrasing it. Either way, I think the best fix is what I've done here. (incidentally it's also what Chrome does)

pavpanchekha:
Ok, I really think the isolated paint draw display list is an easier change. See #1079.

chrishtr:
> Ok, I really think the isolated paint draw display list is an easier change. See #1079.

Not totally sure why 1079 doesn't fix the issue PR 1060 solves. Going to keep thinking about it.

chrishtr:
> > Ok, I really think the isolated paint draw display list is an easier change. See #1079.
> 
> Not totally sure why 1079 doesn't fix the issue PR 1060 solves. Going to keep thinking about it.

Figured it out. Your PR had a bug in that lab15.py was not updated. With that file changed it works!

PR #1063, opened 2023-11-22
Convert all figures to latex-friendly format

chrishtr:
Note: the `<br>` elements are there as a hack to avoid showing the alt text for sighted users. Pandoc shows the alt text as a caption automatically unless it detects text in the same paragraph after the image, which it assumes is a caption. The book already has contextual text indicating what the image is, so I removed the caption via `<br>`.

Latex also doesn't understand GIF, so I converted one image from GIF to PNG.


chrishtr:
This is based on top of PR #1062.

pavpanchekha (approved):

PR #1065, opened 2023-11-22
Add support for example transclusion and use it in chapter 14

chrishtr:
Transclusion reads a file and inlines it into the chapter (as opposed to "cmd", which runs a command and inlines the result).

Use it to inline examples into the print edition for Chapter 14.

If this pattern lands, I plan to next go through all chapters and provide an inline version of all examples in the print edition.


chrishtr:
I guess an alternate solution would be :

```
``` {.cmd}
cat <filename>
``` <end>
```
Up to if you prefer that.

PR #1067, opened 2023-11-24
Remove dependency of children on anything

chrishtr:
Also fixes some horizontal scroll issues in chapter 16

### DIFF
--- a/src/lab16.py
+++ b/src/lab16.py
@@ -348,7 +348,7 @@ class BlockLayout:
         self.y = ProtectedField(self, "y", self.parent, y_dependencies)
 
         self.children = ProtectedField(self, "children", self.parent, None,
-            [self.height])
+            [])
 
         self.has_dirty_descendants = True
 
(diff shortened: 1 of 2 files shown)


src/lab16.py
```diff
@@ -345,10 +345,11 @@ def __init__(self, node, parent, previous, frame):
             y_dependencies = [self.previous.y, self.previous.height]
         else:
             y_dependencies = [self.parent.y]
-        self.y = ProtectedField(self, "y", self.parent, y_dependencies)
+        self.y = ProtectedField(
+            self, "y", self.parent, y_dependencies)
 
         self.children = ProtectedField(self, "children", self.parent, None,
-            [self.height])
+            [])
```
pavpanchekha, line 352: This is the interesting change, right? Come to think of it, I'm not sure why this was ever here, or even why we have a frozen_invalidations field.

pavpanchekha (approved):

src/lab16.py
chrishtr: Yes this is the interesting one.

PR #1068, opened 2023-11-25
Add example interaction with the server to chapter 10

chrishtr:
This could be in lieu of a figure with the same information.


pavpanchekha:
Hmm, I'm worried this will be hard to read and take up a lot of space. I think I would rather a figure.

PR #1070, opened 2023-11-26
Fix print outlines

chrishtr:
Also omit the main function from the outlines - why include it?


pavpanchekha (approved):
Looks good.
- I assume we'll eventually want a custom LaTeX version of the outline but it doesn't seem that important right now.
- Omitting `IfMain` seems right to me.

PR #1075, opened 2023-11-26
At least do something with scroll offset when zooming

pavpanchekha:
This is not really the "right" fix (see added footnote) but I think doing nothing is worse because it suggests that you don't need to worry about scroll offset when zooming. (If there's anything written somewhere about what Chrome does, we can add that to the footnote if you'd like.)

### DIFF
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -233,10 +233,13 @@ class Tab:
         else:
             self.zoom *= 1/1.1
             self.scroll *= 1/1.1
+        self.scroll_changed_in_tab = True
         self.set_needs_render()
 
     def reset_zoom(self):
+        self.scroll /= self.zoom
         self.zoom = 1
+        self.scroll_changed_in_tab = True
         self.set_needs_render()
 ```
 
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -229,8 +229,10 @@ class Tab:
     def zoom_by(self, increment):
         if increment:
             self.zoom *= 1.1
+            self.scroll *= 1.1
         else:
             self.zoom *= 1/1.1
+            self.scroll *= 1/1.1
         self.set_needs_render()
 
     def reset_zoom(self):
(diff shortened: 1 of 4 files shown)


book/accessibility.md
```diff
@@ -239,8 +241,17 @@ class Tab:
...
+zoom the page we also need to adjust the scroll
+position.[^zoom-scroll] We also need to reset the zoom level when we
+navigate to a new page:
+
+[^zoom-scroll]: In a real browser, adjusting the scroll position when
+    zooming is more complex than just multiplying. That's because zoom
+    not only changes the height of individual lines of text, but also
+    changes line breaking, meaning more or fewer lines of text. This
+    means there's no easy correspondence between old and new scroll
+    positions.
```
chrishtr, line 254:
```suggestion
    positions. Most real browsers implement a much more general algorithm called [scroll anchoring](https://drafts.csswg.org/css-scroll-anchoring-1/) that handles all kinds of changes beyond just zoom.
```

src/lab14.py
```diff
@@ -1089,9 +1089,10 @@ def advance_tab(self):
     def zoom_by(self, increment):
         if increment:
             self.zoom *= 1.1
+            self.scroll *= 1.1
         else:
             self.zoom *= 1/1.1
-        print(self.zoom)
+            self.scroll *= 1/1.1
```
chrishtr, line 1095: Also fix lab15 and lab16

chrishtr:
I think you also need to adjust active_tab_scroll in Browser.increment_zoom.

pavpanchekha:
Ok, I thought about this, and I think there's no need to change `active_tab_scroll` in `increment_zoom`, because we only want the new scroll position _after_ the tab re-renders at the larger size. So instead what we need to do is set `scroll_changed_in_tab` when incrementing zoom so that the `Tab` can overwrite the `active_tab_scroll` when it commits the zoomed layout.

book/accessibility.md
```diff
@@ -229,18 +229,32 @@ class Tab:
...
+        self.scroll_changed_in_tab = True
         self.set_needs_render()
 ```
 
 Note that we need to set the `needs_render` flag when we zoom to
-redraw the screen after zooming is complete. We also need to reset the
-zoom level when we navigate to a new page:
+redraw the screen after zooming is complete. Also note that when we
+zoom the page we also need to adjust the scroll
+position.[^zoom-scroll] We also need to reset the zoom level when we
```
chrishtr, line 249:
```suggestion
position,[^zoom-scroll] and reset the zoom level when we
```

chrishtr (approved):

PR #1079, opened 2023-11-27
Isolate effects in draw display list

pavpanchekha:
Hi @chrishtr, I'm really sorry, I know you said *not* to do this, but it really does seem like a simpler change than #1060, which felt to me a lot like working around a bug. I've tested on a bunch of pages and it seems to work without any visual residue or whatever.


pavpanchekha:
Just as an example that this does work, here's a screenshot:

image

Here the blue and green overlap. Green clips. Blue is partially transparent (it pulses in and out, and compositing works). And then the whole thing is wrapped into an opacity, which applies atomically. Here's the `fix-animation-bugs` branch (which is necessary for things to run):

image

Note that the two objects bleed through because they're in different composited layers.

chrishtr:
So I think we should land this PR (once you confirm it's correct) but we still need 1060.

chrishtr (changes requested):
See comment for requested changes.

chrishtr (approved):
LGTM

PR #1082, opened 2023-11-29
Add screenshots demonstrating the performance of GPU raster and compositing

chrishtr:
Also, fixed a bug in `composited_raster_and_draw` where we shouldn't do anything (including tracing) if none of the three dirty bits are set.


book/animations.md
```diff
@@ -1374,8 +1418,8 @@ class Browser:
 
     def composite_raster_and_draw(self):
         if not self.needs_composite and \
-            len(self.composited_updates) == 0 \
-            and not self.needs_raster and not self.needs_draw:
+            not self.needs_raster and \
+            not self.needs_draw:
```
pavpanchekha, line 1422: Why is this right / why does this matter? In what situation would we have an empty `composited_updates` (meaning, can't composite) but not need to composite? I suppose if we composited on the last frame? Is this just about speeding up idle frames? I'm confused.

pavpanchekha:
I'm definitely a fan of speeding things up a lot but I don't get this change.

book/animations.md
pavpanchekha: Oh, I overinterpreted. Easy PR.

PR #1085, opened 2023-11-29
LaTeX blocks

pavpanchekha:
This adds support for blocks in the LaTeX output.

The main latex code is in `latex/macros.tex`. Like all latex code it is just the worst, most horrible code in the world doing the most horrible things. (1 point for `\csname`, 1 point for `\expandafter`.) I assure you I refactored it before PRing.

The main Lua code is in `Div` at the end, but many many other random Lua changes were necessary because we now _don't_ want any blocks in the LaTeX code except the blocks we care about. (If this becomes burdensome we could do an allowlist of blocks instead of a denylist.)

Also added support for hyperlinks (easy).

Here's what it looks like:

image

Colors are configurable but I sorta tried to match the website.


pavpanchekha:
I'll look into turning hyperlinks into endnotes later.

pavpanchekha:
Or perhaps the publisher can deal with endnotes.

infra/filter.lua
```diff
@@ -68,13 +71,15 @@ end
...
   if not config.show_todos and el.classes[1] == "todo" then
     return pandoc.RawBlock("html", "")
-  elseif not config.print and el.classes[1] == "print-only" then
-    return pandoc.RawBlock("html", "")
-  elseif config.print and el.classes[1] == "web-only" then
-    return pandoc.RawBlock("html", "")
-  elseif config.show_signup and el.classes[1] == "signup" then
-    local signup = assert(io.open("infra/signup.html")):read("*all")
-    return pandoc.RawBlock("html", signup)
+  elseif el.classes[1] == "print-only" then
```
chrishtr, line 74: is there an easy Lua syntax for

"print-only" in el.classes

to be robust to order of classes?

infra/filter.lua
```diff
@@ -114,6 +119,13 @@ function Div(el)
     local div = pandoc.CodeBlock( io.read('a'))
     div.classes = el.classes
     return div
+  elseif config.print and #el.classes > 0 and el.classes[1] ~= "center" then
```
chrishtr, line 122: Is this "if print and not centered, then use bookblock to do left-aligned layout"?

latex/macros.tex
```diff
@@ -0,0 +1,43 @@
...
+  \expandafter\def\csname @val@#1@#2\endcsname{#3}
+}
+
+\newcommand{\getprop}[2]{%
+  \ifcsname @val@#1@#2\endcsname%
+  \csname @val@#1@#2\endcsname%
+  \fi
+}
+
+% Next, here, we define  the different types of boxes
```
chrishtr, line 17:
```suggestion
% Next, here, we define the different types of boxes
```

chrishtr:
Looks like this broke outlines in the web version.

infra/filter.lua
pavpanchekha: I rewrote this code, it was overly obscure. But this is the code that says "for all other blocks, in print mode, use bookblock to make them show up like a colored box.

pavpanchekha:
That was a challenging bug to fix (I had to learn a lot of Lua and a lot more about Pandoc) but I think it is fixed now.

pavpanchekha:
I do love that "make colored boxes in a PDF file" ends up being harder than, like, "add clip compositing" or whatever to the browser :)

infra/filter.lua
chrishtr: Ok, what's up with the classes != center part then? Why is that needed?

pavpanchekha: Eliminated now

infra/filter.lua
```diff
@@ -16,6 +16,7 @@ function LoadMeta(meta)
   meta.rel = config.rel
   meta.base = config.base
   meta.draft = config.draft
+  meta.colorlinks = true
```
chrishtr, line 19: What does this line do?

chrishtr (approved):

infra/filter.lua
pavpanchekha: Make links blue in the PDF. Weirdly, this is setting a pandoc setting, which then causes our template to generate different LaTeX with different settings. It would be cleaner to make this change entirely in LaTeX but right now the template is a bit of a mess and I didn't want to fight it too much.

PR #1090, opened 2023-12-01
Add non-hand-drawn figures to Chapters 1-10

pavpanchekha:
These are all the non-hand-drawn figures for Chapters 1-10. The hand-drawn ones, annoyingly, require me to upgrade to macOS 14 (what?!) but they are all drawn.

Also deleted one obsolete figure that we no longer use.


pavpanchekha:
Let me also note that the way images look on the web version is currently not 100% great but I think that can be fixed in, like, a week. The images also don't look amazing in the LaTeX version but a lot of that will be up to the typesetter anyway.

book/html.md
```diff
@@ -16,7 +16,18 @@ A tree of nodes
...
-and for each span of text.[^1] So for our browser to use a tree, tokens need to
+and for each span of text.[^1]
+
+[^1]: In reality there are other types of nodes too, like comments,
+    doctypes, and `CDATA` sections, and processing instructions. There
+    are even some deprecated types!
+
+::: {.print-only .center}
+![An annotated HTML document](im/html-syntax.png)
+^[An HTML document, showing tags, text, and the nesting structure.]
```
chrishtr, line 27: Why the footnote? Why not just rely on the alt text?

book/text.md
```diff
@@ -39,6 +39,23 @@ typing. Variations—like bold or italic letters—were called that type's
 
 [^2]: The word is related to *foundry*, which would create the little
     metal shapes.
+    
+::: center
```
chrishtr, line 43:
```suggestion
::: {.center}
```

chrishtr (approved):

book/text.md
pavpanchekha: Actually this works but you're right, we typically do this this way.

book/html.md
pavpanchekha: Yeah, uhh, I don't know. This is how we do it in `history.md` and it looks terrible. I figured now wasn't the time to figure it out :(

PR #1097, opened 2023-12-03
Fix figures in all chapters to use a consistent approach

chrishtr:
Figures:

* Centered
* Has a caption visible
* Copyright in a footnote so it doesn't distract the reader

Footnotes:
* Named, not numbered

Examples:

* Have a gray border

Browser widgets:

* Referenced in the Summary section of every chapter
* Fixed bugs in the chapter 12 and 13 browser widgets (double-import of some constants; `skia.Color` constructor; accidental double-declaration of the `Chrome` class.


chrishtr:
Examples still don't render in the printed edition with a border because of something to do with how they use "code" styling.

book/animations.md
```diff
@@ -218,10 +218,15 @@ and draw a given page can be quite complex.
...
 [gpu]: https://en.wikipedia.org/wiki/Graphics_processing_unit
 
+::: {.installation}
+
 First, we'll need to install the OpenGL library:
 
     pip3 install PyOpenGL
 
+
+:::
```
pavpanchekha, line 228: Not sure I agree with this one; we usually use an `installation` box for optional steps that one some users will need.

book/visual-effects.md
```diff
@@ -1181,7 +1181,7 @@ conceptual `Pixel` class I introducd earlier.
 When we apply a `Paint` with an `Alphaf` parameter, the first thing
 Skia does is add the requested opacity to each pixel:
 
-``` {.python file=examples}
+``` {.python file=examples .example}
 class Pixel:
```
pavpanchekha, line 1185: This might be turning off the auto-comparison for these snippets (they're normally compared to, I think, `example11.py`. I think auto-compare skips things labeled `example`. I'm not against the change though. Maybe we need to tweak something in `infra/compare.py`.

src/lab12.py
```diff
@@ -732,92 +734,9 @@ def handle_quit(self):
             tab.task_runner.set_needs_quit()
         sdl2.SDL_DestroyWindow(self.sdl_window)
 
-@wbetools.patch(Chrome)
```
pavpanchekha, line 735: Why delete this? I thought the change was using `active_tab_url` instead of just `tab.url`, which is a potential race?

pavpanchekha: Oh, weird, we had two. Huh. Ok, got it.

pavpanchekha (approved):

book/animations.md
chrishtr: Shouldn't "installation" be for all software installation instructions?

pavpanchekha: Well, we've never really defined these boxes well, but I think installation instructions are normally part of the main part of the text, with quirks described elsewhere. Rationalizing the boxes would be good, but maybe that's a separate PR—we'd want to look over how installation instructions are treated in other chapters. (Also maybe `quirk` and `installation` boxes are in fact the same thing, if they both describe installation quirks?)

pavpanchekha:
Anyway, once you merge I'll deploy it.

PR #1099, opened 2023-12-29
Update renderingng link in scheduling.md

xiaochen-z:
The link [https://developer.chrome.com/blog/renderingng/](https://developer.chrome.com/blog/renderingng/) referenced by text "only recently" is outdated.

I guess the link you want to link is [https://developer.chrome.com/docs/chromium/renderingng](https://developer.chrome.com/docs/chromium/renderingng). Please feel free to edit to the desired link.


chrishtr:
Thanks for the fix!

PR #1106, opened 2023-12-31
Don't crash on malformed URLs

chrishtr:
Instead, fall back to the WBE homepage.

Even better would be to properly show nothing or an error message, but this is very simple and works.

I omitted this detail from the book because it doesn't seem relevant enough to justify inclusion in the text. However,
it constantly happens when trying to type URLs and making mistakes. Therefore the code we provide should avoid
crashing in this situation.

### DIFF
--- a/book/http.md
+++ b/book/http.md
@@ -790,7 +790,7 @@ itself what you need to replace.)
 
 Encrypted HTTP connections usually use port 443 instead of port 80:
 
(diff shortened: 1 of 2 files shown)


pavpanchekha (approved):
I’m wary of making the book code different from the Python code but this does look safe.

PR #1108, opened 2023-12-31
Add chapter 14 browser

chrishtr:
(does not yet add new key bindings for accessibility)


book/accessibility.md
```diff
@@ -873,7 +873,7 @@ class Browser:
 
     def cycle_tabs(self):
         self.lock.acquire(blocking=True)
-        active_ids = self.tabs.index(self.active_tab)
+        active_idx = self.tabs.index(self.active_tab)
```
pavpanchekha, line 876: Since it looks like we use `.find` and `.index`, which I think differ only in whether they return `None` or throw when missing, should we change things to only ever use one? Minor, I know, but it'd be one less Python thing to learn for readers.

pavpanchekha (approved):

book/accessibility.md
chrishtr: `find` is for strings, but `index` works for arrays...

PR #1113, opened 2024-01-09
Add lab16 browser

chrishtr:
This required a number of changes:

* Adding support for named arguments, via custom destructuring code
* Adding support for `dict`
* Adding support for `set`s
* Fixing patching of functions to actually work (was broken in lab15, I just didn't notice it). The problem is that global variables in a JS module are const from an external perspective. The only way to change them is via cod within that same module. So I added `patchable` notation to trigger generation of special functions that patch global variables.
* Making read not a built-in function, because `ProtectedValue` uses it.
* Adding a new hint mechanism for more cases: where `get` is called on a non-`ProtectedValue` (seemed easier than adding hints for all cases), and in list comprehensions types
* Various fixes I also needed for chapter 15's browser


infra/asttools.py
```diff
@@ -113,9 +120,17 @@ def visit_ImportFrom(self, cmd):
...
+        self.patchables = {}
 
     def visit_FunctionDef(self, cmd):
-        if not cmd.decorator_list:
+        if not cmd.decorator_list or not is_patch_decorator(cmd.decorator_list[0]):
+            if cmd.decorator_list:
+                if not is_patchable_decorator(cmd.decorator_list[0]):
+                    print(cmd.decorator_list)
+                    print(isinstance(cmd, ast.Call))
+                    raise Exception()
```
pavpanchekha, line 131: I assume this is debugging code. IIRC there's some mechanism in the compiler for nice error messages, don't remember for sure though.

infra/compile.py
```diff
@@ -106,14 +106,14 @@ def check_args(args, ctx):
...
     "index": "indexOf",
-    "copy": "slice",
 }
 
 RENAME_FNS = {
     "int": "parseInt",
     "float": "parseFloat",
     "print": "console.log",
-    "AnimationClass": "AnimationClass"
+    "AnimationClass": "AnimationClass",
```
pavpanchekha, line 115: IIRC this is no longer used, since we only have one kind of Animation.

infra/compile.py
```diff
@@ -300,6 +299,10 @@ def load_outline(module):
 def compile_method(base, name, args, ctx):
     base_js = compile_expr(base, ctx)
     args_js = [compile_expr(arg, ctx) for arg in args]
+    needs_call = False
+    if base_js[0].isupper():
+        base_js = base_js + ".prototype"
+        needs_call = True
```
pavpanchekha, line 305: What's the Python code that needs this? It seems reasonable but I'm not sure what it's for.

infra/compile.py
```diff
@@ -640,7 +663,13 @@ def compile_expr(tree, ctx):
         e = compile_expr(tree.elt, ctx2)
         if e != arg:
             if "await" in e:
-                out = "(await Promise.all(" + out + ".map(async (" + arg + ") => " + e + ")))"
+                try:
+                    t = find_hint(gen.iter, "type")
+                    out = 'Object.keys(' + out + ')'
+                    assert t == "map"
+                except:
```
pavpanchekha, line 670: Kind ugly to do a general-purpose catch here, but OK

pavpanchekha: Probably we just need a default parameter to `find_hint`?

pavpanchekha (approved):
Looks good and the handling of patchable functions is great.

infra/compile.py
chrishtr: It's for chapter 16, which instead of calling `super().layout()` for `EmbedLayout` subclasses, calls `EmbedLayout.layout`. This code detects that pattern and uses the `call` syntax which is equivalent to it, and allows passing `this`.

The reason lab16 uses that pattern in Python is because it's using the patching mechanism, which messes up inheritance, and using `super()` therefore throws an exception saying you can't do that if one class is not inheriting from another.

infra/compile.py
chrishtr: Great idea - done.

PR #1123, opened 2024-02-13
Fix style.md tests and the test runner

chrishtr:
The test runner had a bug where it wouldn't add up failures across multiple chapters. That's why the bug fixed
here crept in.

### DIFF
--- a/book/styles.md
+++ b/book/styles.md
@@ -521,8 +521,9 @@ pre { background-color: gray; }
 Let's store that in a new file, `browser.css`, and have our browser
 read it when it starts:
 
-``` {.python replace=browser.css/browser6.css}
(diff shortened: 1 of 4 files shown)


pavpanchekha (commented):
Wait, I'm confused about the actual bug: the fact that only book code is changed, not the source code, and also the fact that it seems like the two code blocks are equivalent?

chrishtr:
The source code didn't match the book. It failed the test locally, but that failure was masked by the other thing I had to fix.

chrishtr:
Ok you're right, I forgot about the _style suffix. Fixed.

pavpanchekha (approved):

PR #1129, opened 2024-03-08
Fix deprecation warnings on AST module

pavpanchekha:
Python 3.12 adds deprecation warnings to the `ast` module for use of the Python 3.9-style AST node `Str`. This PR adds two new functions, `is_str_ast` and `str_value`, that use some type interrogation to not use those AST nodes and thus avoid the deprecation warning.

### DIFF
--- a/infra/compile.py
+++ b/infra/compile.py
@@ -4,6 +4,7 @@ import ast, asttools
 import json
 import warnings
 import outlines
+import platform
 
 INDENT = 2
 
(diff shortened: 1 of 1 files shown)


chrishtr (approved):
Please make sure all the compiled code and demo browsers for each chapter still work.

pavpanchekha:
Confirmed, no changes to compiled code at all.

PR #1130, opened 2024-03-08
Fix default style sheet location

pavpanchekha:
In earlier versions of the book, the default style sheet was stored in a field on `Browser` but it's now a global constant and not all the code got updated.

### DIFF
--- a/src/lab10.py
+++ b/src/lab10.py
@@ -160,7 +160,7 @@ class Tab:
             except dukpy.JSRuntimeError as e:
                 print("Script", script, "crashed", e)
 
-        self.rules = self.default_style_sheet.copy()
+        self.rules = DEFAULT_STYLE_SHEET.copy()
         links = [node.attributes["href"]
                  for node in tree_to_list(self.nodes, [])
                  if isinstance(node, Element)
--- a/src/lab8.py
+++ b/src/lab8.py
@@ -243,6 +243,12 @@ def paint_tree(layout_object, display_list):
 
 @wbetools.patch(Tab)
 class Tab:
+    def __init__(self, tab_height):
+        self.url = None
+        self.history = []
+        self.tab_height = tab_height
+        self.focus = None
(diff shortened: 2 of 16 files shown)


src/lab10.py
```diff
@@ -160,7 +160,7 @@ def load(self, url, payload=None):
             except dukpy.JSRuntimeError as e:
                 print("Script", script, "crashed", e)
 
-        self.rules = self.default_style_sheet.copy()
+        self.rules = DEFAULT_STYLE_SHEET.copy()
```
chrishtr, line 163: Did lab10 just not work previously?

pavpanchekha: I think `lab10` was using the `default_style_sheet` field being created in `lab8`. So yes `lab8` was redundantly storing the default style sheet in _both_ places, and some labs were using one place while some were using the other.

src/lab8.py
```diff
@@ -243,23 +243,14 @@ def paint_tree(layout_object, display_list):
 
 @wbetools.patch(Tab)
 class Tab:
-    def __init__(self, tab_height):
```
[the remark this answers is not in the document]
pavpanchekha: You're right.

src/lab8.py
```diff
@@ -369,7 +360,6 @@ def __init__(self):
         self.tabs = []
         self.active_tab = None
         self.focus = None
-        self.address_bar = ""
```
chrishtr, line 372: why remove this?

pavpanchekha: I think it now lives in the Chrome

chrishtr: I guess this was just a bug?

chrishtr (approved):

src/lab8.py
pavpanchekha: Sure… more like a harmless leftover from a refactor.

Issue #1126, opened 2024-02-17
Improve "Word By Word" section


pavpanchekha:
This is feedback from a student, and I think I agree. The ["Word By Word" section in Chapter 3](https://browser.engineering/text.html#word-by-word) is a short section on how to update the text drawing loop to handle multiple words, but it's a bit confusing:

1. It's not clear where it goes in the book. I think we can clarify that this is in the `layout` function.
2. It would probably be nicer if instead of one code block and then three paragraphs of commentary, maybe the code block can be split up

Filing this as an issue instead of as a PR because I'm not sure whether we should be making changes as large as #2 at this point. If not, we can just do #1 and add a `def layout` header.

PR #1131, opened 2024-03-08
Rewrite the "word by word" section

pavpanchekha:
Several students noted that this section was hard to follow, both because it was unclear where the code is supposed to go and also because there was a lot of text describing a lot of code and it was hard to match one with the other.

This PR breaks the section up into pieces, fixing #1126. The before/after is below: the section got longer, but I think that's fine because it was a short section to begin with:

Screenshot 2024-03-08 at 2 13 22 PM
Screenshot 2024-03-08 at 2 13 28 PM

### DIFF
--- a/book/text.md
+++ b/book/text.md
@@ -298,35 +298,47 @@ a time:[^whitespace]
     including for identifying word boundaries.
 
 ``` {.python expected=False}
-w = font.measure(word)
-if cursor_x + w > WIDTH - HSTEP:
-    cursor_y += font.metrics("linespace") * 1.25
-    cursor_x = HSTEP
-self.display_list.append((cursor_x, cursor_y, word))
-cursor_x += w + font.measure(" ")
+def layout(text):
+    # ...
+    for word in text.split():
+        # ...
+    return display_list
 ```
 
-There's a lot of moving parts to this code. First, we measure the
-width of the text, and store it in `w`. We'd normally draw the text at
-`cursor_x`, so its right end would be at `cursor_x + w`, so we check
-if that's past the edge of the page. If it is, we make space by
-wrapping to the next line. Now we have the location to *start* drawing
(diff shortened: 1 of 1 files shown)


book/text.md
```diff
@@ -298,35 +298,47 @@ a time:[^whitespace]
     including for identifying word boundaries.
 
 ``` {.python expected=False}
```
chrishtr, line 300: Why expected=false?

book/text.md
```diff
@@ -298,35 +298,47 @@ a time:[^whitespace]
...
+    # ...
+    for word in text.split():
+        # ...
+    return display_list
+```
+
+Unlike Chinese characters, words are different sizes, so we need to
+measure the width of each word:
+
+``` {.python expected=False }
```
chrishtr, line 311: Why expected=false?

book/text.md
```diff
@@ -298,35 +298,47 @@ a time:[^whitespace]
...
+        w = font.measure(word)
+    # ...
+```
+
+Here I've chosen to use Tk's default font. Now, if we draw the text at
+`cursor_x`, its right end would be at `cursor_x + w`. That might be
+past the right edge of the page, and in this case we need to make
+space by wrapping to the next line:
+
+``` {.python expected=False}
```
chrishtr, line 325: Why expected=false?

book/text.md
pavpanchekha: Same as above; the `for` loop moves elsewhere and the `font` line changes when we add styling so there's not much left to match.

book/text.md
pavpanchekha: Similar to above, though at least in principle we could make this match _something_ by dropping the cursor_x/y lines (which move into `flush`) and at least matching the if statement.

PR #1135, opened 2024-03-19
Fix a couple of confusions in `Chrome.paint`

pavpanchekha:
This PR makes a few text improvements to Ch 7:

- Clarify that we're building up a display list in `Chrome.paint()` and that we return it at the end
- Clarify that `DrawLine` and `DrawText` also need `rect` fields

### DIFF
--- a/book/chrome.md
+++ b/book/chrome.md
@@ -736,17 +736,28 @@ wider than numbers like 1---but it is close enough, and anyway, the
 letter X is typically as wide as the widest number.
 
 To actually draw the UI, we'll first have the browser chrome paint a
-display list, which the `Browser` will then draw to the screen. Let's
-start by first painting the new-tab button:
+display list, which the `Browser` will then draw to the screen:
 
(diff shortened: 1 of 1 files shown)


pavpanchekha:
I'm not sure why one of the code blocks is `expected=False` but I have double-checked that that's a differ bug, not wrong code.

book/chrome.md
```diff
@@ -833,18 +844,20 @@ class Chrome:
...
         cmds.append(DrawLine(
             0, self.bottom, WIDTH,
             self.bottom, "black", 1))
+        # ...
 ```
 
-I also added a line at the bottom of the chrome to separate it from
-the page. Note how I also changed `DrawRect` to pass a `Rect` instead of the
-four corners; this requires a change to `BlockLayout`:
+Make the background is drawn before any other part of the chrome. I
```
chrishtr, line 857: Remove "is?"

PR #1138, opened 2024-03-21
Explain the `Tab`/`Browser` split better

pavpanchekha:
Before there was a paragraph with vague instructions on how to split things. Instead I suggest renaming `Browser` to `Tab` and then writing a new `Browser` class. I've tested that executing these instructions by hand works.


pavpanchekha:
This confused a lot of students.

PR #1148, opened 2024-03-26
Fix timer race condition when switching tabs

chrishtr:
If one tab has an animation timer in progress when a tab is switched, then the new tab will never render, because
we don't schedule frames if `animation_timer` is set.

This is fixed by clearing `animation_timer` when changing tabs.

I was able to reproduce this in lab15 by loading a new tab, then switching back to the first tab before browser.engineering finishes loading.


pavpanchekha (approved):
Good bug catch & fix.

PR #1149, opened 2024-03-30
Fix typo and code error

morinokami:


book/text.md
```diff
@@ -559,7 +559,6 @@ self.cursor_x = HSTEP
 self.cursor_y = VSTEP
 self.weight = "normal"
 self.style = "roman"
-self.size = 16
```
morinokami, line 562: The `size` field will be introduced later at https://github.com/browserengineering/book/blob/main/book/text.md?plain=1#L633 and is not necessary for the current context.

chrishtr (approved):
Thanks!

PR #1151, opened 2024-03-30
Remove obsolete exercise (which also has a broken link)

chrishtr:
You fixed the architectural flaw in the text a few months ago.


chrishtr:
I grepped for broken footnotes in the HTML and this was the only one I found other than the one fixed earlier today.

chrishtr:
It seems pandoc has no way to warn or fail for these failure modes...

PR #1153, opened 2024-04-03
Remove `__repr__` from lab3.py

### LINKED ISSUE #1150: The outline provided in Chapter 3 does not align with the content of the chapter
In the outline provided at https://browser.engineering/text.html#outline, the `__repr__` methods for the Text and Tag classes are listed. However, it appears that there is no mention of these methods in the main text. It may be necessary to either refer to these methods in the text or remove them from the outline to ensure consistency.

morinokami:
Closes #1150.

As mentioned in #1150, the `__repr__` method is not covered in the text of Chapter 3 and, therefore, should not be included in its Outline section. Since this method is actually introduced in Chapter 4 (https://github.com/browserengineering/book/blob/main/book/html.md?plain=1#L285-L293), I deemed it appropriate to remove it from the Outline of Chapter 3.

### DIFF
--- a/src/lab3.py
+++ b/src/lab3.py
@@ -16,15 +16,9 @@ class Text:
     def __init__(self, text):
         self.text = text
 
-    def __repr__(self):
-        return "Text('{}')".format(self.text)
-
 class Tag:
     def __init__(self, tag):
         self.tag = tag
-
-    def __repr__(self):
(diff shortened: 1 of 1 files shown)


pavpanchekha:
I'm going to fix this a different way in #1154

PR #1157, opened 2024-04-11
Introduce `style_url` and `script_url` earlier

pavpanchekha:
In Chapter 10, we refer to `style_url` and `script_url` variables which aren't previously introduced. It's no harm to introduce them earlier. Since we're already here, made it so if we fail to download a script, we just skip it. (This was already the case for styles, but incongruously not for scripts.)


chrishtr (approved):
Please make sure to test the various browser versions you're changing.

PR #1159, opened 2024-04-11
Don't highlight builtins like `type`

pavpanchekha:
Our copyeditor pointed out that in Chapter 1 (and by the way same issue in Chapter 9), when we create a socket with `socket.socket(type=...)`, the word `type` should not be syntax-highlighted because it's not actually a builtin function but just a named parameter. I noticed that the website _doesn't_ have this problem. It turns out that's because we have a very minimal highlighting style on the website: comments, strings, and keywords are the only things with a color. So I made the LaTeX in the book do the same thing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,9 +44,9 @@ src/lab%.full.py: src/lab%.py infra/inline.py infra/asttools.py
 
 CHAPTER=all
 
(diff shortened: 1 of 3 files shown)


chrishtr (approved):
I'm assuming you checked the output and liked it?

pavpanchekha:
Yep, it really does look just like the website. I don't know if the colors will be quite right in print but I suppose that'll be handled during typesetting anyway.

PR #1161, opened 2024-04-12
Indicate where non-browser code blocks go

pavpanchekha:
Code blocks not from the browser now look like this:

image

The tag content comes directly from `file=` annotations so they're sort of related to our linter. They become transparent on hover to avoid occluding code, but this is pretty rare since we normally have a function header there anyway.

I added a new `.output` annotation for code blocks that show program output, to label those better.


chrishtr:
Is this ready for review?

pavpanchekha:
Yep, forgot to click the button.

book/animations.md
```diff
@@ -269,9 +269,9 @@ browser is using; this will help you verify that it's actually using
 your GPU. I'm using a Chromebook to write this chapter, so for me it
 says:[^virgl]
 
-::: {.example}
-    OpenGL initialized: vendor=b'Red Hat', renderer=b'virgl'
-:::
+``` {.output}
```
chrishtr, line 272: What's the difference between ::: and ```?

book/http.md
```diff
@@ -78,8 +78,8 @@ telnet example.org 80
 ```
 
 ::: {.web-only}
-(Note: When you see a gray outline, it means that the code in question is
-an example only, and *not* actually part of our browser's code.)
+(Note: when a code block has an "Example Code" tag, that means that
```
chrishtr, line 81: The "Trying 93.184.216.34..." block in this file is also output. Maybe there are more beyond that?

www/book.css
```diff
@@ -213,6 +212,28 @@ code .im { color: #a0a; }
 code .co { color: #C4A484; }
 }
 
+/* Tagging code blocks with where they go */
```
chrishtr, line 215: I tested this PR locally and it doesn't seem to work? Also, some code blocks look broken, like the attached example.
Screenshot 2024-04-12 4 36 29 PM

book/animations.md
pavpanchekha: I think basically the ``` means a `<pre>` tag and a ::: means a `<div>` tag. The reason we were getting typewriter font before was that indenting by four spaces also means a `<pre>` tag. (Also why I have now out-dented them.)

pavpanchekha:
Ok, I did a full pass through every code block in the book and there were quite a lot to fix up, but it should now work better.

Along the way I made a few unrelated changes, apologies.

book/accessibility.md
```diff
@@ -1132,7 +1133,7 @@ If you print out `focusable_nodes` for the
 [focus example](examples/example14-focus.html), you should
 get this:
 
-``` {.python .example}
+``` {.output}
```
chrishtr, line 1136: .python also?

book/scheduling.md
```diff
@@ -177,7 +177,7 @@ we'll implement `setTimeout` by saving the callback on in a JavaScript
 variable and creating a handle by which the Python-side code can call
 it:
 
-``` {.javascript file=runtime}
```
chrishtr, line 180: runtime is no longer necessary because it's the default?

src/runtime12.js
```diff
@@ -93,10 +93,7 @@ function requestAnimationFrame(fn) {
 }
 
 function __runRAFHandlers() {
-    var handlers_copy = [];
```
chrishtr, line 96: Why the change in this file? If it's a separate fix please make a separate PR.

www/book.css
```diff
@@ -213,6 +213,32 @@ code .im { color: #a0a; }
...
+pre.example::after, pre.output::after, .sourceCode[data-file]::after {
+    position: absolute; top: 0; right: 0; color: #888;
+    font: bold 83% 'Vollkorn', 'Garamond', serif;
+    padding: 1px 1ex; border: 2px solid #aaa;
+    pointer-events: none; /* So you can select / scroll under tag */
+}
+
+/* Note that ::after:hover is not supported in CSS so this is best we can do */
+pre.example:hover::after, pre.output:hover::after, .sourceCode[data-file]:hover::after {
+    opacity: 30%; /* In case it covers some code */
```
chrishtr, line 228: How about putting it above the code and making space? I find the hover affect distracting and annoying.

Please test on mobile also.

chrishtr:
In http.md the headers are stated as "example code" which is not quite right. They are "examples" but not code?

chrishtr:
Add a footnote to server.py explaining what we're doing indicating the server vs client code.

www/book.css
```diff
@@ -213,6 +213,32 @@ code .im { color: #a0a; }
...
+}
+
+pre.example::after { content: "Example code"; }
+pre.output::after { content: "Example output"; }
+.sourceCode[data-file=server]::after { content: "server.py" }
```
chrishtr, line 233:
```suggestion
.sourceCode[data-file=server]::after { content: "Server code" }
```

chrishtr (commented):
Does

chrishtr:
Also make it work in the print edition?

pavpanchekha:
We discussed today in meeting and decided to pause this until typesetting starts in earnest so we can sync printed & online book.

pavpanchekha:
Superseded by #1343. The new plan is to go with simpler styling that looks more like the book: gray background and border for examples, otherwise no tag.

PR #1166, opened 2024-04-13
Editorial feedback from readers on the style chapter

chrishtr:


src/lab6.py
```diff
@@ -249,6 +249,14 @@ def paint(self):
                                          word, font, color))
         return cmds
 
+    @wbetools.js_hide
```
chrishtr, line 252: These methods were removed in this chapter due to a refactor; this seems a reasonable way to get rid of them from the outline.

Makefile
```diff
@@ -69,7 +69,7 @@ latex/book.pdf: latex/book.tex latex/macros.tex
 	(cd latex && ln -f -s ../www/im/ im)
 	(cd latex && pdflatex book.tex)
 
-www/%.html: book/%.md infra/template.html infra/signup.html infra/filter.lua config.json
+www/%.html: book/%.md infra/template.html infra/signup.html infra/filter.lua config.json src/lab*.py
```
pavpanchekha, line 72: What does this do? Also, maybe nothing?

src/lab6.py
pavpanchekha: Yeah it would be nice to have something nice but this is definitely a good fix for now.

pavpanchekha (approved):

Makefile
chrishtr: The book chapters depend on the source code for the outline. This addition expresses that for this particular rule.

PR #1170, opened 2024-04-15
Change `top_level_url` to `referrer` (close #1155)

### LINKED ISSUE #1155: Can we rename `top_level_url` to `referer` or `referrer`?
Right now, the `URL.request` method has a `top_level_url` parameter, which is added in Chapter 10 to implement `SameSite` cookies. I'm trying to figure out if it would be clearer and easier to call this a `referrer`. The challenge I'm struggling with is whether the `top_level_url` and the `referrer` are actually the same concept.

Standard defining `Referer`: https://www.rfc-editor.org/rfc/rfc9110#field.referer

> specify a URI reference for the resource from which the [target URI](https://www.rfc-editor.org/rfc/rfc9110#target.resource) was obtained

Standard defining `SameSite`: https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-cookie-same-site-00#section-2.1

>  For a document displayed in a top-level browsing context, we can stop
>   here: the document's "site for cookies" is the top-level site.

As usual the standards are somehow still too vague (what does it mean that a URI "was obtained" from a particular resources) but here's the key difference: if `a.html` links to `b.css` which refers to `c.jpg`, then the referrer for `c.jpg` is `b.css` but its `top_level_url` is `a.html`. Another place a difference pops up is with nested, cross-origin iframes; in this case, to a first approximation requests in nested frames always fail `SameSite` checks but they may still have referrers.

Anyway, my questions:
- In our browser, the `top_level_url` is actually the referrer; should we rename it?
- Should we handle the (very ugly) nested iframe check in Chapter 15 when we introduce the iframes? (If not, we should footnote it.)

pavpanchekha:
This PR closes #1155 by renaming `top_level_url` to `referrer` and leaving a footnote noting the difference.

### DIFF
--- a/book/animations.md
+++ b/book/animations.md
@@ -269,9 +269,9 @@ browser is using; this will help you verify that it's actually using
 your GPU. I'm using a Chromebook to write this chapter, so for me it
 says:[^virgl]
 
-::: {.example}
-    OpenGL initialized: vendor=b'Red Hat', renderer=b'virgl'
-:::
+``` {.output}
+OpenGL initialized: vendor=b'Red Hat', renderer=b'virgl'
+```
 
 [^virgl]: The `virgl` renderer stands for "virtual GL", a way of
 hardware-accelerating the Linux subsystem of ChromeOS that works with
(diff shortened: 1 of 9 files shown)


chrishtr:
Looks like you accidentally based this on top of the other CL that reformats code blocks.

PR #1171, opened 2024-04-16
Don't fill out a `display_list` field on every layout object

pavpanchekha:
In Chapter 5, we currently make each `BlockLayout` fill out a `display_list` field during the `layout` pass, and then later convert that to a separate paint pass. This is confusing and we just don't need to do it—just don't do the display list thing until we add the paint pass later.

### DIFF
--- a/book/layout.md
+++ b/book/layout.md
@@ -55,7 +55,6 @@ class Browser:
         # ...
         self.document = Layout(self.nodes)
         self.document.layout()
-        self.display_list = self.document.display_list
         #...
 ```
 
@@ -95,7 +94,7 @@ call it `DocumentLayout`:
 the root layout object also computes its size and position
 differently, as we'll see later in this chapter.
 
-``` {.python replace=%20Layout/%20BlockLayout dropline=display_list}
+``` {.python replace=%20Layout/%20BlockLayout}
 class DocumentLayout:
     def __init__(self, node):
         self.node = node
@@ -106,7 +105,6 @@ class DocumentLayout:
         child = Layout(self.node, self, None)
         self.children.append(child)
         child.layout()
-        self.display_list = child.display_list
 ```
 
 Note an interesting thing about this new `layout` method: its role is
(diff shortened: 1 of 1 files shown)


book/layout.md
```diff
@@ -55,7 +55,6 @@ class Browser:
         # ...
         self.document = Layout(self.nodes)
         self.document.layout()
-        self.display_list = self.document.display_list
```
chrishtr, line 58: This is in lab5.py. Isn't it still needed?

book/layout.md
```diff
@@ -310,9 +297,8 @@ complex layout trees!
 ![An example of an HTML tree and the corresponding layout tree](im/layout-tree.png)
 :::
 
-Oh, and you might also notice that the text on these web pages is now
```
chrishtr, line 313: Is this obsolete?

book/layout.md
```diff
@@ -281,17 +279,6 @@ class BlockLayout:
             child.layout()
 ```
 
-We also need to gather their `display_list` fields into a single
-array:
-
-``` {.python expected=False}
-class BlockLayout:
```
chrishtr, line 288: This code blocks just looks stale, so good to remove?

book/layout.md
pavpanchekha: Is it? The corresponding code in `lab5.py` is:
```
class Browser:
    def load(self, url):
        # ...
        self.display_list = []
        paint_tree(self.document, self.display_list)
        # ...
```

book/layout.md
pavpanchekha: So, in the old version of the chapter, we add the layout tree and initially implement a "stop-gap" painting strategy (wherein each node paints itself during layout, and propagates display lists up three tree). Then a little later, we add a proper, separate paint phase.

In the proposed new version, we just never implement the stopgap. The benefit is we don't have to describe this straw-man stop-gap and also you don't have to remember to take it out. The downside is that the browser now doesn't show anything until we implement paint properly—so this observation about all the text overlapping doesn't make sense anymore. Hence it is removed. It's possible to maybe get the benefits of both, by moving painting up and then rejiggering the whole chapter, but I wasn't really confident in attempting that this late stage in editing.

book/layout.md
pavpanchekha: Yes, this implements that "stop-gap" described above.

PR #1174, opened 2024-04-17
Make font sizes in all chapters use pixels and points consistently

### LINKED ISSUE #1173: Change Chapter 3–5 text size to 12pt
In Chapter 3, we compute the size via a simple integer field:

    self.size = 16
    # ...
    font = tkinter.font.Font(
        size=self.size,
        weight=self.weight,
        slant=self.style,
    )

Here the `size` argument to `Font` is in points, not pixels; I confirmed that in the documentation but also by taking screenshots and counting the pixels.

In Chapter 6, we switch to using CSS:

    size = int(float(node.style["font-size"][:-2]) * .75)

Note the 0.75, which converts the pixel value into points. The default is:

    INHERITED_PROPERTIES = {
        "font-size": "16px",
        # ...
    }

The issue is that in Chapter 3, we have a 16 point default size, and then in Chapter 6 we switch to a 16 pixel, meaning 12 point, font, making all the text smaller. (This is frustrating my efforts to build a single browser that passes all tests from all chapters.)

My proposal: change Chapter 3 to do size 12 by default. (This will change a lot of tests, ugh.) I'm not even sure it needs a footnote or any text, there's no discussion of that default size right now.

pavpanchekha:
Fix #1173 by changing default font size in Chapters 3–5 to be 12pt meaning 16px, just like the default later on. All tests updated (and a few are reformatted to look nicer).


book/accessibility.md
```diff
@@ -322,12 +322,12 @@ class BlockLayout:
 	# ....
     def word(self, node, word):
     	# ...
-        size = dpx(float(node.style["font-size"][:-2]),
+        size = dpx(float(node.style["font-size"][:-2]) * 0.75,
```
chrishtr, line 325: Was the previous code here and in various other cases just wrong? In chapter 6 we multiply by 0.75, but I guess that was incorrectly lost in later chapters?

chrishtr: Indeed - skia.Font expects points, not pixels.

pavpanchekha: Yes, it really looks like!

pavpanchekha: It was hard to notice because we always test on the book website, which specifies its own font size in CSS, and then I guess we'd just never compared Skia and Tk renderings to notice the font size difference.

chrishtr (approved):
This is a bad bug, I'm glad you found it!

PR #1177, opened 2024-04-30
Always create and indent `skia.Paint` objects in the same way

pavpanchekha:
This PR makes sure we always use `skia.Paint` in the same way, namely like this:

```
paint = skia.Paint(
    prop=value,
    ...
)
canvas.drawXXX(xxx, paint)
```

In other words, we now:

- Always create the whole `Paint` object at once, instead of using `setXXX` methods
- Always pass it to `drawXXX` as a positional argument, not a keyword one
- Always put the property/value pairs one per line with a comma at the end

The specific style isn't actually important (willing to discuss other options; I wouldn't be against, for example, never creating `Paint` objects at all and always using convertability from dicts) but just that we always do the same thing the same way so that readers aren't confused.

### DIFF
--- a/book/visual-effects.md
+++ b/book/visual-effects.md
@@ -434,9 +434,9 @@ class DrawLine:
         path = skia.Path().moveTo(self.x1 - scroll, self.y1) \
                           .lineTo(self.x2 - scroll, self.y2)
         paint = skia.Paint(
-            Style=skia.Paint.kStroke_Style,
-            StrokeWidth=self.thickness,
             Color=parse_color(self.color),
+            StrokeWidth=self.thickness,
+            Style=skia.Paint.kStroke_Style,
         )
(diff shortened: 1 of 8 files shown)


book/visual-effects.md
```diff
@@ -433,9 +433,11 @@ class DrawLine:
     def execute(self, canvas, scroll):
         path = skia.Path().moveTo(self.x1 - scroll, self.y1) \
                           .lineTo(self.x2 - scroll, self.y2)
-        paint = skia.Paint(Color=parse_color(self.color))
-        paint.setStyle(skia.Paint.kStroke_Style)
-        paint.setStrokeWidth(self.thickness)
+        paint = skia.Paint(
```
chrishtr, line 436: Sort argument names alphabetically?

PR #1181, opened 2024-05-12
Make exercises numbered by chapter-number for http.md

chrishtr:
The copy edited book numbers exercises by <chapter>-<exercise number> so that you can easily refer to them by unique number, even across chapters.

Another option is to number them only by their chapter-relative number (i.e 1,2 ... ) on the web and have the first part be implicit.

The copy editor also referred to exercises like this:

Declare exercise:

```
\begin{enumerate}[label=\thechapter-\arabic*]
\item\label{ex:HTTP/1.1} \emph{\texttt{HTTP/1.1}.}
```

refer to it:

```
Exercise~\ref{ex:HTTP/1.1}
```

We could also as a second step do the same.

### DIFF
--- a/book/http.md
+++ b/book/http.md
@@ -893,7 +893,7 @@ can:
 -   connect to that host using the `sockets` and `ssl` libraries;
 -   send an HTTP request to that host, including a `Host` header;
 -   split the HTTP response into a status line, headers, and a body;
--   print the text (and not the tags) in the body.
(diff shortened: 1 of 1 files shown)


chrishtr:
If you approve this PR then I'll do the same for chapters 2-16.

pavpanchekha (approved):
Let’s do it & later we can think about adding permalinks or nice styling.

PR #1183, opened 2024-05-13
Set a maximum line width

pavpanchekha:
This PR adds an enforced maximum line width, 70 characters, to code blocks in the book. There's only one code block in the book that fails this test (and currently causes a horizontal scroll bar); I updated it and three lookalikes.

In case you're interested, here are the current statistics line length statistics, from before this fix:

| Count | Length |
|-------|--------|
| 56 | 61 |
| 46 | 62 |
| 30 | 63 |
| 35 | 64 |
| 27 | 65 |
| 30 | 66 |
| 21 | 67 |
| 16 | 68 |
| 20 | 69 |
| 10 | 70 |
|  1 | 71 |

### DIFF
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -358,8 +358,8 @@ class TextLayout:
 	# ...
     def layout(self):
     	# ...
-        size = dpx(
-        	float(self.node.style["font-size"][:-2]) * 0.75, self.zoom)
+        px_size = float(self.node.style["font-size"][:-2])
+        size = dpx(px_size * 0.75, self.zoom)
 ```
 
 And the fixed `INPUT_WIDTH_PX` for text boxes:
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -319,26 +319,23 @@ We'll do this conversion to adjust the font sizes in the `text` and
 
 ``` {.python}
 class BlockLayout:
-	# ....
     def word(self, node, word):
     	# ...
-        size = dpx(float(node.style["font-size"][:-2]) * 0.75,
-            self.zoom)
+        px_size = float(node.style["font-size"][:-2])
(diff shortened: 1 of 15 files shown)


pavpanchekha:
My thinking is that the couple of lines 70–80 characters long are probably mistakes, or special cases, and should be handled somehow, but looking at a bunch of the 60-character lines they really do look like normal Python code where shrinking them would look pretty weird! Like, these three lines are all 61 characters:

```
        self.font = ProtectedField(self, "font", self.parent,
        if self.node.is_focused and self.node.tag == "input":
                    elif event.key.keysym.sym == sdl2.SDLK_a:
    response += "Content-Security-Policy: {}\r\n".format(csp)
```

I'll go through these long lines from longest first and see what I find.

pavpanchekha:
Update—the table above was wrong (it was after text replacement and re-indentation was applied, not before) and there are now significantly fewer problematic lines, now spanning 61–71 characters. I'm still not sure that it's possible to go down to 61 characters but I think 8pt font should give us about 70 characters (which should be quite doable) and 8.5pt should give us about 65 (which is probably also achievable).

pavpanchekha:
Ok, and by chapter:


| Chapter | Too long |
|-|-|
| 1 | 3 |
| 2 | 0 |
| 3 | 4 |
| 4 | 3 |
| 5 | 2 | 
| 6 | 13 |
| 7 | 16 |
| 8 | 6 |
| 9 | 8 |
| 10 | 15 |
| 11 | 15 |
| 12 | 10 |
| 13 | 13 |
| 14 | 38 |
| 15 | 41 |
| 16 | 47 |

Basically most of the work is Chapters 14, 15, and 16; everything else can probably be done quickly by hand.

chrishtr (approved):

PR #1185, opened 2024-05-15
Number figures in the history.md and http.md chapters

chrishtr:


chrishtr:
If this pattern looks ok I'll repeat it for the other chapters.

pavpanchekha (approved):

PR #1186, opened 2024-05-15
Apply copy editor changes to scripts.md

chrishtr:


book/scripts.md
```diff
@@ -368,7 +370,7 @@ browser's console output?) We need to allow JavaScript to modify the
 page.
 
 JavaScript manipulates a web page by calling any of a large set of
-methods collectively called the DOM API, for "Document Object Model".
```
chrishtr, line 371: I'm assuming they proposed this edit because html.md defines this term.

pavpanchekha (approved):

pavpanchekha:
I'm not 100% sure we have an italic code font on the web (probably we get a fake browser fallback) but we can fix that later.

PR #1192, opened 2024-05-17
Apply copy editor changes to animations.md

chrishtr:


book/animations.md
```diff
@@ -352,16 +352,16 @@ width and height arguments).
 [^color-space]: Example detail: a different color space is required
 for GPU mode.
 
-Thanks to SDL's and Skia's thorough support for GPU rendering, that should be
+Thanks to the thorough support in SDL and Skia for GPU rendering, that should be
```
pavpanchekha, line 355:
```suggestion
Thanks to the thorough support for GPU rendering in SDL and Skia, that should be
```

book/animations.md
```diff
@@ -1455,8 +1458,8 @@ class Browser:
             self.draw()
 ```
 
-Then, call `set_needs_raster` from the places that currently call
-`set_needs_raster_and_draw`, such as `handle_down`:
+Then, were we currently call `set_needs_raster_and_draw`, such as
```
pavpanchekha, line 1461:
```suggestion
Then, where we currently call `set_needs_raster_and_draw`, such as
```

pavpanchekha: I think "such as `handle_down`" maybe should be in a parenthetical, but if that's a copyeditor suggestion I'm willing to use commas.

book/animations.md
```diff
@@ -1743,8 +1747,8 @@ Here is how the opacity transition example's composited layers should look
 [flag]: https://docs.python.org/3/library/argparse.html
 
 [^flag-recommendation]: I also recommend you add a mode to your browser that
-disables compositing (i.e. setting `needs_compositing` to `False` for every
-`VisualEffect`), and disables use of the GPU (i.e. going back to the old way of
+disables compositing (i.e., setting `needs_compositing` to `False` for every
+`VisualEffect`), and disables use of the GPU (i.e., going back to the old way of
```
pavpanchekha, line 1751: These two (plus one in the glossary) are the only use of `i.e.` in the book. Can we replace with "that is"?

pavpanchekha: I also looked for `e.g.` and there a few more of those that look harder to replace. But "i.e." here is very replacable.

book/animations.md
```diff
@@ -2023,7 +2031,7 @@ To implement `local_to_absolute`, we first need a new `map` method on
 "contents" of the transform and outputs a rect in post-transform
 space. For example, if the transform was `translate(20px, 0px)` then
 the output of calling `map` on a rect would translate it by 20 pixels
-in the *x* direction.
+in the `x` direction.
```
pavpanchekha, line 2034: Is this right? I think the copyeditor often suggests replacing `*x*` with `$x$`, which I've largely been ignoring (for single letters it will look the same), but backticks for code font are only correct if we're referring to a variable; here we're referring to a coordinate direction.

book/animations.md
```diff
@@ -2073,35 +2081,36 @@ testing is now complete. You should now be able to render
...
+![Figure 8: Wrong rendering because overlap testing is missing](examples/example13-transform-overlap-wrong1.png)
 :::
 
 * Reusing cloned effects (without it, blending and clipping would be wrong);
-if this code were missing it would incorrectly render like this:
+if this code were missing it would incorrectly render like Figure 9.
 
 ::: {.center}
-![Wrong rendering because of incorrect blending](examples/example13-transform-overlap-wrong2.png)
+![Figure 9: Wrong rendering because of incorrect blending](examples/example13-transform-overlap-wrong2.png)
```
pavpanchekha, line 2113: For typesetting it would be nice to put this series of images in a single figure with subfigures. But that's not for right now.

book/animations.md
```diff
@@ -2119,7 +2128,7 @@ each display item:
 [^clipping-notes]: This is very important, because otherwise some
 composited layers can end up huge despite not drawing much to the screen.
 A good example of this optimization making a big difference is loading the
-browser from [Chapter 15](embeds.md) for the `browser.engineering` homepage,
+browser from [Chapter 15](embeds.md) for the browser.engineering homepage,
```
pavpanchekha, line 2131: I think if we write `http://browser.engineering/` it'll render a hyperlink in builds of the book where that's possible.

book/animations.md
```diff
@@ -2242,7 +2251,7 @@ function, and one or two others.
 
 [easing]: https://developer.mozilla.org/en-US/docs/Web/CSS/easing-function
 
-13-3 *Composited & threaded transform and scroll animations*. Our browser supports
+13-3 *Composited and threaded transform and scroll animations*. Our browser supports
```
pavpanchekha, line 2254: Is it possible to shorten this exercise name? It's way longer than any other.

book/animations.md
```diff
@@ -2283,8 +2292,8 @@ CSS property and parsing of `@keyframe` to implement the demos
 [here](examples/example13-opacity-animation.html) and
 [here](examples/example13-width-animation.html).
 
-13-6 *Overlap testing w/transform animations*. (If you've already done the
-transform animations exercise.) Our
+13-6 *Overlap testing w/transform animations*. (If you've already done
```
pavpanchekha, line 2295: I think this is the only use of `w/` in the book. Can we rename somehow?

pavpanchekha: Also starting with a parenthetical is also unique. Can we move the parenthetical one sentence later and rephrase as something like "You'll need to have done Exercise 13-3"?

pavpanchekha (changes requested):

pavpanchekha (approved):

PR #1195, opened 2024-05-17
Copyedit Chapter 1

pavpanchekha:
This chapter was the test copyedit chapter, so there are fewer changes. But still a lot, to account for the fact that we've finally figured out how we'll do figures.

### DIFF
--- a/book/http.md
+++ b/book/http.md
@@ -13,8 +13,7 @@ Connecting to a server
 ======================
 
 Browsing the internet starts with a URL\index{URL},[^url] a short string that
-identifies a particular web page that the browser should visit. A URL
-looks like this:
+identifies a particular web page that the browser should visit.
 
 [^url]: "URL" stands for "uniform resource locator", meaning that it
     is a portable (uniform) way to identify web pages
@@ -35,7 +34,7 @@ Figure 1: The syntax of URLs
 ![Figure 1: The syntax of URLs](im/http-url.png)
 :::
 
-A URL has three parts: the scheme\index{scheme} explains *how* to get the
+A URL has three parts (see Figure 1): the scheme\index{scheme} explains *how* to get the
 information; the host name explains *where* to get it; and the path\index{path}
 explains *what* information to get. There are also optional parts to
 the URL, like ports, queries, and fragments, which we'll see later.
@@ -142,7 +141,7 @@ Requesting information
 Once it's connected, the browser requests information from the server
 by giving its *path*, the path being the part of a URL that comes
 after the host name, like `/index.html`. The structure of the request
-is shown below. You can type it into `telnet` (make sure to type a blank line after the `Host` line):
+is shown in Figure 2. You can type it into `telnet` to try it.
 
 ::: {.cmd .web-only html=True}
     python3 infra/annotate_code.py <<EOF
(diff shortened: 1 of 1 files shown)


book/http.md
```diff
@@ -142,7 +141,7 @@ Requesting information
 Once it's connected, the browser requests information from the server
 by giving its *path*, the path being the part of a URL that comes
 after the host name, like `/index.html`. The structure of the request
-is shown below. You can type it into `telnet` (make sure to type a blank line after the `Host` line):
+is shown in Figure 2. You can type it into `telnet` to try it.
```
pavpanchekha, line 144: Copyeditor said "you can type this" instead of "you can type it" but I think "it" makes more sense.

book/http.md
```diff
@@ -161,9 +160,9 @@ information,[^11] then comes the path, and finally there is the word
 `HTTP/1.0` which tells the host that the browser speaks version 1.0 of
 [HTTP]\index{HTTP}. There are several versions of HTTP ([0.9, 1.0, 1.1, 
 2.0, and 3.0](https://medium.com/platform-engineer/evolution-of-http-69cfe6531ba0)).
-The HTTP 1.1 standard adds a variety of useful features, like
+The HTTP 1.1 standard adds a variety of useful features, like
 keep-alive, but in the interest of simplicity our browser won't use
-them. We're also not implementing HTTP 2.0; it is much more
+them. We're also not implementing HTTP 2.0; it is much more
```
pavpanchekha, line 165: The chance is from a normal space to a non-breaking space.

book/http.md
```diff
@@ -199,7 +198,8 @@ HTTP was designed to be simple to understand and implement, making it
 easy for any kind of computer to adopt it. It's no coincidence that
 you can type HTTP directly into `telnet`! Nor is it an accident that
 HTTP is a "line-based protocol", using plain text and newlines,
-similar to [Simple Mail Transfer Protocol (SMTP)] for email. Ultimately, the whole pattern derives
+similar to Simple Mail Transfer Protocol ([SMTP][SMTP]) for email.
```
pavpanchekha, line 201: This link was broken, now fixed.

book/http.md
```diff
@@ -736,7 +737,7 @@ Encrypted connections
 =====================
 
 So far, our browser supports the `http` scheme. That's a pretty common
-scheme. But more and more, websites are migrating to the
+scheme. But more and more websites are migrating to the
```
pavpanchekha, line 740: Curiously, both "But, more and more," and also "But more and more" are grammatically correct, mean different things, and also those things are both true. I couldn't decide which one we really meant and went with this one, which is what the copyeditor suggested.

book/http.md
```diff
@@ -936,7 +939,7 @@ make it so that, if your browser is started without a URL being given,
 some specific file on your computer is opened. You can use that file
 for quick testing.
 
-1-3 *`data:` scheme*. Yet another scheme is `data`, which allows inlining
+1-3 *`data`*. Yet another scheme is `data`, which allows inlining
```
pavpanchekha, line 942: If we want the colon and the word "scheme" we need to update `view-source` below, but I think it's OK without the "scheme" and there's value in keeping exercise names short.

pavpanchekha:
Still needs review

book/http.md
```diff
@@ -210,7 +210,7 @@ fact, one of the first two browsers had a [line-mode UI][line-mode].
 The server's response
 =====================
 
-The server's response starts with the line below:
+The server's response starts with the line in Figure 3.
```
chrishtr, line 213: Change figure 3 to `{.center}` instead of `{.print-only}`. Same for figure 2 and figure 6.

chrishtr: Figures 4 and 5 also need to be surrounded by `:::`

pavpanchekha: I think this isn't quite the right fix for 2/3/6 (these are "annotated code" blocks, on the web we can use the HTML form instead of the image) but I fixed them to show the caption properly and only on the web, which they did do wrong.

PR #1196, opened 2024-05-17
Apply copy editor changes to embeds.md

chrishtr:


book/embeds.md
```diff
@@ -572,7 +572,7 @@ modalities.
 
 ::: {.further}
 Videos are similar to images, but demand more bandwidth, time, and
-memory; they also have complications like [Digital Rights Management
+memory; they also have complications like [digital rights management
 (DRM)][drm]. The `<video>` tag addresses some of that, with built-in
```
pavpanchekha, line 576: One thing I've noticed the copyeditor prefers "foo bar baz (FBB [N])", in other words they prefer just the acronym to be linked. But generally on the web I think it looks better to link the whole phrase, and in print it doesn't matter much. But we should standardize.

book/embeds.md
```diff
@@ -1290,10 +1273,10 @@ class Tab:
 ```
 
 When an iframe is clicked, it passes the click through to the child
-frame, and immediately return afterwards, because iframes capture
-click events. Note how I subtracted the absolute x and y offset of the
-iframe from the (absolute) x and y click position when recursing into the child
-frame:
+frame, and immediately returns afterward, because iframes capture
+click events. Note how I subtracted the absolute `x` and `y` offsets of the
```
pavpanchekha, line 1277: I think in my chapters I saw the copyeditor want "x" and "y" to be italics, as in `*x*`, not code font.

book/embeds.md
```diff
@@ -1317,17 +1300,18 @@ all the loading and navigation logic that `Tab` used to have, it just
 works without any more changes!
 
 You should now be able to load [this
-example](examples/example15-iframe.html). It should look like this:
+example](examples/example15-iframe.html). It should look like the image
+shown in in Figure 5.
```
pavpanchekha, line 1304:
```suggestion
shown in Figure 5.
```

book/embeds.md
```diff
@@ -2231,14 +2215,15 @@ used to [construct a clock][sab-attack].
 
 ::: {.further}
 The `SharedArrayBuffer` issue caused problems when I [added JavaScript
-support][js-blog] to the embedded browser widgets on this website. I
+support][js-blog] to the embedded browser widgets on
+[the book's website](https://browser.engineering). I
 was using `SharedArrayBuffer` to allow synchronous calls from a
 `JSContext` to the browser, and that required APIs that browsers
 restrict for security reasons. Setting the security headers wouldn't
-work, because [Chapter 14](accessibility.md) embeds a Youtube video,
```
pavpanchekha, line 2238: I kept the links. I think we'll strip them out with Lua for LaTeX and keep them for HTML.

pavpanchekha (changes requested):

book/embeds.md
chrishtr: agree; what do you prefer I do here?

pavpanchekha: I think we should do whole phrase (keep what you have here) and add a bullet point to the Google Doc so when we do a final pass over things (after merging in all copyeditor changes) we make it look like this one.

book/embeds.md
chrishtr: Done

book/embeds.md
chrishtr: In this case I didn't think it really adds value to the web version, so removed it.

pavpanchekha (approved):

PR #1201, opened 2024-05-18
Apply copy editor changes to bibligraphy.md

chrishtr:


pavpanchekha (approved):
The diff is basically unreviewable but I figure there are no substantial changes other than formatting? Definition lists are kind of an odd choice but if it works it works.

PR #1209, opened 2024-05-20
Filter out internal links in print format

pavpanchekha:
This PR modifies `filter.lua` to drop internal links when building in print mode. It also fixes two links that I found when testing. I tested it and the PDF indeed no longer has the links. I also examined the list of filtered-out links and they were all good. (I believe `classes.md` isn't in the book itself, so I've left its link to `index.md` in place.)

### DIFF
--- a/book/preface.md
+++ b/book/preface.md
@@ -45,8 +45,7 @@ to the implementation you have built, the book says "your browser".
 [^jrwilcox-idea]: This idea is from [J.R. Wilcox][jrw], inspired in
 turn by [S. Zdancewic][sz]'s course on compilers.
 
-The code in this book [uses Python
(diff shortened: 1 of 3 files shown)


book/preface.md
```diff
@@ -45,8 +45,7 @@ to the implementation you have built, the book says "your browser".
 [^jrwilcox-idea]: This idea is from [J.R. Wilcox][jrw], inspired in
 turn by [S. Zdancewic][sz]'s course on compilers.
 
-The code in this book [uses Python
-3](blog/why-python.md),\index{Python} and we recommend you follow
```
chrishtr, line 49: just change this to https://browser.engineering/blog/why-python.html instead.

pavpanchekha: Done, but I used the substack link instead.

Issue #1210, opened 2024-05-20
Outline bugs


pavpanchekha:
These are all functions/methods added in a chapter and seemingly important but not described in the text. **These are critical and we need to resolve them stat!** 

- `Browser.clear_pending_tasks` in Chapter 12
- `Browser.new_tab_internal` in Chapter 12
- `Browser.clear_data` in Chapter 13
- `JSContext.dispatch_RAF` in Chapter 15

I think `dispatch_RAF` is probably easy and just needs to be mentioned by name in the text. `new_tab_internal` possibly can do the same. but the two `clear` ones fix real bugs and we don't describe it at all—needs a fix fast!

pavpanchekha:
@chrishtr: can you take a look at this?

PR #1211, opened 2024-05-20
Clean outline

pavpanchekha:
This PR cleans up the outline somewhat, removing obsolete methods, hiding some debugging functions, and so on.

I still plan to do the following in follow-up PRs:

- Make a "template order" for the outline so that the functions / classes are always in a sane order
- Include superclasses in class headers
- Use `diff` to check again that chapter-by-chapter we're adding/removing the right functions (didn't accidentally remove something by forgetting to import it)

See also #1210, which I found while working on the outline.

### DIFF
--- a/infra/compile.py
+++ b/infra/compile.py
@@ -357,6 +357,8 @@ def compile_method(base, name, args, ctx):
         return base_js + "." + name + "(" + ", ".join(args_js) + ")"
     elif base_js in RT_IMPORTS:
         return base_js + "." + name + "(" + ", ".join(args_js) + ")"
+    elif name == "read" and isinstance(base, ast.Call) and isinstance(base.func, ast.Name) and base.func.id == "open":
+        return base_js + ".read()"
     elif name == "keys":
         assert len(args) == 0
         return "Object.keys(" + base_js + ")"
--- a/infra/outlines.py
+++ b/infra/outlines.py
@@ -92,6 +92,12 @@ def write_html(objs, indent=0):
         print("</code>")
 
 def to_item(cmd):
+    if hasattr(cmd, "decorator_list") and any([
+            isinstance(dec, ast.Attribute) and dec.attr == 'outline_hide'
+            and isinstance(dec.value, ast.Name) and dec.value.id == "wbetools"
+            for dec in cmd.decorator_list
+    ]):
+        return None
     if isinstance(cmd, ast.ClassDef):
         return Class(cmd.name, [to_item(scmd) for scmd in cmd.body])
     elif isinstance(cmd, ast.FunctionDef):
--- a/infra/outlines.py
+++ b/infra/outlines.py
@@ -21,7 +21,7 @@ class Function(Item):
         return "def {}({})".format(self.name, ", ".join(args))
 
     def html(self):
-        return self.str().replace("def", "<span class=kw>def</span>")
+        return "<span class=kw>def</span> {}({})".format(self.name, ", ".join(args))
 
     def sub(self):
         return None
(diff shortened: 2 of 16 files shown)


pavpanchekha:
I will test all widgets before I merge this.

infra/compile.py
```diff
@@ -785,7 +787,12 @@ def compile(tree, ctx, indent=0, patches={}, patchables={}):
         return out
     elif isinstance(tree, ast.ClassDef):
         if tree.decorator_list:
-            assert tree.decorator_list[0].func.value.id == "wbetools"
+            assert isinstance(tree.decorator_list[0], ast.Call) and \
```
chrishtr, line 790: What is this logic change checking for?

src/lab14.py
```diff
@@ -1540,11 +1537,13 @@ def speak_document(self):
         else:
             print(text)
 
+    @wbetools.js_hide
```
chrishtr, line 1540: Why should these be hidden?

src/lab16.py
```diff
@@ -475,20 +476,14 @@ def self_rect(self):
...
         cmds = []
-
-        if not self.is_atomic():
-            bgcolor = self.node.style["background-color"].get()
-            if bgcolor != "transparent":
-                radius = dpx(
-                    float(self.node.style["border-radius"].get()[:-2]),
-                    self.zoom.get())
-                cmds.append(DrawRRect(self.self_rect(), radius, bgcolor))
+        bgcolor = self.node.style["background-color"].get()
```
chrishtr, line 481: Why did you change this implementation?

infra/compile.py
pavpanchekha: Allow a class to be decorated with a hide instruction

infra/outlines.py
```diff
@@ -21,7 +21,11 @@ def str(self):
         return "def {}({})".format(self.name, ", ".join(args))
 
     def html(self):
-        return self.str().replace("def", "<span class=kw>def</span>")
+        if len(self.args) > 0 and self.args[0] == "self":
```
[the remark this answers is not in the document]
pavpanchekha: Avoid highlighting "def" in the outline if it's part of a method name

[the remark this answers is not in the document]
pavpanchekha: This doesn't happen but the analog with class does in PseudoclassSelector

src/lab13.py
```diff
@@ -780,8 +780,6 @@ def __init__(self, tab):
             self.XMLHttpRequest_send)
         self.interp.export_function("setTimeout",
             self.setTimeout)
-        self.interp.export_function("now",
```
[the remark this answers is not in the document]
pavpanchekha: Yep. It was in drafts of the animations chapter and we seen to have never gotten rid of it

src/lab16.py
pavpanchekha: Ah yeah. This is a pretty dumb bug, and the fix used to be this is_atomic function, but then later we changed how it worked and it is now is supposed to use should_paint. Looks like we never updated this chapter

src/lab14.py
pavpanchekha: We never discuss them in the chapter. Perhaps that means we should just eliminate the code completely but I wasn't sure if it was useful in debugging so I left them in for now

infra/outlines.py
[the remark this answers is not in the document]
pavpanchekha: No, I mean, there's an actual bug, this and the other line fix it.

chrishtr (approved):

chrishtr:
Looks like you're failing some unittests..

PR #1214, opened 2024-05-21
Add more missing imports

pavpanchekha:
I found these through a pretty ugly mechanism, which I don't want to commit in this PR (will make a separate PR to track progress) but merging these outline fixes seems like a priority.

With this PR, I am now fairly confident that the outlines are correct. Remaining issues I found are in #1210 and need to be resolved urgently.


infra/outlines.py
```diff
@@ -57,21 +57,6 @@ def html(self):
     def sub(self):
         return None
 
-@dataclass
-class IfMain:
```
chrishtr, line 61: I guess this was dead code?

src/lab10.py
```diff
@@ -10,19 +10,19 @@
...
 from lab6 import INHERITED_PROPERTIES, style, cascade_priority
 from lab6 import DrawText, tree_to_list
-from lab7 import DrawLine, DrawOutline, DrawRect
-from lab8 import URL, Browser, Text, Element
+from lab7 import DrawLine, DrawOutline, DrawRect, Rect
+from lab8 import URL, Browser, Text, Element, Chrome
 from lab8 import BlockLayout, InputLayout, DEFAULT_STYLE_SHEET, INPUT_WIDTH_PX
-from lab8 import DocumentLayout, LineLayout, TextLayout
-from lab9 import EVENT_DISPATCH_JS, JSContext, Tab
-import wbetools
```
chrishtr, line 25: Why is this no longer needed?

infra/outlines.py
pavpanchekha: Yes—we used to include the `if __name__` block in the outline but later we refactored to have a `mainloop` function and I decided that listing it added little-to-no value.

src/lab10.py
pavpanchekha: Still needed, just moved up due to how the inliner works. (This is maybe a not-ideal way for the inliner to work but in this PR I just want the outline to work better.

PR #1215, opened 2024-05-21
Test full (inlined) modules against the normal tests

pavpanchekha:
This PR adds some pretty hacky Python code that does some import aliasing to run the "inlined" versions of each chapter's code against that chapter's tests.

All chapters now pass, the issues were all just bad imports that, luckily, don't affect the book at all.


infra/asttools.py
```diff
@@ -220,8 +220,8 @@ def parse(source, filename='<unknown>'):
 def inline(tree):
     tree2 = ResolveImports().visit(copy.deepcopy(tree))
     tree3 = ResolvePatches().double_visit(tree2)
-    tree4 = ResolveJSHide().visit(tree3)
-    return ast.fix_missing_locations(tree4)
+    #tree4 = ResolveJSHide().visit(tree3)
+    return ast.fix_missing_locations(tree3)
```
pavpanchekha, line 224: Responsibility for ignoring `js_hide` methods is moved to the caller, which does so in the modified `is_outline_hide_decorator` method

infra/runtests.py
```diff
@@ -49,6 +49,15 @@ def reload_module(mod):
...
     importlib.reload(mod)
 
+def import_as(file_path, module_name):
+    # Based in example in importlib documentation
+    import importlib.util
+
+    spec = importlib.util.spec_from_file_location(module_name, file_path)
+    module = importlib.util.module_from_spec(spec)
+    sys.modules[module_name] = module
+    spec.loader.exec_module(module)
```
pavpanchekha, line 59: This ugly hack pre-loads `labXXX.full.py` and makes it available as `labXXX`, so that when the tests run and actually import `labXXX` it gets the inlined version.

infra/runtests.py
```diff
@@ -92,6 +101,15 @@ def run_tests(chapter, file_name):
                 print(f"  {t.bold(value)}: Testing {chapter}...", end=" ")
                 results[value] = run_tests(chapter, value)
                 if not results[value][0]: print(t.green("pass"))
+            elif key == "full":
+                print(f"  {t.bold(value)}: Testing inlined {chapter}...", end=" ")
+                full_file = value.replace("-tests.md", ".full.py")
+                mod_name = value.replace("-tests.md", "")
+                os.system("make -C .. src/" + full_file)
```
pavpanchekha, line 108: This is kinda ugly, an alternative would be to add a make dependency

infra/runtests.py
```diff
@@ -49,6 +49,15 @@ def reload_module(mod):
             delattr(mod, attr)
     importlib.reload(mod)
 
+def import_as(file_path, module_name):
```
chrishtr, line 52:
```suggestion
def import_file_as(file_path, module_name):
```

Also add a comment explaining in English what this method does.

infra/runtests.py
chrishtr: How about we change all but lab10-tests and lab16-tests to just import the full inlined code version directly rather than the non-inlined one?

infra/asttools.py
pavpanchekha: You're right, the class isn't needed now

infra/runtests.py
pavpanchekha: Eh. Then we're still trusting the inliner in some way.

pavpanchekha: The ugliness here, to be clear, is that the tester calls Make to build the inlined version. Not that I think about it, it's probably possible to call the inliner through Python & avoid any impacts on the file system.

pavpanchekha: Ok, I messed around a bit with the internals of `import` in Python so now this ugliness is gone. `runtests.py` just uses `asttools` to produce an inlined AST and then imports that AST without having to write anything to disk.

chrishtr (approved):

PR #1216, opened 2024-05-22
Add references for four methods in the code not described in the book

### LINKED ISSUE #1210: Outline bugs

chrishtr:
Fixes #1210.

### DIFF
--- a/book/animations.md
+++ b/book/animations.md
@@ -1527,8 +1527,7 @@ animations that only require the draw step, like our example opacity
 animation, will now run super smoothly.
 
 One final note: the compositing data structures need to be cleared when changing
-tabs, so let's do that (factoring out a `clear_data` method to do it more
-cleanly now that there are several things to clear):
+tabs. Let's do that by factoring out a `clear_data` method that clears everything in one go.
 
 ``` {.python}
 class Browser:
--- a/book/animations.md
+++ b/book/animations.md
@@ -1540,10 +1540,8 @@ class Browser:
         self.composited_updates = {}
 
     def set_active_tab(self, tab):
-        self.active_tab = tab
+        # ...
         self.clear_data()
-        self.needs_animation_frame = True
-        self.animation_timer = None
 ```
 
 
--- a/book/animations.md
+++ b/book/animations.md
@@ -1526,7 +1526,7 @@ Now the draw display list will be based on the new display list, and
 animations that only require the draw step, like our example opacity
 animation, will now run super smoothly.
 
-Final note: the compositing data structures need to be cleared when changing
+One final note: the compositing data structures need to be cleared when changing
 tabs, so let's do that (factoring out a `clear_data` method to do it more
 cleanly now that there are several things to clear):
 
--- a/book/animations.md
+++ b/book/animations.md
@@ -1526,6 +1526,27 @@ Now the draw display list will be based on the new display list, and
 animations that only require the draw step, like our example opacity
 animation, will now run super smoothly.
 
+Final note: the compositing data structures need to be cleared when changing
+tabs, so let's do that (factoring out a `clear_data` method to do it more
+cleanly now that there are several things to clear):
+
+``` {.python}
+class Browser:
+    def clear_data(self):
+        self.active_tab_scroll = 0
+        self.active_tab_url = None
+        self.display_list = []
+        self.composited_layers = []
+        self.composited_updates = {}
+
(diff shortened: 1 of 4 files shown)


book/animations.md
```diff
@@ -1526,6 +1526,25 @@ Now the draw display list will be based on the new display list, and
 animations that only require the draw step, like our example opacity
 animation, will now run super smoothly.
 
+Final note: the compositing data structures need to be cleared when changing
```
pavpanchekha, line 1529:
```suggestion
One final note: the compositing data structures need to be cleared when changing
```

book/animations.md
```diff
@@ -1526,6 +1526,25 @@ Now the draw display list will be based on the new display list, and
...
+tabs, so let's do that (factoring out a `clear_data` method to do it more
+cleanly now that there are several things to clear):
```
pavpanchekha, line 1531:
```suggestion
tabs. Let's do that by factoring out a `clear_data` method that clears everything in one go.
```

book/scheduling.md
```diff
@@ -1360,14 +1360,47 @@ class Browser:
...
+        self.tasks.clear()
+        self.pending_scroll = None
+        self.condition.release()
+```
+
+We also need to split `new_tab` into a version that acquires a lock
+and one that doesn't (`new_tab_internal`):
+
+``` {.python}
     def new_tab(self, url):
```
pavpanchekha, line 1381:
```suggestion
class Browser:
    def new_tab(self, url):
```

book/scheduling.md
```diff
@@ -1360,14 +1360,47 @@ class Browser:
...
+    def new_tab_internal(self, url):
+        new_tab = Tab(self, HEIGHT - self.chrome.bottom)
+        self.tabs.append(new_tab)
+        self.set_active_tab(new_tab)
         self.schedule_load(url)
 ```
 
+...because cases like the `click` method on `Chrome` are called by
+`handle_click`, which has already acquired the lock.^[Using locks properly and
+avoiding race conditions and deadlocks can be quite difficult!]
```
pavpanchekha, line 1395:
```suggestion
This way `new_tab_internal` can be called directly by methods,
like `Chrome`'s `click` method, that already hold the lock.^[Using locks while
avoiding race conditions and deadlocks can be quite difficult!]
```

book/scheduling.md
```diff
@@ -1360,14 +1360,47 @@ class Browser:
...
+
 ``` {.python}
 class Chrome:
+    def click(self, x, y):
+        # ...
+            self.browser.new_tab_internal(
+                URL("https://browser.engineering/"))
```
pavpanchekha, line 1402:
```suggestion
        if self.newtab_rect.contains(x, y):
            self.browser.new_tab_internal(
                URL("https://browser.engineering/"))
```

I believe we never indent without including the line that indents.

src/lab12.py
```diff
@@ -408,8 +408,10 @@ def set_needs_quit(self):
         self.condition.release()
 
     def clear_pending_tasks(self):
+        self.condition.acquire(blocking=True)
         self.tasks.clear()
         self.pending_scroll = None
+        self.condition.release()
```
pavpanchekha, line 414: A bug fix?

book/scheduling.md
```diff
@@ -1383,13 +1416,15 @@ on the web page, we must schedule a task on the main thread:
 ``` {.python}
 class Browser:
     def handle_click(self, e):
+        self.lock.acquire(blocking=True)
```
pavpanchekha, line 1419: A bug in the chapter?

pavpanchekha (approved):
Thank you! I have language changes but otherwise good.

src/lab12.py
chrishtr: yes

book/scheduling.md
chrishtr: yes

PR #1217, opened 2024-05-22
Add an outline template and sort each outline according to the template

pavpanchekha:
This PR is pretty big but the idea is simple. Right now, the outline for each chapter is in "source order". This is OK for early chapters, but eventually that order becomes pretty random because the inliner does not preserve source order well. Fixing the source so that the inliner outputs the order we want is hard (I've tried).

This PR thus introduces a better way to put the outline in a coherent order: the outline template. This template can be found in `book/outline.txt`, and just lists all the classes, constants, and functions in the same syntax as output by the outliner. If you then call the outliner with `--template book/outline.txt`, it'll construct an outline for the chapter you want, and sort it to be in the same order as the template. (If the template is missing any methods, it'll error.) This way, all chapters output classes and functions in the same order, making it easier to compare for changes.

The exact format is documented a bit at the top of `book/outline.txt`, and the parser I wrote for this format is pretty hacky, but I did look over the outputs and they look good.


pavpanchekha:
Note that this PR depends on #1214, because that PR renames some constants and removes some items from the outline.

chrishtr:
Do you think this PR is necessary to meet the Friday deadline?

pavpanchekha:
I think so—I think the outlines count as copyedited parts of the book content, and they might affect page count / page breaking because we want to keep individual classes together when breaking pages in the outline. (Or if the typesetter wants to make the font smaller and do 2 or 3 columns.)

pavpanchekha:
I've spent a bunch of time looking at outlines recently and it really is difficult to compare them chapter-by-chapter without sorting them in the same order. (I did #1211 and #1214 by sorting alphabetically but a semantic order is clearly better. The outline roughly follows the browser rendering pipeline for the order.)

book/outline.txt
```diff
@@ -0,0 +1,465 @@
...
+def get_tabindex(node)
+
+# HTML parser
+def show(body)
+def lex(body)
+class HTMLParser:
+    SELF_CLOSING_TAGS
+    HEAD_TAGS
+    def __init__(body)
+    def parse()
```
chrishtr, line 43: How did you choose the order of the methods? Something better than alphabetical?

pavpanchekha: At the top level, sort of attempting to follow the browser rendering pipeline back to front, with similar methods & classes together. Within a class there were a lot more judgement calls but vaguely similar things should be together, and, say, layout should go before paint.

chrishtr (approved):

PR #1218, opened 2024-05-22
Fix tests which were failing for dumb reasons

pavpanchekha:
I'm not sure what happened (shouldn't CI have caught this) but I fixed a few code blocks to parallel the renamed constants in #1211

### DIFF
--- a/infra/runtests.py
+++ b/infra/runtests.py
@@ -25,7 +25,8 @@ IGNORE_FILES = [
     ".*",
     "lab*.full.py",
     "wbetools.py",
-    "browser.trace"
+    "browser.trace",
+    "outline*.txt",
 ]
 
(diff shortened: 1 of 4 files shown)


infra/runtests.py
```diff
@@ -25,7 +25,8 @@ def green(self, s): return s
     ".*",
     "lab*.full.py",
     "wbetools.py",
-    "browser.trace"
+    "browser.trace",
+    "outline*.txt",
```
chrishtr, line 29: Not needed for this PR

chrishtr (approved):

infra/runtests.py
pavpanchekha: It's helpful because I have the outlines generated on my computer & the tests fail if there are not-accounted-for files.

Issue #1219, opened 2024-05-22
Recheck unexpected outline changes


pavpanchekha:
I have to post some WIP as a bug report, sorry. I'll try to work on this more tonight.

Basically, I went through every outline for every chapter and looked at the diffs. Here are the things I found that were unexpected. Some just need to be checked (to make sure the change is described in the chapter) while others need to be fixed (hopefully in the code, not the book, but we need to make sure the book doesn't need changes).

Issues found:

- [x] URL `__str__` disappears in Chapter 8
- [x] Change `slant` argument to `get_font` into `style` in Chapter 11
- [x] Check that we describe `draw` -> `raster` in Chapter 11
- [x] Arguments of `JSContext.run` change in Chapter 12
- [x] Rename `raster_and_draw` to `composite_raster_and_draw` in Chapter 13
- [x] Why is `self_rect` added to `LineLayout` in Chapter 14?
- [x] `TextLayout` moves `word` argument to the end in Chapter 15
- [x] Add `parent` argument to `AccessibilityNode` in Chapter 15

I'm pretty sure this is exhaustive!

pavpanchekha:
I moved the CSS parsing change to another issue (#1221) and removed it from the list above, so that now #1220 fixes this issue and a separate PR can fix #1221.

pavpanchekha:
Confirmed fixed by #1220.

PR #1220, opened 2024-05-22
More outline fixes

pavpanchekha:
This PR fixes the subtler outline issues identified in #1219.

### DIFF
--- a/book/scripts.md
+++ b/book/scripts.md
@@ -167,7 +167,7 @@ Python into JavaScript. We'll be exporting a lot of functions, so to
 avoid polluting the `Tab` object with many new methods, let's put this
 code in a new `JSContext` class:
 
-``` {.python replace=__init__(self)/__init__(self%2c%20tab)}
+``` {.python replace=__init__(self)/__init__(self%2c%20tab),%2c%20code/%2c%20script%2c%20code}
 class JSContext:
     def __init__(self):
         self.interp = dukpy.JSInterpreter()
@@ -182,7 +182,7 @@ variable values and other state between them.
 
 We create this new `JSContext` object while loading the page:
 
-``` {.python replace=JSContext()/JSContext(self)}
+``` {.python replace=JSContext()/JSContext(self),run(body)/run(script%2c%20body)}
 class Tab:
     def load(self, url, payload=None):
         # ...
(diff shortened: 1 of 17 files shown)


src/lab15.py
```diff
@@ -283,7 +283,7 @@ def add_inline_child(self, node, w, child_class, frame, word=None):
         line = self.children[-1]
         previous_word = line.children[-1] if line.children else None
         if word:
-            child = child_class(node, line, previous_word, word)
+            child = child_class(node, word, line, previous_word)
```
chrishtr, line 286: Was this a real bug in the code?

pavpanchekha: Ah, no, ok, here's what's going on. Between Ch 14 and 15, we change where the `word` argument goes in the argument order. In 14 and prior, the `word` argument goes after `node`; in 15 and 16, it goes at the end. This came about due to various refactors, but isn't necessary now, so I'm fixing it so the argument order doesn't change for no reason.

book/scripts.md
```diff
@@ -309,14 +309,12 @@ ignore the crash. Web pages shouldn't be able to crash our browser!
 You can implement that like this:
 
 ``` {.python}
-class Tab:
```
chrishtr, line 312: yikes

src/lab9.py
```diff
@@ -45,8 +45,11 @@ def __init__(self, tab):
         self.node_to_handle = {}
         self.handle_to_node = {}
 
-    def run(self, code):
-        return self.interp.evaljs(code)
+    def run(self, script, code):
```
chrishtr, line 48: why is this change necessary?

book/scripts.md
pavpanchekha: Yeah, this was kinda ugly. Basically, in Chapter 9 we put the error-handling code in `Tab.load`; in Chapter 12 we want to move it to `run` because then we can use `Task(self.js.run, ...)`. Better put it in its final location from the beginning.

src/lab9.py
pavpanchekha: See above comment.

book/scripts.md
chrishtr: Is this an orthogonal fix you found along the way? Or does it benefit the outline in a concrete way?

pavpanchekha: It benefits the outline because moving the error handling into `JSContext.run` requires adding a `script` parameter, meaning that the outline changes (a new parameter appears).

PR #1223, opened 2024-05-23
Fix `CSSParser` differences between Chapters 13 & 14

### LINKED ISSUE #1221: Unify CSS parsing changes in Chapters 13 & 14
From chapter 6 through Chapter 12, a CSS property–value pair is `<word>: <word>`. However, in Chapter 13 we want to parse `transition` values, which have spaces in them (unlike `word`) and in Chapter 14 we want to parse `outline` values, which have the same issue. So at some point `CSSParser` has to change.

The issue is that the actual change is described in Chapter 14 and uses the `until_chars` method. But in Chapter 13, which comes before it, there's a silent code change to instead use an `until_semicolon` function.

I don't care which approach is which, but we should 1) use the same approach in both places; and 2) describe the change in Chapter 13, not Chapter 14!

pavpanchekha:
This PR is pretty nasty, sorry. It includes a bunch of linked changes, ultimately to fix #1221:

- Change Chapter 13 to parse CSS values with `until_chars`, like in Chapter 14, instead of `until_semicolon`
- Move the corresponding text from Chapter 14 to Chapter 13, with what changes are necessary to preserve continuity
- No longer change `word` in Chapter 13 or 14, because those changes are unnecessary (those changes were there to accommodate complex CSS values in `word`, but we now parse CSS values with `until_chars`, so `word` can stay unchanged)
- Eliminate various other random changes between Chapters 6, 13, and 14 by deduplicating `CSSParser` methods with `wbetools.patch`.

### DIFF
--- a/src/lab13.py
+++ b/src/lab13.py
@@ -306,22 +306,6 @@ def parse_transform(transform_str):
 
 @wbetools.patch(CSSParser)
 class CSSParser:
-    def word(self):
-        start = self.i
-        in_quote = False
-        while self.i < len(self.s):
-            cur = self.s[self.i]
-            if cur == "'":
-                in_quote = not in_quote
-            elif cur.isalnum() or cur in ",/#-.%()\"'" \
-                or (in_quote and cur == ':'):
-                self.i += 1
-            else:
-                break
-        if not (self.i > start):
-            raise Exception("Parsing error")
-        return self.s[start:self.i]
-
     def until_chars(self, chars):
         start = self.i
         while self.i < len(self.s) and self.s[self.i] not in chars:
--- a/src/lab13.py
+++ b/src/lab13.py
@@ -306,19 +306,6 @@ def parse_transform(transform_str):
 
 @wbetools.patch(CSSParser)
 class CSSParser:
-    def __init__(self, s):
-        self.s = s
-        self.i = 0
-
-    def whitespace(self):
-        while self.i < len(self.s) and self.s[self.i].isspace():
-            self.i += 1
-
-    def literal(self, literal):
-        if not (self.i < len(self.s) and self.s[self.i] == literal):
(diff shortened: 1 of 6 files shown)


pavpanchekha:
I believe this is the last outline-driven change.

src/lab13.py
```diff
@@ -306,64 +306,25 @@ def parse_transform(transform_str):
...
-    def whitespace(self):
-        while self.i < len(self.s) and self.s[self.i].isspace():
-            self.i += 1
-
-    def literal(self, literal):
-        if not (self.i < len(self.s) and self.s[self.i] == literal):
-            raise Exception("Parsing error")
-        self.i += 1
-
-    def word(self):
```
chrishtr, line 322: I think it's dangerous to remove this version of word right now. Maybe it was added
to make the parser not crash on real pages, I'm not sure. Since it isn't in the book I don't
think it is urgent to remove it.

pavpanchekha: Ok, that's fair, I can put the `word` function back in. I *am* pretty sure it isn't necessary (we only call `word` for property *names* and media-query property names) but having it in won't affect the outline so doesn't matter much for the book.

src/lab14.py
```diff
@@ -644,78 +644,8 @@ def matches(self, node):
...
-
-    def word(self):
-        start = self.i
-        in_quote = False
-        while self.i < len(self.s):
-            cur = self.s[self.i]
-            if cur == "'":
-                in_quote = not in_quote
-            if cur == "(":
-                in_parens = True
```
pavpanchekha, line 669: These two lines were the only difference from `lab13.py`'s `word` and note that the `in_parens` variable is not used anywhere else, so I think this method is identical to `lab13` and doesn't need to be restored.

PR #1242, opened 2024-05-25
Fold in copyediting fixes from the hybridization effort

pavpanchekha:


pavpanchekha:
Not urgent, these are all already in the final book files.

chrishtr (approved):

PR #1247, opened 2024-05-27
fix savelayer in example

chrishtr:


pavpanchekha (approved):

pavpanchekha:
Do we need to fix the draft?

chrishtr:
> Do we need to fix the draft?

Ideally yes. I'll do that now.

PR #1250, opened 2024-05-30
Make Chapter 6 match Chapter 5's `flush()` method

### LINKED ISSUE #1249: Inconsistency between Ch 5 and 6
I was looking for more code to deduplicate across the labs and found this issue. [In Ch 5](https://browser.engineering/layout.html#cb14), we have this code:

```
class BlockLayout:
    def flush(self):
        # ...
        for rel_x, word, font in self.line:
            x = self.x + rel_x
            y = self.y + baseline - font.metrics("ascent")
            self.display_list.append((x, y, word, font))
        # ...
```

Note `self.line` contains relative `x` positions, and the `self.display_list` contains absolute positions.

However, [in Ch 6](https://browser.engineering/styles.html#cb42) we do this:

```
def flush(self):
    # ...
    metrics = [font.metrics() for x, word, font, color in self.line]
    # ...
    for x, word, font, color in self.line:
        # ...
        self.display_list.append((x, y, word, font, color))
    # ...
```

Note that now the `display_list` stores relative `x` positions, which are converted to absolute positions here:

```
def paint(self):
    # ...
    for x, y, word, font, color in self.display_list:
        cmds.append(DrawText(self.x + x, self.y + y,
                                     word, font, color))
```

So there's an inconsistency, and it's not remarked on in the text and shouldn't be there. We should fix this ASAP.

pavpanchekha:
Fixes #1249 by updating Chapter 6 to match Chapter 5. Unfortunately, this requires changes to the book, specifically to two code blocks:

- In the `flush` code block, the only change is renaming a variable from `x` to `rel_x`; I think this is fine for the typesetter.
- In the `paint` code block, the code changes more dramatically but line length goes down and line count doesn't change, so we're maybe OK.

### DIFF
--- a/book/styles.md
+++ b/book/styles.md
@@ -909,7 +909,7 @@ def flush(self):
     # ...
     metrics = [font.metrics() for x, word, font, color in self.line]
     # ...
-    for x, word, font, color in self.line:
+    for rel_x, word, font, color in self.line:
         # ...
         self.display_list.append((x, y, word, font, color))
     # ...
@@ -921,8 +921,7 @@ That `display_list` is converted to drawing commands in `paint`:
 def paint(self):
     # ...
     for x, y, word, font, color in self.display_list:
-        cmds.append(DrawText(self.x + x, self.y + y,
-                                     word, font, color))
+        cmds.append(DrawText(x, y, word, font, color))
 ```
 
 `DrawText` now needs a `color` argument, and needs to pass it to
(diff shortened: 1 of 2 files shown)


pavpanchekha:
Ok—I found a way to even out the line count while also making the code _more_ correct.

chrishtr (approved):
Do you have an example web page where it draws incorrectly? Add a test if so?

pavpanchekha:
It drew correctly but it used different code in Chapter 5 & 6. (An incomplete update in a prior PR.)

PR #1252, opened 2024-05-30
Add instructions on how to avoid the shader errors in Skia 124

pavpanchekha:
This is on the porting page, so not an issue. Found by @HinTak

### DIFF
--- a/book/porting.md
+++ b/book/porting.md
@@ -40,10 +40,23 @@ class DrawImage:
 ```
 
 The new API is somewhat cleaner and more customizable.
+
+Finally, add these `SDL_GL_SetAttribute` calls to set GL version Skia
+expects:
+
+```
+class Browser:
+    def __init__(self):
+        self.sdl_window = sdl2.SDL_CreateWindow(b'Browser',
+            sdl2.SDL_WINDOWPOS_CENTERED,
+            sdl2.SDL_WINDOWPOS_CENTERED,
+            WIDTH, HEIGHT,
+            sdl2.SDL_WINDOW_SHOWN | sdl2.SDL_WINDOW_OPENGL)
+
+        sdl2.SDL_GL_SetAttribute(sdl2.SDL_GL_CONTEXT_MAJOR_VERSION, 3)
+        sdl2.SDL_GL_SetAttribute(sdl2.SDL_GL_CONTEXT_MINOR_VERSION, 2)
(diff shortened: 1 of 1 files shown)


book/porting.md
```diff
@@ -40,10 +40,23 @@ class DrawImage:
...
+            sdl2.SDL_WINDOW_SHOWN | sdl2.SDL_WINDOW_OPENGL)
+
+        sdl2.SDL_GL_SetAttribute(sdl2.SDL_GL_CONTEXT_MAJOR_VERSION, 3)
+        sdl2.SDL_GL_SetAttribute(sdl2.SDL_GL_CONTEXT_MINOR_VERSION, 2)
+        sdl2.SDL_GL_SetAttribute(sdl2.SDL_GL_CONTEXT_FORWARD_COMPATIBLE_FLAG, True)
+        sdl2.SDL_GL_SetAttribute(sdl2.SDL_GL_CONTEXT_PROFILE_MASK,
+                                 sdl2.SDL_GL_CONTEXT_PROFILE_CORE)
+```
  
-Additionally, as of our latest testing, this release has difficulties
```
chrishtr, line 44: Does your change here fix the issues mentioned in this paragraph?

pavpanchekha: Yes

HinTak:
@pavpanchekha hmm, checking my pysdl2 code here, I think the GL set attribute calls should be made after `SDL_Init`, but __before__ `SDL_CreateWindow`?

HinTak:
Interesting enough, I just checked Google's own code (they have a skia+sdl2 example in c which was tidied away in m93; you need to check out m92 to see it), they made the set attribute calls even before `SDL_Init` (unlikely for them to try/test on mac though). Anyway, I am pretty sure after window creation is too late.

pavpanchekha:
I did some experimenting here, and at least in our case, to make it work we need to call it _after_ `SDL_CreateWindow` and _before_ `SDL_GL_CreateContext`. Doing it earlier or later doesn't work.

HinTak:
@pavpanchekha okay. I'd think it needs to happen before the first drawing operation. I have been thinking maybe I'll put the examples out.

pavpanchekha:
It seems to have to happen before the GL context it created, but after the window. All the drawing operations of course come after the GL context.

HinTak:
Hmm,  I have got the issue 214 sample code under https://github.com/HinTak/skia-python-examples/ and it passes under github CI (except the glfw part), with the set attributes between SDL_Init() and SDL_CreateWindow(). 

The Google Skia+SDL c++ sample is at https://github.com/HinTak/skia-c-examples/ .

PR #1255, opened 2024-06-06
Restyle headers a bit

pavpanchekha:
This PR makes three big changes to styling. In each pair of screenshots below, I show "before" first, than "after".

# First: move the chapter numbers on the main page so they're within the page bounds. This makes them visible on mobile.

Before:
image

After:
Screenshot 2024-06-06 at 3 50 56 PM

# Second: Use CSS counters to number sections, instead of Pandoc. This allows us to style the section number:

Before:
Screenshot 2024-06-06 at 3 51 51 PM

After:
image

Also we don't need to put the counter into the table of contents:

Before:
image

After:
Screenshot 2024-06-06 at 3 52 54 PM

# Third: Use a pseudoelement "#" instead of an underline to suggest clicking a header for a permalink. This has bothered me for a while, the underlines look bad.

Before:
image

After:
Screenshot 2024-06-06 at 3 54 00 PM

### DIFF
--- a/www/book.css
+++ b/www/book.css
@@ -32,11 +32,13 @@ header, h1, strong {
 }
 h1 { font-size: 105%; margin: 1.5em 0 0; }
 
+/* Auto-number section headings, but not on the main page. */
+/* Also, on hover over section headings, show a "#" to indicate permalinks. */
+body > h1 { counter-increment: section; }
+body > h1::before { content: counter(section) ". "; color: #888; }
 .main h1::before { content: none !important; }
 h1:not(.title):hover { cursor: pointer; }
 h1:not(.title):hover::after { content: " #"; color: #888; }
-body > h1 { counter-increment: section; }
-body > h1::before { content: counter(section) ". "; color: #888; }
 
 header { text-align: center; margin: 3em 0 1em; }
 header h1 { font-size: 200%; font-weight: bold; }
@@ -54,6 +56,8 @@ a:active { color: #f64; }
 
 li { margin: 1em 0; }
 
+/* On the main page, <ol> contains chapter names and <hi> contains
+ * Part names; all should be big. */
 .main ol, .main h1 { font-size: 120%; }
 .main header { margin-bottom: 3em; }
 .main > h1 { margin: 0; }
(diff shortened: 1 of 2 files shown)


www/book.css
```diff
@@ -48,10 +54,10 @@ a:active { color: #f64; }
 
 li { margin: 1em 0; }
 
-.main ol, .main > h1 { font-size: 120%; }
+.main ol, .main h1 { font-size: 120%; }
```
chrishtr, line 57: Why is this change necessary?

www/book.css
```diff
@@ -31,9 +31,15 @@ header, h1, strong {
 }
 }
 h1 { font-size: 105%; margin: 1.5em 0 0; }
+
+.main h1::before { content: none !important; }
```
chrishtr, line 35: What is this rule for?

www/book.css
pavpanchekha: Ah—sort of bizarrely the "Part XXX" headers were bigger than the "Introduction" and "Conclusion" headers because the latter are inside a <section>.

www/book.css
pavpanchekha: On the main page (`index.html`), we don't want to auto-number the headings.

chrishtr: Please add a comment saying that then.

www/book.css
chrishtr: Can you add a comment explaining?

www/book.css
```diff
@@ -31,9 +31,17 @@ header, h1, strong {
 }
 }
 h1 { font-size: 105%; margin: 1.5em 0 0; }
+
+/* Auto-number section headings, but not on the main page. */
```
chrishtr, line 35: Idea: how about we include the section numbers on the homepage and implement it with counter-reset?

pavpanchekha: I'm a bit confused. When you say "section numbers", do you mean:

- Auto-number the parts?
- Auto-number the chapters?
- Auto-number the sections in each chapter?

chrishtr: I  mean sections within chapters

pavpanchekha: Right now the homepage doesn't show the sections. You want to add them? I guess that would look more like the printed book, not a bad idea. But perhaps a separate PR?

chrishtr: ok

PR #1256, opened 2024-06-06
Fix JavaScript version of `Font.metrics`

pavpanchekha:
When we first wrote this code, different browsers implemented different Canvas TextMetrics fields, so we had some browser-specific code. Moreover, different browsers have different values for the same fields. That's all still the case, but it *has* gotten a bit better: all browsers implement `fontAscentBoundingBox` and give it the same value. So this PR switches to using that for font metrics. This makes the UI look a little uglier in Chrome (because the spacing of that UI depends on font metrics, and the use of `fontXXXBoundingBox` is a little larger meaning more whitespace than what we had before) but on the other hand it totally fixes the `lab3-baselines` widget, which I think makes this a clear win.

Before:
image

Note that the text is not actually on the baseline! It's not actually on any baseline.

After:
image

The text is now lined up nicely on the baseline.

### DIFF
--- a/www/widgets/rt.js
+++ b/www/widgets/rt.js
@@ -363,9 +363,6 @@ class tkinter {
                     ctx.font = this.string;
                     let m = ctx.measureText("Hxy");
 
-                    // Only Safari provides emHeight properties as of 2021-04
-                    // We fake them in the other browsers by guessing that emHeight = font.size
-                    // This is not quite right but is close enough for many fonts...
                     let asc = ctx.measureText("Hxy").fontBoundingBoxAscent / rt_constants.ZOOM;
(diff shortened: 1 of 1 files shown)


www/widgets/rt.js
```diff
@@ -362,21 +362,12 @@ class tkinter {
                     ctx.textBaseline = "alphabetic";
                     ctx.font = this.string;
                     let m = ctx.measureText("Hxy");
-                    let asc, desc;
 
                     // Only Safari provides emHeight properties as of 2021-04
```
chrishtr, line 366: Remove this comment as obsolete?

chrishtr (approved):

PR #1259, opened 2024-06-07
Update `lab16.py` to sync with `lab15.py` bug fixes

pavpanchekha:
This PR makes a few cleanups to `lab16.py` by synchronizing it with `lab15.py`:

- Remove some unneeded patch methods (that are identical to lab15)
- Load tabs asynchronously
- Avoid running JavaScript on discarded tabs
- Fix self_rect for `DocumentLayout`s

### DIFF
--- a/src/lab16.py
+++ b/src/lab16.py
@@ -1380,6 +1380,24 @@ class Tab:
 
         self.browser.measure.stop('render')
 
+@wbetools.patch(AccessibilityNode)
+class AccessibilityNode:
+    def compute_bounds(self):
+        if self.node.layout_object:
+            return [absolute_bounds_for_obj(self.node.layout_object)]
+        if isinstance(self.node, Text):
+            return []
+        inline = self.node.parent
+        bounds = []
+        while not inline.layout_object: inline = inline.parent
+        for line in inline.layout_object.children.get():
+            line_bounds = skia.Rect.MakeEmpty()
+            for child in line.children:
+                if child.node.parent == self.node:
+                    line_bounds.join(skia.Rect.MakeXYWH(
+                        child.x.get(), child.y.get(), child.width.get(), child.height.get()))
+            bounds.append(line_bounds)
+        return bounds
 
 if __name__ == "__main__":
     wbetools.parse_flags()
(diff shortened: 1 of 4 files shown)


src/lab16.py
```diff
@@ -319,14 +319,11 @@ def layout(self, width, zoom):
...
 
-    def paint(self):
-        return []
-
     def paint_effects(self, cmds):
         if self.frame != self.frame.tab.root_frame and self.frame.scroll != 0:
             rect = skia.Rect.MakeLTRB(
-                self.x, self.y,
-                self.x + self.width, self.y + self.height)
+                self.x.get(), self.y.get(),
```
chrishtr, line 325: Why the changes to add `.get()`?

pavpanchekha: Yeah this is a good question. I am fairly confident the previous code was wrong (`DocumentLayout` no longer has a float `self.x` field) though I don't know how to trigger a crash in the old code (logically, scrolling inside a frame, but that doesn't seem to work?)

chrishtr: Did you try a page with an iframe? That's what this code is for.

pavpanchekha: Ok—I tested more and learned:
- Yes, this bug fix is necessary; scrolling in iframes in lab16 is broken on `main`
- There are other issues that have to be fixed, including a minor (no change in line count) fix to the book
- There's still a bug, wherein scrolling a frame doesn't seem to cause paint.

pavpanchekha:
So please re-review, there have been a number of changes, largely necessary and caused by more testing (and then debugging) with iframes.

There's also still a bug with scrolling iframes. I'm not sure what causes it, but here's how to reproduce:

1. Make a page with an `<iframe>`
2. Make the embedded document longer than the iframe allows
3. Click inside iframe, press down to scroll
4. No scroll occurs
5. Click somewhere at random
6. Scroll occurs now

I've checked that a commit is made with an empty list of composited updates, which I imagine is the problem. We need to mark at least the root document a composited update.

pavpanchekha:
Scrolling issue is tough, fixed in #1264 but requires book changes that the publisher will be unhappy about. But I think this PR is safe to merge and requires very minimal book changes. Maybe I should update book-edited.tex in this PR too?

chrishtr:
I tested locally and there were some errors matching the chapter text to this PR

pavpanchekha:
Forgot to push, sorry!

PR #1260, opened 2024-06-07
Add multiple-choice quizzes

ashton314:
This PR adds multiple-choice quizzes to the web version of the book.

Quizzes look like this:

```
::: {.mc-quiz #unique-quiz-name}
This is where the question text goes. Only one paragraph right now.

 - First item must be the answer.
 - Other items will be mixed in.
 - Internally they're called "distractors".
 - The quiz widget library we're using handles the shuffling.

--------------

You can have multiple questions separated with a hr line.

 - What?
 - I don't understand.
 - Where's the tea?
:::
```

The `.mc-quiz` class is important; that's how the Lua filter knows to pass this div off to the quiz parsing handler.

# Implementation details

This required a 3rd-party JSON library; I just snarfed the Lua code and dumped it straight in the `infra/` tree.

Additionally, I needed to add the `quiz-embed.iife.js` file; this is what looks for elements with the `data-quiz-questions` attribute and turns them into interactive quizzes. I found this checked into the source of the `mdbook-quiz` that we're modeling these quizzes after.


pavpanchekha:
@chrishtr this is something I've been collaborating with another prof on. Happy to discuss on Friday.

ashton314:
Screenshots of what it looks like:
image
image (1)

book/layout.md
```diff
@@ -142,6 +142,34 @@ tree][gecko-tree].
 [webkit-tree]: https://webkit.org/blog/114/webcore-rendering-i-the-basics/
 [gecko-tree]: https://wiki.mozilla.org/Gecko:Key_Gecko_Structures_And_Invariants
 
+::: {.mc-quiz #layout-tree-quiz}
```
chrishtr, line 145:
```suggestion
::: {.web-only .mc-quiz #layout-tree-quiz}
```

book/layout.md
```diff
@@ -142,6 +142,34 @@ tree][gecko-tree].
...
+ - Only at a layout object with no children.
+ - It never bottoms out.
+ - Only at an `InlineLayout` object.
+ - Only at a `DocumentLayout` object.
+
+-----------
+
+What type of object is at the root of the layout tree? The intermediate nodes? The leaves?
+
+ - The root is a DocumentLayout, all other nodes are BlockLayout
```
chrishtr, line 166:
```suggestion
 - The root is a DocumentLayout, all other nodes are BlockLayout.
```

book/layout.md
```diff
@@ -142,6 +142,34 @@ tree][gecko-tree].
...
+ - The root is an InlineLayout, the intermediate nodes are DocumentLayout, the leaves are BlockLayout
```
chrishtr, line 167:
```suggestion
 - The root is an InlineLayout, the intermediate nodes are DocumentLayout, the leaves are BlockLayout.
```

book/layout.md
```diff
@@ -142,6 +142,34 @@ tree][gecko-tree].
...
+ - The root and intermediate nodes are BlockLayout, the leaves are InlineLayout
```
chrishtr, line 168:
```suggestion
 - The root and intermediate nodes are BlockLayout, the leaves are InlineLayout.
```

book/layout.md
```diff
@@ -142,6 +142,34 @@ tree][gecko-tree].
...
+ - The root is a DocumentLayout, the intermediate nodes are BlockLayout, the leaves are InlineLayout or BlockLayout
```
chrishtr, line 169:
```suggestion
 - The root is a DocumentLayout, the intermediate nodes are BlockLayout, the leaves are InlineLayout or BlockLayout.
```

infra/filter.lua
```diff
@@ -1,5 +1,7 @@
 -- Pass 1: Load configuration data
 
+local json = require('infra/dkjson')
```
chrishtr, line 3: There is already a JSON file (config.json) which is somehow parsed by pandoc and passed to this filter. Can we extend it instead of importing a second JSON parser? Another option is to pass additional yaml files on the commandline (see section 8.10.2 [here](https://pandoc.org/chunkedhtml-demo/8.10-metadata-blocks.html)).

infra/template.html
```diff
@@ -17,6 +17,7 @@
 $for(css)$  <link rel="stylesheet" href="$base$$css$" />$endfor$
   <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Vollkorn%7CLora&display=swap" />
   <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Vollkorn:400i%7CLora:400i&display=swap" />
+  <link rel="stylesheet" href="$base$quiz_style.css" />
```
chrishtr, line 20: Please put these imports behind a default-off flag configured by config.json.

book/layout.md
chrishtr: Also: Please put these imports behind a default-off flag configured by config.json.

chrishtr (commented):
Also, it looks like you quiz-embed.iife.js inlines React plus your actual quiz code. I guess I don't object to including React, but can you separate the two into different JS files?

book/layout.md
ashton314: I'll put the imports behind a default-off flag. How about I make the `.mc-quiz` class act like `.web-only` by default, since the quizzes won't make sense anywhere else?

infra/filter.lua
ashton314: Thanks you your comment, I discovered that `pandoc.json.encode` exists, so I was able to use that to encode the JSON representation of the quiz. It's not called explicitly anywhere in the project, (pandoc calls it implicitly during the build process) which might explain why I didn't find it the first time. 😅 Using the JSON encoder provided by pandoc is a much better solution; thank you for pushing back on this!

ashton314:
> quiz-embed.iife.js as two separate files

Can I ask why? I didn't build this file; I just snarfed it from [mdbook-quiz](https://github.com/cognitive-engineering-lab/mdbook-quiz)'s source. I could figure out how to build React and the quiz driver separately, but that will take me a bit of time, and I want to know if it'd be worth the effort.

book/layout.md
ashton314: I've added a `show_quiz` flag; it's on only in the `draft` mode. Assets and quizzes only get emitted when in `draft` mode.

ashton314:
@pavpanchekha @chrishtr Ready for another review. Thanks for feedback and comments Chris!

ashton314:
(Just rebased)

chrishtr:
> > quiz-embed.iife.js as two separate files
> 
> Can I ask why? I didn't build this file; I just snarfed it from [mdbook-quiz](https://github.com/cognitive-engineering-lab/mdbook-quiz)'s source. I could figure out how to build React and the quiz driver separately, but that will take me a bit of time, and I want to know if it'd be worth the effort.

My concerns are:

* How do we maintain this code if there are bugs found?
* Does it involve re-importing from that github project?
* Are there instructions on how to do that?
*  Is that project trustworthy? (Can we run this code in sandboxed iframe to keep it separate from the book if we're not sure?)

ashton314:
Good concerns.

> How do we maintain this code if there are bugs found?

The source is here: https://github.com/cognitive-engineering-lab/mdbook-quiz/tree/main/js

I was able to clone this repository and build the project by following the instructions included. (`depot build` took care of everything; built js bundle was identical (shasum verified) to the one I copied into this project.)

If there are bugs, ideally we would submit them upstream. Another option would be to fork the repository and move under our tree. That seems like a lot of work. (I'm happy to do it, but that would take me away from other things.)

I don't know of a way to include this JS bundle; the project was originally made to work with Rust's [mdBook](https://github.com/rust-lang/mdBook) system.

> Instructions

Yes! The instructions are here: https://github.com/cognitive-engineering-lab/mdbook-quiz/blob/main/CONTRIBUTING.md

Short and to-the-point. Worked flawlessly for me, which I found a little astounding given the typical state of academic software. 😂 

> Is that project trustworthy?

I certainly think so. @bennn knows Will Chrichton and suggested using this library; I think @pavpanchekha is fine with it, but we can affirm/correct me.

--------------------

Short of it: I wasn't able to split this file into two blobs. (Possible probably, but I didn't have time to go spelunking through different JS build systems.) I consider the project to be trustworthy enough to just include that static JS blob. It would not be too much effort to fork the project if needed and fix bugs ourselves. @pavpanchekha thoughts?

chrishtr:
Thanks for the context and explanation.  I agree you shouldn't go to the trouble of splitting the React file out given what you said.

Can you add a README file explaining the process to update the JS to a new version?

ashton314:
Done! Let me know if you're satisfied with the instructions. I tested them on my machine, so I'm pretty confident that they'll work. That said, that's me saying It Works on My Machine™ :)

ashton314:
@pavpanchekha quizzes should look at least a *little* better now, I hope.

README.md
```diff
@@ -70,3 +70,18 @@ different chapters, named `server8.py`, `server9.py`, and so on.
 You can also run the book's unit tests with:
 
     make test
+
+## Rebuilding the quiz JavaScript
+
+The source for the interactive multiple-choice quizzes is taken from [mdbook-quiz](https://github.com/cognitive-engineering-lab/mdbook-quiz). To rebuild the JS blob that we include on the quiz pages (located in `www/quiz-embed.iife.js`) do the following:
+
+ - Install [depot](https://github.com/cognitive-engineering-lab/depot) (made by the same people who made `mdbook-quiz`; tldr: `cargo install depot-js --locked`).
```
chrishtr, line 78: How do you install cargo?

chrishtr (approved):

chrishtr:
@pavpanchekha  are you also good with it?

pavpanchekha:
I think we discovered there was no light theme, so I think Ashton was going to do that and then we merge. I'd like to get the follow-up PR (#1275) merged soon after so we could test and then maybe release on the production site.

ashton314:
# Light mode
image
image


# Dark mode
image
image

Really simple styling per Pavel's instructions. I'm happy to tweak anything else, but I think this looks less awful than before, no?

ashton314:
(Pictures now out-of-date; removed `border-radius` to match `div.further`.

PR #1262, opened 2024-06-10
Update outlines to the latest

chrishtr:


chrishtr:
I cut-and-pasted from our latest latex to theirs. Please check the outlines section of PDF to verify, I'll do the same.

chrishtr:
I reviewed all of them and they seem correct? Would be great if you can run your diff on them if that is feasible.

pavpanchekha (approved):

pavpanchekha:
The diff is a mess but I looked over them by hand and they look right.

Issue #1257, opened 2024-06-06
Widgets / JS Browser crashes in Safari


pavpanchekha:
I get bizarre and varying and hard-to-reproduce (how??) crashes for the JS Browser in Safari.

For example, for lab10, I get `TypeError: undefined is not an object (evaluating 'this.line.push')`

For lab6, I get: `TypeError: undefined is not an object (evaluating 'node.style["font-weight"]')`

For lab14, I get: `selector.matches is not a function. (In 'selector.matches(node)', 'selector.matches' is undefined)`

The widgets seem to work fine in Firefox and Chrome.

pavpanchekha:
Ok, I've looked into this and it looks like Safari does module import resolution wrong somehow? I'm not 100% sure, but it looks like it initializes modules in the wrong order somehow, and as a result the wrong patches get applied? Not 100% sure.

pavpanchekha:
Frustratingly, this behavior doesn't reproduce in the debugger or if you do other stuff to reorder execution so it has proven hard to debug. I suspect that file download order affects execution order but it's hard to be sure.

PR #1265, opened 2024-06-10
Remove duplicate imports, this helps Safari somehow

pavpanchekha:
This PR helps #1257 somehow, though I'm not sure how. But these duplicate imports seem to mess up, for some reason, the order in which patches are applied, and that results in broken widgets.


chrishtr (approved):
I sent a note to the Chrome team member who standardized JS modules in the first place, to see what's up with this bug.

But this CL looks good, thanks for finding a workaround!

PR #1266, opened 2024-06-10
Fix block layout on `lab6.py`

pavpanchekha:
This PR, which does not affect the book, fixes a bug where lab6.py forgot to add the block y position when drawing text.

### DIFF
--- a/src/lab6.py
+++ b/src/lab6.py
@@ -259,11 +259,6 @@ class BlockLayout:
     @wbetools.delete
     def close_tag(self, tag): pass
 
-    @wbetools.js_hide
-    def __repr__(self):
-        return "BlockLayout[{}](x={}, y={}, width={}, height={})".format(
-            self.layout_mode(), self.x, self.y, self.width, self.height)
-
 @wbetools.patch(DrawText)
 class DrawText:
     def __init__(self, x1, y1, text, font, color):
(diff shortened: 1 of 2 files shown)


src/lab6.py
```diff
@@ -231,7 +231,7 @@ def flush(self):
         baseline = self.cursor_y + 1.25 * max_ascent
         for rel_x, word, font, color in self.line:
             x = self.x + rel_x
-            y = baseline - font.metrics("ascent")
+            y = self.y + baseline - font.metrics("ascent")
```
chrishtr, line 234: Can you add a unittest? Also, does this draw the browser.engineering homepage wrong? I would think everything looks terrible without this bugfix.

pavpanchekha: Yeah it looks like crap. I found this because I was debugging Safari and thought it was a Safari bug. then I checked Chrome, thought it was a compiler bug, eventually thought to try the Python. I can add a test.

pavpanchekha: Added a test. Along the way, I removed two `__repr__` methods in lab6 that were desynced from lab5 and which made writing the test hard.

pavpanchekha: (The test is just a copy-and-paste of a lab5 test)

src/lab6.py
```diff
@@ -259,11 +259,6 @@ def open_tag(self, tag): pass
     @wbetools.delete
     def close_tag(self, tag): pass
 
-    @wbetools.js_hide
```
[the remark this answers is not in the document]
pavpanchekha: The string representation used in lab6 is different from lab5, which made it harder to copy tests back and forth. Moreover, the string representation for `DrawText` calls leaves off the position information, which made testing position impossible! It'd be possible (though annoying) to keep the `BlockLayout` repr, but I think it'd be better if our string representations didn't change chapter to chapter, so I think of this as a small cleanup that helps this PR.

PR #1267, opened 2024-06-10
Deduplicate lab 13

pavpanchekha:
This PR uses `wbetools.patch` to remove duplicate definitions in `lab13.py`. This doesn't affect the book. There are still a non-trivial number of duplicate methods because this chapter adds superclasses to the `DrawXXX` methods. We can't patch in a superclass (it's possible in Python, but I don't think it's possible in JavaScript) so we have to copy and paste some code. But a lot less than is in there now!

The benefit of deduplicating is that code won't get desynchronized. For example, I noticed while doing this that our changes to `previous_word` in Chapter 7 hadn't made it into Chapter 13.

### DIFF
--- a/src/lab13.py
+++ b/src/lab13.py
@@ -30,6 +30,7 @@ from lab11 import FONTS, get_font, parse_color, NAMED_COLORS, parse_blend_mode, 
 from lab11 import paint_tree
 from lab12 import MeasureTime, SingleThreadedTaskRunner, TaskRunner
 from lab12 import Tab, Browser, Task, REFRESH_RATE_SEC, Chrome, JSContext
+from lab12 import CommitData
 
 @wbetools.patch(Text)
 class Text:
@@ -1047,8 +1048,6 @@ class Tab:
         clamped_scroll = self.clamp_scroll(self.scroll)
         if clamped_scroll != self.scroll:
             self.scroll_changed_in_tab = True
-        if clamped_scroll != self.scroll:
-            self.scroll_changed_in_tab = True
         self.scroll = clamped_scroll
 
         self.browser.measure.stop('render')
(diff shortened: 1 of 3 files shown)


pavpanchekha:
I've confirmed that the compiled `lab13.js` still works great.

src/lab13.py
```diff
@@ -354,120 +355,12 @@ def body(self):
...
-        self.cursor_x += w + font.measureText(" ")
-
-    def input(self, node):
-        w = INPUT_WIDTH_PX
-        if self.cursor_x + w > self.width:
-            self.new_line()
-        line = self.children[-1]
-        input = InputLayout(node, line, self.previous_word)
-        line.children.append(input)
-        self.previous_word = input
```
chrishtr, line 448: self.previous_word is not defined in lab11

chrishtr (commented):
Is it easy to print out the fully inlined versions and diff them? Worried I missed other things.

src/lab13.py
```diff
@@ -1045,8 +776,6 @@ def render(self):
             self.needs_paint = False
 
         clamped_scroll = self.clamp_scroll(self.scroll)
-        if clamped_scroll != self.scroll:
```
chrishtr, line 1048: Why is this no longer needed?

src/lab13.py
pavpanchekha: Correct—this was a simplification we failed to propagate to lab13.

src/lab13.py
pavpanchekha: If you look a line or two down, this `if` statement is duplicated. It doesn't have any effects we need to repeat twice, either. Probably a copy-and-paste bug.

pavpanchekha:
I went through a diff. This PR fixes two more tiny bugs:
- in `BlockLayout.word`, current main calls `get_font(size, weight, size)`, meaning that we always measure roman text even if it's supposed to be italics.
- The space after input areas didn't do pt to px conversion so was too big.
Plus it pulls in a small refactor to `layout_mode` and some changes to dead code / record calls. Basically, I checked the diff and the diff is good.

PR #1268, opened 2024-06-14
Write about fixing iframe scrolling

pavpanchekha:
This PR adds two paragraphs and a code block to describe the iframe scrolling bug fixed in #1264. Also I remove a line that I added in #1264 which isn't needed (the `Tab`, not the `Frame`, sets needs paint).

### DIFF
--- a/book/embeds.md
+++ b/book/embeds.md
@@ -1452,9 +1452,6 @@ class Tab:
         # ...
 ```
 
-It's possible to composite or even thread iframe scrolling, but for
-the sake of expediency we won't do it here.
-
 There's one more subtlety to scrolling. After we scroll, we want to
 *clamp* the scroll position, to prevent the user scrolling past the
 last thing on the page. Right now `clamp_scroll` uses the window
--- a/book/embeds.md
+++ b/book/embeds.md
@@ -1437,6 +1437,24 @@ class Tab:
         self.set_needs_paint()
 ```
 
+If a frame other than the root frame is scrolled, we'll just set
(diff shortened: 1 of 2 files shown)


book/embeds.md
```diff
@@ -1437,6 +1437,24 @@ class Tab:
         self.set_needs_paint()
 ```
 
+If a frame other than the root frame is scrolled, we'll just set
+`needs_composite` so the browser has to reraster from scratch:
```
chrishtr, line 1441:
```suggestion
`needs_composite` so the browser has to re-raster from scratch:
```

book/embeds.md
```diff
@@ -1437,6 +1437,24 @@ class Tab:
...
+        # ...
+        for (window_id, frame) in self.window_id_to_frame.items():
+            if frame == self.root_frame: continue
+            if frame.scroll_changed_in_frame:
+                needs_composite = True
+                frame.scroll_changed_in_frame = False
+        # ...
+```
+
+It's possible to composite or even thread iframe scrolling, but for
```
chrishtr, line 1455: Remove this sentence, because it's redundant with the one on lines 1388-1389.

chrishtr (commented):
Looks like this branch is not up-to-date? Has the other PR inlined into it.

pavpanchekha:
Great, only one paragraph & one code block.

chrishtr:
The branch still seems to be out of date? Shouldn't this one be just the text and not a lot of changes to code?

PR #1271, opened 2024-06-17
Deduplicate chapters 14 and 15

pavpanchekha:
These were easier than 13; mostly some unused functions in the layout classes.


chrishtr (approved):
OK to land, assuming you've diffed them and tested all these features affected on both labs.

pavpanchekha:
Confirmed—some minor changes came for free, I don't think any looked like they had any effect.

PR #1274, opened 2024-06-26
More fix widgets

pavpanchekha:
Fixes broken widgets across all three browsers. The issues all seem to cluster around the order of imports and patch applications, so there were two big parts of the PR:

**Patch compilation**: The biggest part of this PR is a rewrite of how patches are handled in the compiler. Basically, we now handle patches in the same pass as non-patched methods, instead of doing them separately like before. This means we can generate patches in the right place in the source code, which is important to Safari for inscrutible reasons.

**New drivers**: The drivers now use `import` as a statement, which it looks like Safari is somehow happier about. Look—do not ask me why, I do not know, it is a weird bug in Safari that seems to have to do to import timing. But the new code is shorter, nicer, only slightly slower (oh well), and works in Safari and all other browsers too.

As of my just-now careful check, the all widgets now work in all browsers. Yay!

### DIFF
--- a/src/lab14.py
+++ b/src/lab14.py
@@ -36,6 +36,7 @@ from lab9 import EVENT_DISPATCH_JS
 from lab10 import COOKIE_JAR, URL
 from lab11 import FONTS, NAMED_COLORS, get_font, linespace, paint_tree
 from lab11 import parse_color, parse_blend_mode
+from lab11 import BlockLayout, DocumentLayout, TextLayout, LineLayout, InputLayout
 from lab12 import MeasureTime, SingleThreadedTaskRunner, TaskRunner, Chrome
 from lab12 import Task, REFRESH_RATE_SEC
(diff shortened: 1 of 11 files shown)


pavpanchekha:
Update: I rewrote how patches are handled in the compiler (very similar output to before, but now the output is in the right order, which matters for Safari) and that made some progress but Chapters 11 and 16 are still not working in Safari. Will continue tomorrow.

src/lab14.py
```diff
@@ -36,6 +36,7 @@
 from lab10 import COOKIE_JAR, URL
 from lab11 import FONTS, NAMED_COLORS, get_font, linespace, paint_tree
 from lab11 import parse_color, parse_blend_mode
+from lab11 import BlockLayout, DocumentLayout, TextLayout, LineLayout, InputLayout
```
chrishtr, line 39: lab13 patches BlockLayout; is this correct?

pavpanchekha: Good catch

chrishtr (approved):

PR #1275, opened 2024-06-28
Add quiz telemetry

ashton314:
Gets telemetry from quizzes (just a little bit of JSON) and appends it to the end of a file on the server.

This is a *rough draft* and *edits are needed*. Just looking to get some feedback on where to go from here.


ashton314:
Note: this has all the commits from adding quizzes, so **DO NOT MERGE** until #1260 is merged.

infra/api.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/python3
 
 import bottle
+from bottle import response
```
bennn, line 4: why not `bottle.response` below? is this import even used?

ashton314: Fixed it; the language server snuck this import in when I was doing local testing. I've removed it.

pavpanchekha:
We should update the branch and mark it ready for review so Chris can look at it, but this seems like a super safe and easy PR.

pavpanchekha:
Do you know why it still has a diff for everything? Do you need to rebase & force-push or something?

ashton314:
Ach, yes. I'll fix that. I spun this branch off of my other dev branch and they got out of sync.

ashton314:
Ugh—that was a messy rebase. Smaller diff with main now.

I hid the quiz bug report button; I figure if there's an issue, people can submit a comment via the usual mechanism. If you *want* that separate bug reporting mechanism, I can figure out how to enable that in a separate PR.

pavpanchekha (approved):

infra/api.py
```diff
@@ -10,6 +10,8 @@
 
 NOPASSWORD = False
 
+QUIZ_TELEMETRY_FILE = 'quiz_telemetry.txt'
```
pavpanchekha, line 13: Maybe end it with `.json`?

infra/api.py
```diff
@@ -121,6 +123,13 @@ def comment():
     data = json.load(bottle.request.body)
     DATA.chapter_comment(**data)
 
+@bottle.post("/api/quiz_telemetry", method=['OPTIONS', 'POST'])
```
pavpanchekha, line 126: Do quiz questions have some kind of stable ID? If we change the questions how do we track 1) which version the answers are for, and 2) what question it was?

infra/api.py
ashton314: It'll be a bunch of JSON objects, not suitable for parsing as-is. You'd have to wrap brackets around it. IDK if calling that `.json` will mess up any processing tools. I can call it that. Any other changes to that file name/path while I'm changing that?

infra/api.py
ashton314: I don't think there's a versioning system—that's a really good idea though. I can see what I can put together this week.

infra/api.py
pavpanchekha: I think the file name is fine. JSON is sort of a disaster of a format, but I guess we can use `.json` anyway. As you can see earlier in the file, we use a single-object `.json` for some other data, but I agree that loading & saving the whole object isn't a good idea, it'll grow to eventually be super slow.

www/quiz_style.css
```diff
@@ -151,21 +151,21 @@
   overflow-y: hidden;
 }
 .mdbook-quiz .prompt .bug-report {
-  position: absolute;
-  max-width: 70%;
-  right: 0.5em;
-  top: 0.5em;
-  text-align: right;
+  display: none;
+  /* position: absolute; */
```
chrishtr, line 155: If these are going to get commented out, just remove them in this PR

ashton314: Done; I took out all the `.bug-report` styles

PR #1052, opened 2023-11-20
Footnote/margin note workaround

pavpanchekha:
Grabbed Ian's workaround here: https://bugs.chromium.org/p/chromium/issues/detail?id=1499290

PR #1280, opened 2024-07-09
This PR removes the workaround added in #1052

### LINKED ISSUE #1098: Undo whitespace hacks once Chrome stable is shipped
And the fix for this bug is in the stable channel:

https://bugs.chromium.org/p/chromium/issues/detail?id=1499290

pavpanchekha:
This PR closes #1098; Chrome now fixes the bug this code was working around.


chrishtr:
Can you confirm it no longer shows up on the specified page?

pavpanchekha:
Yes, I tried and could not reproduce it, though it's hard to say categorically because the big was always kind of finicky

PR #1283, opened 2024-07-19
Fix error when images don't load

chrishtr:
the `image_url` variable was not set, leading to a crash in the JS widget for `lab15-browser.html` and `lab16-browser.html` on the website.

### DIFF
--- a/book/embeds.md
+++ b/book/embeds.md
@@ -171,7 +171,7 @@ class Tab:
             try:
                 # ...
             except Exception as e:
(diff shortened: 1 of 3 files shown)


pavpanchekha:
Uh oh, another book change. How long is the new line? Does it fit into 70 characters?

chrishtr:
The new line is 62 characters.

I don't think we necessarily need to change the book again, it's not a big deal.

PR #1290, opened 2024-07-26
Add persistent user ID to the telemetry payload

ashton314:
Enhances telemetry by adding a persistent, randomly-generated user ID to each telemetry payload. This helps us relate quiz answers together. The user ID is a UUIDv4 that gets generated for the initial request, then cached in local storage for subsequent request.s

### DIFF
--- a/www/book.js
+++ b/www/book.js
@@ -85,7 +85,34 @@ function setup_close() {
 
 window.addEventListener("DOMContentLoaded", setup_close);
 
+function gen_uuid() {
+  const array = new Uint8Array(16);
+  crypto.getRandomValues(array);
+  array[6] = (array[6] & 0x0f) | 0x40; // Set the 13th character to '4'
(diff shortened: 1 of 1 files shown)


www/book.js
```diff
@@ -85,7 +85,34 @@ function setup_close() {
 
 window.addEventListener("DOMContentLoaded", setup_close);
 
+function gen_uuid() {
```
pavpanchekha, line 88: If you C&P this from elsewhere and you still can find where, good to put a URL in a comment.

pavpanchekha (approved):

www/book.js
ashton314: Hehe… that was generated by GPT-4. I read through it to understand what it does though, of course.

pavpanchekha: Got it.

PR #1294, opened 2024-08-02
Change "socket" to "server" in ex 1-6

jesusgollonet:
original text reads

> if another request is made to the same socket reuse the same socket

but I think you mean

> if another request is made to the same server reuse the same socket


chrishtr:
Looks correct to me. Pavel?

chrishtr:
Thanks for the fix!

PR #1299, opened 2024-08-04
edits from proofreading text.md

chrishtr:


book/text.md
```diff
@@ -91,9 +91,11 @@ type at a fixed size, style, and weight. For example:[^after-tk]
     Tk objects, after calling `tkinter.Tk()`, which is why I'm putting
     this code in the `Browser` constructor.
 
-``` {.python expected=False}
+``` {.python}
 import tkinter.font
+```
 
+``` {.python .example}
```
chrishtr, line 98: This and below are examples, so...

book/text.md
```diff
@@ -446,8 +448,8 @@ any accumulated text as a `Text` object. Otherwise, if you never saw
 an angle bracket, you'd return an empty list of tokens. But unfinished
 tags, like in `Hi!<hr`, are thrown out.[^errortag]
 
-[^errortag]: This may strike you as an odd decision: why not raise an error,
-    or finish up the tag for the author? I don't know, but dropping
+[^errortag]: This may strike you as an odd decision: why
```
pavpanchekha, line 451:
```suggestion
[^errortag]: This may strike you as an odd decision: why not
```

The footnote is trying to say, "If a document ends with `<foo`, one might imagine the browser should silently add a `>` character, so as not to lose data. But that's not what browsers do, they just drop the half-finished tag entirely."

book/text.md
```diff
@@ -1000,7 +1002,7 @@ will have different lengths.[^center-tag]
 
 3-2 *Superscripts*. Add support for the `<sup>` tag. Text in this tag
 should be smaller (perhaps half the normal text size) and be placed so
-that the top of a superscript lines up with the top of a normal
+that the top of a superscript line up with the top of a normal
```
pavpanchekha, line 1005: No, the top of the super script "lines up" with the top of normal text.

book/text.md
```diff
@@ -1013,7 +1015,7 @@ draw a hyphen when you break a word. The word
 "super­cali­fragi­listic­expi­ali­docious"
 is a good test case.
 
-[^entity]: If you've done a [previous exercise](http.md#exercises) on
+[^entity]: If you've done Exercise 1-4 on
```
pavpanchekha, line 1018: Good but let's keep the link.

pavpanchekha (changes requested):

book/text.md
chrishtr: Right. My edit was to remove the part about "raise an error", because we already know that raising an error is a bad design pattern. Whether to silently add `>` OTOH is something that could have been decided on, so it makes sense to leave in the footnote.

book/text.md
chrishtr, line 1005:
```suggestion
that the top of a superscript lines up with the top of a normal
```

book/text.md
chrishtr, line 1018:
```suggestion
[^entity]: If you've done [Exercise 1-4](http.md#exercises) on
```

chrishtr:
Updated, please take a look.

book/text.md
pavpanchekha: Right I'm just saying we have to say "why not finish the tag" since in fact browser's don't finish the tag. I think you just accidentally deleted a word.

pavpanchekha (approved):

PR #1302, opened 2024-08-06
proofreading for styles.md

chrishtr:


pavpanchekha (approved):
Fine but is it important? I'm worried this could change line breaking, which the publisher won't like.

chrishtr:
It's not critical, but it is inconsistent with the main text. Let's land it and then maybe not ask to edit the published book.

PR #1318, opened 2024-08-12
Text chapter changes

pavpanchekha:


book/text.md
```diff
@@ -21,8 +22,8 @@ What is a Font?
 ===============
 
 So far, we've called `create_text` with a character and two
-coordinates to write text to the screen. But we never specified the
-font\index{font}, the size, or the color. To talk about those things,
+coordinates to write text to the screen. But we never specified its
+font\index{font} or size. To talk about those things,
```
pavpanchekha, line 26: So I'm not thrilled with my change by my problem with the original text is that you don't need font objects to change the color. Maybe a better fix is just changing "the color" to "the text style"?

book/text.md
```diff
@@ -88,23 +89,18 @@ Yet Tk's *font objects* correspond to the older meaning of font: a
 type at a fixed size, style, and weight. For example:[^after-tk]
 
 [^after-tk]: You can only create `Font` objects, or any other kinds of
-    Tk objects, after calling `tkinter.Tk()`, which is why I'm putting
-    this code in the `Browser` constructor.
-
-``` {.python}
-import tkinter.font
-```
```
pavpanchekha, line 96: This example/real code split will look terrible in print due to the background for examples. Also if it's example code it can't go in the `Browser` constructors. My proposal here is a totally different example code block, and then `tkinter.font` appears later.

book/text.md
```diff
@@ -217,8 +213,6 @@ of course, since different letters have different widths:[^widths]
 5
 >>> bi_times.measure("!")
 7
->>> 13 + 5 + 7
-25
```
pavpanchekha, line 221: Arguable, but we only refer to this sum in a footnote, so perhaps it doesn't need to be in the main text?

book/text.md
```diff
@@ -915,8 +909,7 @@ logic itself in a new `get_font` function:
 def get_font(size, weight, style):
     key = (size, weight, style)
     if key not in FONTS:
-        font = tkinter.font.Font(size=size, weight=weight,
-            slant=style)
+        font = tkinter.font.Font(size=size, weight=weight, slant=style)
```
pavpanchekha, line 914: The line wrapping is not needed, we have enough space.

pavpanchekha: Update: this line is otherwise 72 characters, so maybe wrapping is indeed needed? But in the book proof there is clearly enough space that we don't need to wrap.

pavpanchekha: This is what's causing tests to fail. I suggest we bump the minimum line width to something like 75 in the checker and suggest the change. But I won't insist on it, it doesn't matter that much.

book/text.md
chrishtr, line 26:
```suggestion
font\index{font}, size or style. To talk about those things,
```

book/text.md
chrishtr: I'd prefer to leave the code as-is and not mess around with line limits, this doesn't look critical to me.

src/lab3.py
```diff
@@ -52,8 +52,7 @@ def lex(body):
 def get_font(size, weight, style):
     key = (size, weight, style)
     if key not in FONTS:
-        font = tkinter.font.Font(size=size, weight=weight,
-            slant=style)
+        font = tkinter.font.Font(size=size, weight=weight, slant=style)
```
chrishtr, line 55: ditto

PR #1323, opened 2024-08-13
Chrome changes

pavpanchekha:


book/chrome.md
```diff
@@ -93,8 +93,8 @@ their own `layout` and `paint` methods, but before we get to those we
 need to think about how the `LineLayout` and `TextLayout` objects will
 be created. That has to happen during word wrapping.
 
-Recall [how word wrapping happens (see Chapter 3)](text.md)
-inside `BlockLayout`'s `word` method. That method updates a `line` field,
+Recall [how word wrapping (see Chapter 3)](text.md)
+inside `BlockLayout`'s `word` method works. That method updates a `line` field,
```
pavpanchekha, line 97: Otherwise I parsed "Recall how X happens" as meaning the same thing as "Recall that X happens" and that made the next sentence weird.

book/chrome.md
```diff
@@ -270,7 +270,7 @@ browser might resolve this paradox with multi-phase layout.
 There are many considerations and optimizations of this kind that are
 needed to make text layout super fast.
 
-That method will pilfer the code from the old `flush` method. First,
+That method will pilfer code from the old `flush` method. First,
```
pavpanchekha, line 273: "The" seems to imply all of it, like this is a command for the reader to do before moving on.

book/chrome.md
```diff
@@ -330,7 +330,7 @@ class TextLayout:
 
 Now we don't need a `display_list` field in `BlockLayout`, and we can
 also remove the part of `BlockLayout`'s `paint`\index{paint} that
-handles it. Instead, `BlockLayout` can just recurse into its children
+handles it. Instead, `paint_tree` can just recurse into its children
```
pavpanchekha, line 333: Probably left over from a refactor. `paint_tree` does the recursion now.

book/chrome.md
```diff
@@ -416,8 +416,7 @@ objs = [obj for obj in tree_to_list(self.document, [])
 ```
 
 In principle there might be more than one layout object in this
-list.^[This is actually not possible in our browser at this point, but
-in real browsers there are all sorts of ways this could happen, like
+list.^[In real browsers there are all sorts of ways this could happen, like
```
pavpanchekha, line 419: It's not true that it's not possible: typically if a node is in the list, so are its parents. It's also not a relevant point here.

book/chrome.md
```diff
@@ -544,6 +543,8 @@ class Browser:
         self.window.bind("<Button-1>", self.handle_click)
 ```
 
+Remove these lines from `Tab`'s constructor.
```
pavpanchekha, line 546: We usually give explicit instruction like this when refactoring, felt like this was missing.

book/chrome.md
```diff
@@ -912,7 +913,7 @@ class Browser:
 
 Note that this display list is always drawn at the top of the window,
 unlike the tab contents (which scroll). Make sure to draw the chrome
-*after* the main tab contents, so that the chrome ends up on top.
+*after* the main tab contents, so that the chrome is drawn over it.
```
pavpanchekha, line 916: We mean "top" in terms of stacking but it reads like it could mean 2D layout.

book/chrome.md
```diff
@@ -1086,10 +1087,10 @@ class Chrome:
...
         cmds.append(DrawOutline(self.address_rect, "black", 1))
         url = str(self.browser.active_tab.url)
-            cmds.append(DrawText(
-                self.address_rect.left + self.padding,
-                self.address_rect.top,
-                url, self.font, "black"))
+        cmds.append(DrawText(
+            self.address_rect.left + self.padding,
+            self.address_rect.top,
+            url, self.font, "black"))
```
pavpanchekha, line 1093: I know why this happened but it is bizarre and incorrect in the actual book.

book/chrome.md
chrishtr: Not critical IMO

book/chrome.md
```diff
@@ -668,7 +669,7 @@ class Browser:
 
 ```
 
-So, let's design the browser chrome. Ultimately, I think it should
+Let's design the browser chrome. Ultimately, I think it should
```
chrishtr, line 672: not critical IMO

book/chrome.md
chrishtr: not critical IMO

chrishtr (approved):

PR #1334, opened 2024-08-15
Fix interaction of rAF with wrap

chrishtr:
`wrap` is not defined until later, after a discussion of the global window object. For now we can just skip that line
and get the point across. The section that defines `wrap` says also "We'll need to call `wrap` any time we use `evaljs`", which implies rAF also.

### DIFF
--- a/book/embeds.md
+++ b/book/embeds.md
@@ -963,7 +963,7 @@ In this code I used a new `dispatch_RAF` method:
 ``` {.python expected=False}
 class JSContext:
     def dispatch_RAF(self):
-        self.interp.evaljs(code)
+        self.interp.evaljs("window.__runRAFHandlers()")
 ```
 
 Note that the `needs_accessibility`, `pending_hover`, and other flags
--- a/book/embeds.md
+++ b/book/embeds.md
@@ -959,11 +959,11 @@ class Tab:
 ```
 
 In this code I used a new `dispatch_RAF` method, which is just like the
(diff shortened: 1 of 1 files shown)


book/embeds.md
```diff
@@ -964,7 +964,6 @@ pre-iframe code but wraps the call for the specified `window_id`:
 ``` {.python}
 class JSContext:
     def dispatch_RAF(self, window_id):
```
pavpanchekha, line 966: But isn't window_id not defined yet?

pavpanchekha: I think the text above this is also out of place and needs fixing...

chrishtr: Fixed PTAL

book/embeds.md
```diff
@@ -958,13 +958,11 @@ class Tab:
...
-pre-iframe code but wraps the call for the specified `window_id`:
+In this code I used a new `dispatch_RAF` method:
 
-``` {.python}
+``` {.python expected=False}
 class JSContext:
-    def dispatch_RAF(self, window_id):
-        code = self.wrap("window.__runRAFHandlers()", window_id)
+    def dispatch_RAF(self):
         self.interp.evaljs(code)
```
pavpanchekha, line 966:
```suggestion
    def dispatch_RAF(self):
        self.interp.("window.__runRAFHandlers()")
```

chrishtr:
fixed

PR #1336, opened 2024-08-16
A typo in Chapter 7

bunhere:
As a non-native english speaker, "and" looks missing.

### DIFF
--- a/book/chrome.md
+++ b/book/chrome.md
@@ -509,7 +509,7 @@ tease tabs and browsers apart.
 Here's the plan: the `Browser` class will own the window and canvas
 and all related methods, such as event handling. And it'll also
 contain a list of `Tab` objects and the page chrome. But the web page
-itself its associated methods will live in a new `Tab` class.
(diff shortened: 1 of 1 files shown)


book/chrome.md
```diff
@@ -509,7 +509,7 @@ tease tabs and browsers apart.
 Here's the plan: the `Browser` class will own the window and canvas
 and all related methods, such as event handling. And it'll also
 contain a list of `Tab` objects and the page chrome. But the web page
```
bunhere, line 511: By the way, can I understand that page chrome and browser chrome mean the same thing?
page chrome is not explained while browser chrome is explained in the section below.
(The term 'chrome' first appeared in this chapter.)

pavpanchekha:
You're right, thanks for finding it. Both @chrishtr and I missed it when doing the proofreading over the last few days!

book/chrome.md
pavpanchekha: This is also a good catch, it should be browser chrome. In the interests of time, I fixed it in #1337

PR #1342, opened 2024-08-23
Fix quizzes so they don't show up in the normal build

pavpanchekha:
A messed-up refactor was causes quizzes to show up, unprocessed, on the "public" version of the book. I already updated the site, this fixes the code.


pavpanchekha:
@ashton314 Ok to merge?

ashton314 (approved):

PR #1343, opened 2024-08-23
Classify what is an example and what isn't

pavpanchekha:
This PR marks some code blocks as examples that incorrectly weren't marked that way. It also marks some example tags as `.output` instead; this is an idea from #1161 that I thought made marking code blocks easier. All of the marks were cross-referenced with the printed book. I think of this both as superseding #1161 and also as preparation for trying to make the style of the online book look more like the printed book.

### DIFF
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -579,6 +579,7 @@ we also need to set dark mode when changing tabs, since all tabs should
 be either dark or light:
 
 ``` {.python}
+class Browser:
    def set_active_tab(self, tab):
         # ...
         task = Task(self.active_tab.set_dark_mode, self.dark_mode)
@@ -1135,7 +1136,7 @@ If you print out `focusable_nodes` for the
 [focus example](examples/example14-focus.html), you should
 get this:
 
-``` {.python .example}
+``` {.python .output}
 [<a tabindex="1" href="/">,
  <button tabindex="2">,
  <div tabindex="3">,
@@ -1353,7 +1354,9 @@ means you can write a selector like this:
 
 [pseudoclass]: https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes
 
-    div:focus { ... }
+``` {.css .example}
+div:focus { ... }
+```
 
 And then that selector applies only to `<div>` elements that are
 currently focused.[^why-pseudoclass]
@@ -1422,7 +1425,9 @@ customize the focus outline itself and not just the element. That can be done
 by adding support for the CSS [`outline` property][outline], which looks like
 this (for a 3-pixel-thick red outline):[^outline-syntax]
 
-    outline: 3px solid red;
+``` {.css .example}
+outline: 3px solid red;
+```
 
 [outline]: https://developer.mozilla.org/en-US/docs/Web/CSS/outline
 
@@ -1695,7 +1700,7 @@ class AccessibilityNode:
 Here is the accessibility tree for the
 [focus example](examples/example14-focus.html):
 
-``` {.text .example}
+``` {.output}
  role=document
    role=button
      role=focusable text
@@ -1965,7 +1970,7 @@ This text construction logic is, of course, pretty naive, but it's
 enough to demonstrate the idea. Here is how it works out for the
 [focus example](examples/example14-focus.html):
 
-``` {.text .example}
+``` {.output}
  role=document text=Document
    role=button text=Button
      role=focusable text text=Focusable text: This is a button
--- a/book/html.md
+++ b/book/html.md
@@ -310,7 +310,7 @@ print_tree(nodes)
 
 You'll see something like this at the beginning:
 
-``` {.example}
+``` {.output}
  <!doctype html>
    '\n'
    <html lang="en-US" xml:lang="en-US">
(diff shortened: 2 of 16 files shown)


book/accessibility.md
```diff
@@ -579,6 +579,7 @@ we also need to set dark mode when changing tabs, since all tabs should
 be either dark or light:
 
 ``` {.python}
+class Browser:
```
pavpanchekha, line 582: This is I think an error in the proof that we *didn't* point out to the publisher. Honestly, it looks kind of fine in the book, not the end of the world, so it might not be worth pointing out anyway (it might repaginate). But we might as well correct it in the online version where it won't repaginate anything.

book/html.md
```diff
@@ -514,16 +514,16 @@ def add_tag(self, tag):
...
+``` {.output}
+ <html>
+    <head>
+      <meta>
+      <link>
+      <link>
+      <link>
+      <link>
+      <link>
+      <meta>
```
pavpanchekha, line 526: Added a space because our printer does do that.

book/scripts.md
```diff
@@ -60,11 +60,8 @@ the `duktape` library that `dukpy` uses.
...
 To test whether you installed DukPy correctly, execute this:
 
-``` {.python}
-import dukpy
-```
-
 ``` {.python .example}
+import dukpy
 dukpy.evaljs("2 + 2")
 ```
```
pavpanchekha, line 66: We ultimately decided to do this in the book.

src/runtime12.js
```diff
@@ -93,10 +93,7 @@ function requestAnimationFrame(fn) {
 }
 
 function __runRAFHandlers() {
-    var handlers_copy = [];
-    for (var i = 0; i < RAF_LISTENERS.length; i++) {
-        handlers_copy.push(RAF_LISTENERS[i]);
-    }
+    var handlers_copy = RAF_LISTENERS;
```
chrishtr, line 96: why this change?

infra/compare.py
```diff
@@ -131,6 +131,7 @@ def compare_files(book, code, language, file):
                 print(f"{block.loc}:", error)
                 failure += 1
             continue
+        if "output" in block.classes: continue
```
chrishtr, line 134: don't we need a change to book.css?

book/accessibility.md
```diff
@@ -1135,7 +1136,7 @@ If you print out `focusable_nodes` for the
 [focus example](examples/example14-focus.html), you should
 get this:
 
-``` {.python .example}
+``` {.python .output}
```
chrishtr, line 1139: I don't fully understand the change from example to output. What impact does it have on the rendered book?

book/html.md
```diff
@@ -514,16 +514,16 @@ def add_tag(self, tag):
...
-   <head>
-     <meta>
-     <link>
-     <link>
-     <link>
-     <link>
-     <link>
-     <meta>
+``` {.output}
+ <html>
```
chrishtr, line 518: why new spaces?

book/scheduling.md
```diff
@@ -170,7 +170,7 @@ we'll implement `setTimeout` by saving the callback in a JavaScript
 variable and creating a handle by which the Python-side code can call
 it:
 
-``` {.javascript file=runtime}
```
chrishtr, line 173: why is file= not needed anymore?

src/runtime12.js
pavpanchekha: Turns out the book had diverged from the source code and it wasn't being caught before.

pavpanchekha: This moves the book code into the code code. I think the book code is nicer.

infra/compare.py
pavpanchekha: Yes—I was planning to do all the style updates in a separate PR but I can do it all in this one if you'd like. Or just make the minimal change in this PR and restyle in another.

book/accessibility.md
pavpanchekha: None, I found it helpful. `example` is code, `output` isn't code. If you don't like it I can make them all `example` but I sometimes find myself asking "is this really example code" and often that's because it's actually output.

book/html.md
pavpanchekha: Our pretty-printer always prints a space between the indentation and the data.

book/scheduling.md
pavpanchekha: `runtime` is the default file.

pavpanchekha: For JavaScript code.

infra/compare.py
chrishtr: style in another PR is fine

src/runtime12.js
chrishtr: This new code is a pointer copy instead of a deep copy. Don't we need a deep copy?

pavpanchekha: Ah great question. It's not a copy at all—it's a move, on the next line we reset `RAF_LISTENERS` to a newly-allocated vector. So what's important here is that `RAF_LISTENERS` and `handlers_copy` are disjoint, and the move is more convenient (and I guess faster, not that we care) than the deep copy & clear.

chrishtr (approved):
Please try out the rAF example from the examples directory for chapter 12 and make sure it works before landing this PR.

pavpanchekha:
rAF examples fixed by #1346, put in pro-forma styling so we can merge this instead of leaving it hanging.

PR #1346, opened 2024-09-13
Fix examples involving script for the JS browser in chapter 9+

chrishtr:
There were several bugs:

1. The lab*.hints files for certain chapters unnecessarily hinted that `open` should map to `filesystem.open`. While that is correct, it caused code in `compiler.py` to detect calls to `open` and insert files on disk into the top of the JS file.
2. lab13.hints had incorrect hints for dispatching JS. Since the code was unchanged from lab9, I just deleted it.
3. lab*-browser.html for chapters 12+ reset needs_animation_frame after calling `render` rather than before. This is incorrect if rAF callbacks set it during `render`'s callouts to JS.
4. localhost was not whitelisted for allowing direct fetching from `rt.js` (only affects debugging flows).

For good measure, I modified rt.js to special-case URLs with "example" in them so that they didn't cause localhost URLs to end up going to `server*.js`. (Alternatively one could run the localhost server at a port other than 8000.)

Now `examples/lab13-opacity-raf.html` works correctly in the JS cross-compiled browser.

### DIFF
--- a/www/widgets/dukpy.js
+++ b/www/widgets/dukpy.js
@@ -16,6 +16,7 @@ addEventListener("message", (e) => {
             val = eval?.(e.data.body);
         } catch (e) {
             console.log('Script crashed');
+            console.log(e);
         }
         if (val instanceof Function || val instanceof Object) val = null;
         $$POSTMESSAGE({"type": "return", "data": val});
--- a/www/widgets/rt.js
+++ b/www/widgets/rt.js
@@ -117,7 +117,8 @@ class socket {
             let [line1] = this.input.split("\r\n", 1);
             let [method, path, protocol] = line1.split(" ");
             this.url = this.scheme + "://" + this.host + path;
-            if (this.host == "localhost" && rt_constants.URLS["local://" + this.port]) {
+            if (this.host == "localhost" && rt_constants.URLS["local://" + this.port] &&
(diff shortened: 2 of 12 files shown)


chrishtr:
?

www/widgets/dukpy.js
```diff
@@ -16,6 +16,7 @@ addEventListener("message", (e) => {
             val = eval?.(e.data.body);
         } catch (e) {
             console.log('Script crashed');
+            console.log(e);
```
pavpanchekha, line 19: Yeah this is fine but we can probably use like `console.trace` or `console.error`, right? Not really worth optimizing I guess.

www/widgets/rt.js
```diff
@@ -117,7 +117,8 @@ class socket {
             let [line1] = this.input.split("\r\n", 1);
             let [method, path, protocol] = line1.split(" ");
             this.url = this.scheme + "://" + this.host + path;
-            if (this.host == "localhost" && rt_constants.URLS["local://" + this.port]) {
+            if (this.host == "localhost" && rt_constants.URLS["local://" + this.port] &&
+                path.indexOf("example") == -1) {
```
pavpanchekha, line 121: Smart

pavpanchekha (approved):

PR #1352, opened 2024-10-20
Ch 11: follow the Rect class introduced in #974

bunhere:
In #974, the Rect class is introduced in Chapter 7 (chrome.md),
but it seems that visual-effects.md hasn't been updated accordingly.

### DIFF
--- a/book/visual-effects.md
+++ b/book/visual-effects.md
@@ -441,8 +441,8 @@ documentation for more on the Skia API.
 ``` {.python replace=%2c%20scroll/,%20-%20scroll/}
 class DrawLine:
     def execute(self, canvas, scroll):
-        path = skia.Path().moveTo(self.x1 - scroll, self.y1) \
-                          .lineTo(self.x2 - scroll, self.y2)
+        path = skia.Path().moveTo(self.rect.left - scroll, self.rect.top) \
+                          .lineTo(self.rect.right - scroll, self.rect.bottom)
         paint = skia.Paint(
             Color=parse_color(self.color),
             StrokeWidth=self.thickness,
@@ -472,8 +472,8 @@ class DrawText:
             AntiAlias=True,
             Color=parse_color(self.color),
         )
-        baseline = self.top - scroll - self.font.getMetrics().fAscent
-        canvas.drawString(self.text, float(self.left), baseline,
+        baseline = self.rect.top - scroll - self.font.getMetrics().fAscent
+        canvas.drawString(self.text, float(self.rect.left), baseline,
             self.font, paint)
 ```
 
@@ -508,21 +508,17 @@ rectangle (e.g., `left`) were checked, replace them with the
 corresponding function on a Skia `Rect` (e.g., `left()`). Also replace
 calls to `containsPoint` with Skia's `contains`.
 
-While we're here, let's also add a `rect` field to the other drawing
-commands, replacing its `top`, `left`, `bottom`, and `right`
-fields:
-
 ``` {.python}
 class DrawText:
     def __init__(self, x1, y1, text, font, color):
+        self.rect = skia.Rect.MakeLTRB(
+            x1, y1, x1 + font.measure(text), y1 + font.metrics("linespace"))
         # ...
-        self.rect = \
-            skia.Rect.MakeLTRB(x1, y1, self.right, self.bottom)
 
 class DrawLine:
     def __init__(self, x1, y1, x2, y2, color, thickness):
-        # ...
         self.rect = skia.Rect.MakeLTRB(x1, y1, x2, y2)
+        # ...
 ```
 
 To create an outline, draw a rectangle but set the `Style` parameter of
--- a/src/lab11.py
+++ b/src/lab11.py
@@ -137,12 +137,10 @@ class DrawRect:
 @wbetools.patch(DrawText)
 class DrawText:
     def __init__(self, x1, y1, text, font, color):
-        self.left = x1
-        self.top = y1
-        self.right = x1 + font.measureText(text)
-        self.bottom = y1 - font.getMetrics().fAscent + font.getMetrics().fDescent
-        self.rect = \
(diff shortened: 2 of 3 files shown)


chrishtr:
Thanks. Can you also update the same classes in lab13?

book/visual-effects.md
```diff
@@ -441,8 +441,8 @@ documentation for more on the Skia API.
 ``` {.python replace=%2c%20scroll/,%20-%20scroll/}
 class DrawLine:
     def execute(self, canvas, scroll):
-        path = skia.Path().moveTo(self.x1 - scroll, self.y1) \
-                          .lineTo(self.x2 - scroll, self.y2)
+        path = skia.Path().moveTo(self.rect.left - scroll, self.rect.top) \
```
chrishtr, line 444:
```suggestion
        path = skia.Path().moveTo(self.rect.left() - scroll, self.rect.top()) \
```

book/visual-effects.md
```diff
@@ -441,8 +441,8 @@ documentation for more on the Skia API.
...
+                          .lineTo(self.rect.right - scroll, self.rect.bottom)
```
chrishtr, line 445:
```suggestion
                          .lineTo(self.rect.right() - scroll, self.rect.bottom())
```

book/visual-effects.md
```diff
@@ -472,8 +472,8 @@ class DrawText:
             AntiAlias=True,
             Color=parse_color(self.color),
         )
-        baseline = self.top - scroll - self.font.getMetrics().fAscent
-        canvas.drawString(self.text, float(self.left), baseline,
+        baseline = self.rect.top - scroll - self.font.getMetrics().fAscent
```
chrishtr, line 475:
```suggestion
        baseline = self.rect.top() - scroll - self.font.getMetrics().fAscent
```

book/visual-effects.md
```diff
@@ -472,8 +472,8 @@ class DrawText:
...
+        canvas.drawString(self.text, float(self.rect.left), baseline,
```
chrishtr, line 476:
```suggestion
        canvas.drawString(self.text, float(self.rect.left()), baseline,
```

book/visual-effects.md
```diff
@@ -508,21 +508,17 @@ rectangle (e.g., `left`) were checked, replace them with the
...
 calls to `containsPoint` with Skia's `contains`.
 
-While we're here, let's also add a `rect` field to the other drawing
-commands, replacing its `top`, `left`, `bottom`, and `right`
-fields:
-
 ``` {.python}
 class DrawText:
     def __init__(self, x1, y1, text, font, color):
+        self.rect = skia.Rect.MakeLTRB(
```
chrishtr, line 514: This doesn't match the source code. You can run `make test` from the commandline to see the differences and fix them.

book/visual-effects.md
chrishtr, line 475: Missing parentheses in function calls.

book/visual-effects.md
chrishtr, line 444: Missing parentheses in function calls.

book/visual-effects.md
bunhere: Ditto.

book/visual-effects.md
chrishtr: Your changes to `lab11.py` change from the `Rect` class to `skia.Rect` `DrawRect` (`DrawLine` already uses it); the `Rect` class is no longer used from chapters 11 onward. So it should be `left()` etc.

The book's build system has checks to make sure the chapter's code is exactly the same as the corresponding labNN.py file, which is how I noticed this discrepancy; that's why I recommended `make test` to debug your PR.

bunhere: Please understand my question for a double check, as I'm still not sure.

In my understanding,
Chapter 11 explains the following step by step.

```md
Rasterizing with Skia
=====

(1) ... To draw a line, you use Skia's path object ...
<--- this code here

... similar to draw text using `drawString`;
<-- another diff

(2) ... the `rect` field need to become a skia `Rect` object. ...

class DrawRect:
   ...

(3) Our browser also needs font metrics and measurements. ...

```

In the current flow of Chapter 11,
I'm not sure if you want to introduce the left() method in this code before paragraph (2), which mentions the Skia Rect change.
Similarly, my patch doesn't modify the font measurements and metrics in the code before paragraph (3).

If you'd like, you can fix it instead of my PR.
I'm happy just to report the issue and help you.

Anyway, I found one of my mistakes and fixed it using `make test`

And, I encountered another JS-related error, so I've attached the result.
visual-effects.md looks fine.
[test_result.txt.zip](https://github.com/user-attachments/files/17501771/test_result.txt.zip)

chrishtr: Good point. I think the text that talks about switching to `skia.Rect` should move up before `DrawLine` to avoid that. I can make that change after merging your PR, and make then update the website with both changes together.

Thanks for the thorough investigation, I really appreciate it!

chrishtr: If you change all of the `left` etc to `left()` now then I can approve the PR. Or if you like I could merge it for you now and then adjust it for you, no problem.

bunhere: > Good point. I think the text that talks about switching to `skia.Rect` should move up before `DrawLine` to avoid that. I can make that change after merging your PR, and make then update the website with both changes together.
> 
> Thanks for the thorough investigation, I really appreciate it!

I absolutely agree with you about that switching.
It will be better to read.

> If you change all of the `left` etc to `left()` now then I can approve the PR. Or if you like I could merge it for you now and then adjust it for you, no problem.

IMHO, I like each commit to be complete, but it doesn't matter in any case.

If you're okay with it, how about accepting this PR and continuing the work?
Alternatively, you can discard this PR and create a new one. I don't mind. :)

chrishtr: > If you're okay with it, how about accepting this PR and continuing the work?

I'm definitely okay with it. Will finish reviewing now.

PR #1362, opened 2024-12-11
Avoid JS syntax errors when visiting our site from our browser

pavpanchekha:
This PR rewrites the JavaScript files `book.js` and `feedback.js` so that they don't raise errors in our own browser. That involved:

- Rewrite `let`, `for of`, and arrow functions into old-school JavaScript
- Use IIFEs for callbacks in loops, just like we did in the 00s
- Test for the existence of `window`, `window.addEventListener`, and `document.addEventListener` before using them
- Only loosely related, but also update the patcher to work in Python 3.13

### DIFF
--- a/www/feedback.js
+++ b/www/feedback.js
@@ -2,6 +2,9 @@
 
 // Thanks for reading the code! You can hit Ctrl-E / Cmd-E to access the feedback tools.
 
+// This code is written in a bit of an odd style to avoid error
+// messages in the WBE browser. There's a lot of old-school JS.
+
 var chapter_overlay;
 
 function ctrl_key_pressed(e) {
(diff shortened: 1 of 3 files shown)


www/feedback.js
```diff
@@ -20,6 +23,7 @@ function ctrl_key_name() {
     }  
 }
 
+if (document.addEventListener)
```
chrishtr, line 26: You don't need to check for the document object?

pavpanchekha: Yep, we define `document` as basically the very first thing in Chapter 9 (so we can define `document.querySelectorAll` so I figure it's OK to assume it exists.

pavpanchekha:
Yes

PR #1363, opened 2024-12-11
Fix a typo in `lab6`

pavpanchekha:
This PR fixes a small bug in lab6 that made all the lines appear indented. We changed `cursor_x` from absolute to relative coordinates and didn't update lab6. Checked—lab7 does it right.

### DIFF
--- a/src/lab6.py
+++ b/src/lab6.py
@@ -233,7 +233,7 @@ class BlockLayout:
             x = self.x + rel_x
             y = self.y + baseline - font.metrics("ascent")
             self.display_list.append((x, y, word, font, color))
-        self.cursor_x = self.x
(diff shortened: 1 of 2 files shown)


src/lab6.py
```diff
@@ -233,7 +233,7 @@ def flush(self):
             x = self.x + rel_x
             y = self.y + baseline - font.metrics("ascent")
             self.display_list.append((x, y, word, font, color))
-        self.cursor_x = self.x
+        self.cursor_x = 0
```
chrishtr, line 236: Can you add a unittest?

pavpanchekha:
Done, tested the unit test both ways.

chrishtr (approved):

PR #1364, opened 2024-12-11
Make the website look a little nicer in the WBE browser

pavpanchekha:
This PR uses a few weird parser tricks to make the browser look a little better in lab6.py, including overriding the (too dark) background for `<pre>` tags and also making code in pre tags not super small. I also made sure that, if the user implemented class selectors, that footnotes would appear in gold like they're supposed to.

### DIFF
--- a/www/book.css
+++ b/www/book.css
@@ -1,3 +1,18 @@
+/* In a few places, this file has to use parser tricks to make the WBE
+ * browser do something different from a real browser. The WBE browser
+ * supports:
+ *
+ * Ch 6: font-{size,weight,style}, color, background-color
+ * Ch 11: border-radius, opacity, mix-blend-mode, overflow
+ * Ch 13: transform, transition
+ * Ch 14: outline
+ * Ch 15: image-rendering
+ *
+ * Of these, basically only the first four (fonts & colors) matter.
+ * Some effort has been put toward making class selectors do something
+ * nice: highlighting info boxes, syntax highlighting, and making
+ * footnotes stand out. */
+
 html { font-size: 24px; line-height: 1.3; padding: 0.5ex; }
 body {
     max-width: 65ex; margin: 0 auto; font-family: 'Lora', 'Times', sans-serif;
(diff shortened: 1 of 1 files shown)


www/book.css
```diff
@@ -13,7 +28,11 @@ body {
 }
 
 pre, code { hyphens: none; -webkit-hyphens: none; }
-pre { font-size: 18px; overflow: auto; padding-left: 2ex; }
+pre {
+    font-size: 18px; overflow: auto; padding-left: 2ex;
+    background-color: #ddd; /* For WBE browser only */
```
chrishtr, line 33: Does the WBE browser properly ignore comments?

www/book.css
```diff
@@ -30,7 +49,7 @@ header, h1, strong {
     color:black;
 }
 }
-h1 { font-size: 105%; margin: 1.5em 0 0; text-align: left; }
+h1 { font-size: 105%; font-weight: bold; margin: 1.5em 0 0; text-align: left; }
```
chrishtr, line 52: Why this change?

www/book.css
```diff
@@ -180,35 +201,33 @@ body.main { padding: 0 4ex; }
     font-weight: bold; font-family: 'Vollkorn', 'Garamond', serif;
     text-align: center;
 }
-p
```
chrishtr, line 183: What was this p?

www/book.css
pavpanchekha: Ah—our browser's default style sheet doesn't bold `h1` tags by default, figured this was safe to add since all other browsers do.

www/book.css
pavpanchekha: No, or, uhh, sortof. It doesn't have handling for comments, but when it sees a comment it treats it as a syntax error which causes it to eat everything up to the next semicolon. So this comment sort of deletes the next property (`background: transparent`). This doesn't change how our browser behaves because it doesn't support `background` anyway.

PR #1365, opened 2024-12-13
Website style to match printed book

pavpanchekha:
This PR updates the website CSS to match the printed book. Here's a side-by-side, with the old one on the left and the new one on the right. (Click to zoom.)

image

Changes include:

- Change the default body and code fonts to match the book (using Crimson Pro instead of Minion 3)
- Change spacing; paragraphs are now indented instead of having margins in between
- Change text width, mostly to be wider, and change headings to be centered
- Change what "info boxes" look like, with thinner borders, underlined titles (except Go Further)

However, in some places I diverged from the book:

- Go furthers have a border-all-around because I'm using `outline`, which can't be customized side-by-side. This is because our browser supports `outline` in later chapter, and using it allows the book to look nicer in the WBE browser
- The book makes code kinda small, I didn't do that on the web. But we could I guess
- I kept the online version more colorful, including info boxes, links, footnotes, and a couple of other things being colored.


pavpanchekha:
@chrishtr I think the best review / feedback is to look at the screenshot in detail and say which one you like more.

chrishtr:
The new look seems mostly good to me. Feedback:

1. There isn't enough minimum margin on the sides for narrow browser windows.
2. The printed book doesn't do it, but adding a bit of space between paragraphs would make things look less over-dense and easier to read.
3. The self link hash mark (`#`) on section titles is now gone, re-add it?

pavpanchekha:
All noted issues fixed. I also add all the prefetch styles like you did for the old style, and also updated the "thanks" page, which I missed last time. I think this is ready to merge

infra/template.html
```diff
@@ -2,24 +2,24 @@
...
   <meta name="color-scheme" content="dark light">
-  <meta name="generator" content="pandoc" />
   <meta name="viewport" content="width=device-width,initial-scale=1.0,user-scalable=yes" />
+
+  <link rel="preconnect" href="https://fonts.googleapis.com">
+  <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+  <link href="https://fonts.googleapis.com/css2?family=Crimson+Pro:ital,wght@0,200..900;1,200..900&family=Inconsolata:wght@200..900&display=swap" rel="stylesheet">
+  <link rel="prefetch" crossorigin href="https://fonts.gstatic.com/s/crimsonpro/v24/q5uBsoa5M_tv7IihmnkabARekYNwDeChrlU.woff2" as=font type="font/woff2">
+  <link rel="prefetch" crossorigin href="https://fonts.gstatic.com/s/crimsonpro/v24/q5uDsoa5M_tv7IihmnkabARboYF6CsKj.woff2" as=font type="font/woff2">
+  <link rel="prefetch" crossorigin href="https://fonts.gstatic.com/s/inconsolata/v32/QlddNThLqRwH-OJ1UHjlKENVzkWGVkL3GZQmAwLyya15IDhunA.woff2" as=font type="font/woff2">
```
chrishtr, line 13: It needs to be preload, not prefetch.

chrishtr: Also, you can tell it works by reloading and checking for zero flash of fallback fonts.

pavpanchekha:
Done. I also saw we weren't async-loading the JS, which is inessential, so I fixed that too.

chrishtr:
Add a couple more missing preloads and removed the async attribute (defer is enough and async overrides it).

PR #1368, opened 2024-12-23
Fix padding around go-further/quirk/example/etc blocks

chrishtr:
The change to use outlines made the padding between the text and the outline box too small. This PR fixes that.

I also changed the go further title to look the same as quirk, example etc. I don't think there is a good reason to have it look different?


chrishtr:
Going to merge this because the site is a bit broken at present.

PR #1369, opened 2024-12-24
Add feature to deny all feedback for a given user

chrishtr:
The WBE site currently has a lot of spam from a user named "pHqghUme". Either someone attacking our server or a script gone rogue...

So I added this feature to clean it up with one click.

### DIFF
--- a/infra/api.py
+++ b/infra/api.py
@@ -104,6 +104,7 @@ class Data:
            for d in self.data:
                 if d['name'] == self.data[i]['name']:
                     d['status'] = 'denied'
+            return
         self.data[i]['status'] = status
         self.save()
 
--- a/infra/api.py
+++ b/infra/api.py
@@ -100,6 +100,10 @@ class Data:
         }
 
     def set_status(self, i, status):
+        if status == "denied-all":
+           for d in self.data:
+                if d['name'] == self.data[i]['name']:
+                    d['status'] = 'denied'
         self.data[i]['status'] = status
         self.save()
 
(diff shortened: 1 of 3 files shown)


infra/api.py
```diff
@@ -100,6 +100,10 @@ def contributors(self):
         }
 
     def set_status(self, i, status):
+        if status == "denied-all":
+           for d in self.data:
+                if d['name'] == self.data[i]['name']:
+                    d['status'] = 'denied'
         self.data[i]['status'] = status
```
pavpanchekha, line 107: Doesn't this then set the one you denied-all to have a weird new status?

www/entries.js
```diff
@@ -10,7 +10,7 @@ function feedback_mode() {
         elts[i].addEventListener('click', function() {
             var pw = window.localStorage["pw"];
             submit_status(parseInt(this.parentNode.dataset.id), this.className, pw);
-            this.parentNode.remove();
+            location.reload();
```
pavpanchekha, line 13: I assume this is fast enough to not be annoying?

pavpanchekha (approved):

www/entries.js
chrishtr: > I assume this is fast enough to not be annoying?

I think it should be. Currently there are a ton of entries on the feedback page and it loads pretty fast.

infra/api.py
chrishtr: Oops, forgot a return statement. Added

PR #1370, opened 2024-12-24
Fix is_focused invalidation for chapters 8-10

chrishtr:
A reader found this bug. Previously the is_focused code would do nothing because focus was set to `None` right at the beginning of `click`.

### DIFF
--- a/book/forms.md
+++ b/book/forms.md
@@ -360,8 +360,9 @@ sheets[^update-styles] every time you type!
     area definitely can't make such changes, so let's skip this in our
     browser.
     
-Now when we click an `input` element and clear its contents, we can
-call `render` to redraw the page with the `input` cleared:
+Now when we click an `input` element and clear its contents, we should
+call `render` to redraw the page with the `input` cleared. To make things
+easier, I'm just always calling `render` regardless of whether things changed:
 
 ``` {.python}
 class Tab:
@@ -369,7 +370,8 @@ class Tab:
         while elt:
             elif elt.tag == "input":
                 elt.attributes["value"] = ""
-                return self.render()
+                break
+        self.render()
 ```
 
 So that's clicking in an `input` area. But typing is harder. Think
@@ -510,14 +512,13 @@ element:
 ``` {.python}
 class Tab:
     def click(self, x, y):
+        if self.focus:
+            self.focus.is_focused = False
         while elt:
             elif elt.tag == "input":
                 elt.attributes["value"] = ""
-                if self.focus:
-                    self.focus.is_focused = False
                 self.focus = elt
                 elt.is_focused = True
-                return self.render()
 ```
 
 Note that we have to un-focus the currently focused element,
--- a/src/lab8-tests.md
+++ b/src/lab8-tests.md
@@ -172,6 +172,7 @@ Testing focus
 
     Clicking back on the content area unfocuses it
     >>> browser.handle_click(test.Event(200, 200))
+    Ignoring HTML contents inside button
     >>> browser.focus
     'content'
     >>> browser.chrome.focus
@@ -183,6 +184,7 @@ Testing focus
     >>> browser.chrome.focus
 
     >>> browser.handle_click(test.Event(200, 200))
+    Ignoring HTML contents inside button
     >>> browser.focus
     'content'
     >>> browser.chrome.focus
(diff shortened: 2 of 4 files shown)


book/forms.md
```diff
@@ -360,16 +360,18 @@ sheets[^update-styles] every time you type!
...
 
 ``` {.python}
 class Tab:
     def click(self, x, y):
         while elt:
             elif elt.tag == "input":
                 elt.attributes["value"] = ""
-                return self.render()
+                break
+        self.render()
```
pavpanchekha, line 374: How does this work with form submission? I think we'll end up rendering twice, once when loading the new page and once again when returning back to the click handler.

src/lab8-tests.md
```diff
@@ -172,6 +172,7 @@ Testing focus
 
     Clicking back on the content area unfocuses it
     >>> browser.handle_click(test.Event(200, 200))
+    Ignoring HTML contents inside button
```
pavpanchekha, line 175: Why do these change?

src/lab8.py
```diff
@@ -279,12 +279,14 @@ def render(self):
...
     def click(self, x, y):
+        if self.focus:
+            self.focus.is_focused = False
         self.focus = None
         y += self.scroll
         objs = [obj for obj in tree_to_list(self.document, [])
                 if obj.x <= x < obj.x + obj.width
                 and obj.y <= y < obj.y + obj.height]
-        if not objs: return
+        if not objs: return self.render()
```
pavpanchekha, line 289: I think this is the core of the bug fix, and it's a good catch.

pavpanchekha (commented):
I like the bug fix but I'm not sure why we need to move `self.render()` around and I think it might cause issues.

book/forms.md
chrishtr: Yes, but is that so bad? It's just a bit redundant, right?

src/lab8-tests.md
chrishtr: Because calling `render` can cause `InputLayout.paint` to be called, and
we're now calling `render` even if nothing was hit tested (to cover situations
where you have an input focused and then click in an empty location on the
page). Line 287 in the code.

src/lab8.py
chrishtr: Yes, but also cases where the input element is not in the ancestor chain?

book/forms.md
```diff
@@ -360,16 +360,18 @@ sheets[^update-styles] every time you type!
     area definitely can't make such changes, so let's skip this in our
     browser.
     
-Now when we click an `input` element and clear its contents, we can
-call `render` to redraw the page with the `input` cleared:
+Now when we click an `input` element and clear its contents, we should
+call `render` to redraw the page with the `input` cleared. To make things
+easier, I'm just always calling `render` regardless of whether things changed:
```
pavpanchekha, line 365:
```suggestion
call `render` to redraw the page with the `input` cleared. We also need
to call `render` if we clicked off an input element, since we might have unfocused
an input element in the process:
```

book/forms.md
```diff
@@ -510,14 +512,13 @@ element:
...
+        if self.focus:
+            self.focus.is_focused = False
         while elt:
             elif elt.tag == "input":
                 elt.attributes["value"] = ""
-                if self.focus:
-                    self.focus.is_focused = False
                 self.focus = elt
                 elt.is_focused = True
-                return self.render()
```
pavpanchekha, line 520: Keep this one line actually

src/lab8.py
```diff
@@ -294,17 +296,16 @@ def click(self, x, y):
                 return self.load(url)
             elif elt.tag == "input":
                 elt.attributes["value"] = ""
-                if self.focus:
-                    self.focus.is_focused = False
                 self.focus = elt
                 elt.is_focused = True
-                return self.render()
+                break
```
pavpanchekha, line 301:
```suggestion
                return self.render()
```

src/lab9.py
```diff
@@ -146,14 +148,15 @@ def click(self, x, y):
                     self.focus.is_focused = False
                 self.focus = elt
                 elt.is_focused = True
-                return self.render()
+                break
```
pavpanchekha, line 151:
```suggestion
                return self.render()
```

pavpanchekha (approved):
Ok, I reread the patch and the original code, and I think I now agree that the proposed patch is correct and my objections were off base. I'd prefer to make the code a little less tricky by making the extra `self.render()` only handle null clicks, but either way we can merge.

PR #1372, opened 2024-12-24
Rename the nodes field on browser/tab to node

chrishtr:
...since it's just a root node.

A reader got confused about it since `DocumentLayout`'s constructor takes a node but it looks like it is being passed `nodes`.

### DIFF
--- a/book/accessibility.md
+++ b/book/accessibility.md
@@ -601,7 +601,7 @@ class Tab:
                 INHERITED_PROPERTIES["color"] = "white"
             else:
                 INHERITED_PROPERTIES["color"] = "black"
-            style(self.nodes,
+            style(self.node,
                 sorted(self.rules, key=cascade_priority))
 ```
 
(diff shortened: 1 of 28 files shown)


pavpanchekha:
I'm not sure this change is an improvement.

I agree that `nodes` is not a good name, but `node` maybe isn't either? I think the name of the HTML and layout tree fields have to be considered together:

- Currently `self.nodes` is the HTML tree and `self.document` is the layout tree.
- From the POV of the `Frame` it's the whole tree, not a single node, an ideal name would reflect that.
- In a real browser the "document" is the HTML tree.

Probably an idea naming scheme is, like, `self.document` for the HTML tree and `self.layout` for the layout tree? (I think I recall I didn't want to use `layout` as a field name because there's a method on layout objects called `layout`, but still seems better than what we have now.)

Anyway, unless there's a good reason to make this change I don't think it really helps consistency and is maybe more confusing (in that it's singular).

chrishtr:
> I agree that `nodes` is not a good name, but `node` maybe isn't either? I think the name of the HTML and layout tree fields have to be considered together:

How about `self.html` or `self.html_tree`? I think the main thing is to avoid a plural name, which makes it sound like a list.

pavpanchekha:
Yes, agree on both counts. I think ideally we'd rename `self.document`, which is currently the layout tree, at the same time. `self.html_tree` and `self.layout_tree` work, `self.document` and `self.layout` also work, happy to pick either one.

PR #1375, opened 2024-12-24
Rename containsPoint to contains_point

chrishtr:
It was inconsistent.

Spotted by a reader.


pavpanchekha:
Yeah I guess we use underscores normally. It's not called `contains` because of issues in our JS compiler right?

pavpanchekha (approved):

chrishtr:
> It's not called `contains` because of issues in our JS compiler right?

Right

PR #1378, opened 2024-12-27
Specify connection: keep-alive in exercise 1-6

chrishtr:
Noted by a reader.

### DIFF
--- a/book/http.md
+++ b/book/http.md
@@ -984,11 +984,11 @@ print the entire HTML file as if it was text. You'll want to have also
 implemented Exercise 1-4.
 
 1-6 *Keep-alive*. Implement Exercise 1-1; however, do not send
-the `Connection: close` header. Instead, when reading the body from
-the socket, only read as many bytes as given in the `Content-Length`
-header and don't close the socket afterward. Instead, save the
-socket, and if another request is made to the same server reuse the
-same socket instead of creating a new one. This will speed up repeated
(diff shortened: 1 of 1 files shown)


book/http.md
```diff
@@ -984,12 +984,14 @@ print the entire HTML file as if it was text. You'll want to have also
...
-header and don't close the socket afterward. Instead, save the
-socket, and if another request is made to the same server reuse the
-same socket instead of creating a new one. This will speed up repeated
-requests to the same server, which are common.
+the `Connection: close` header (send `Connection: keep-alive` instead).
+When reading the body from the socket, only read as many bytes as given
+in the `Content-Length` header and don't close the socket afterward.
+Instead, save the socket, and if another request is made to the same server
+reuse the same socket instead of creating a new one. 
+(You'll also need to pass the `"rb"` option to `makefile` or the value reported
```
chrishtr, line 992: This was also reported by two readers who got stuck for a long time on exercise 1-6 because of this.
Is there a reason we're using r and not rb? I recall we discussed it but don't remember the conclusion.

pavpanchekha: Uhh, we switch to `rb` in Chapter 15 but don't here to avoid a discussion of encodings. But, you know, I'm not, like, so dedicated to this decision. We have little notes about encodings anyway, so maybe trying to skip it wasn't sensible anyway.

pavpanchekha (approved):

PR #1406, opened 2025-04-08
Two minor tweaks

pavpanchekha:
- The `round` CSS function avoids a tiny white gap in widgets caused by iframes rounding to the nearest device pixel size
- The hint fixes compilation for Chapter 16. All other chapters already compile cleanly.

### DIFF
--- a/www/book.css
+++ b/www/book.css
@@ -224,7 +224,7 @@ img { max-width: 100%; }
 .demo:before { content: "Demonstration"; color: gray; }
 .further { outline: 2px solid green; }
 .further:before { content: "Go further"; font-weight: bold; color: green; }
-.widget { width: 100%; border: 2px solid navy; }
+.widget { width: round(down, 100%, 1px); border: 2px solid navy; box-sizing: border-box; }
 .center { text-align: center }
 .example, .output { outline: 2px solid lightgray; padding: 1em }
 
--- a/src/lab16.hints
+++ b/src/lab16.hints
@@ -22,5 +22,6 @@
 {"code": "DEFAULT_STYLE_SHEET.copy()", "js": "constants.DEFAULT_STYLE_SHEET.slice()"},
 {"code": "node.attributes.get('rel') == 'stylesheet'", "js": "node.attributes['rel'] == 'stylesheet'"},
 {"code": "CSS_PROPERTIES", "type": "dict"},
-{"code": "self.node.attributes", "type": "dict"}
+{"code": "self.node.attributes", "type": "dict"},
+{"code": "img.attributes", "type": "dict"}
 ]


www/book.css
```diff
@@ -224,7 +224,7 @@ img { max-width: 100%; }
 .demo:before { content: "Demonstration"; color: gray; }
 .further { outline: 2px solid green; }
 .further:before { content: "Go further"; font-weight: bold; color: green; }
-.widget { width: 100%; border: 2px solid navy; }
+.widget { width: round(down, 100%, 1px); border: 2px solid navy; box-sizing: border-box; }
```
chrishtr, line 227: Can you give me an example that this fixes? I'd like to investigate.

pavpanchekha: Sure. Here's a screenshot of a random chapter's widget on my M1 Macbook Pro:

image

The link is here: https://browser.engineering/layout.html#summary

Here's what's going on: the widget itself is an `iframe`; prior to this change it was given `width: 100%` making it 614.66 pixels wide:

image

Subtract the two-pixel border and the contents of the `iframe` should be 610.66 pixels wide—but they're not:

image

I assume that iframes round their dimensions to integer multiples of the display pixels, which is what leads to the 610.5px width. But this leaves a gap of .16 CSS pixels = .32 device pixels which is drawn in in white.

By rounding down the size of the iframe, there's no gap between iframe content and iframe border.

I'll note that one thing I wondered was whether positions are also rounded (to align the pixel grid of the iframe with the parent, I thought maybe this mattered for some kind of raster stuff) but it doesn't.

chrishtr: Is the 1px gap in the iframe the gap between the top blue bar and the RHS of the iframe's border? I see that in your screenshot but not Chrome (even with a non-integral device pixel ratio or zoom).

pavpanchekha: Yes, that's the gap I meant. I got this screenshot from Arc, which is a kinda bootleg Chrome at this point. But! I **did** reproduce this in normal Chrome. You're not going to like the steps:

1. Go to the web page, open the dev tools
2. Zoom in on the web page. Zoom a non-integer amount, like 175%
3. Resize the dev tools, making sure that the width of the page is non-integer
4. Click in the page content to focus it
5. Zoom way out, like maybe 90%

Sometimes the gap reappears, sometimes not, it probably depends on the exact sizes and who knows what else, but I can usually get it to appear after 2–3 tries.

image

This was on Version 135.0.7049.84 (Official Build) (arm64); I'm updating Chrome and will report back.

pavpanchekha: (You can click on the screenshot above and zoom in to see the gap)

pavpanchekha: Updated Chrome and just zooming out to 90%, with no dev tools nonsense, is enough to reproduce.

pavpanchekha: Current version is Version 135.0.7049.85 (Official Build) (arm64); new repro steps are just to go to that page & zoom out to 90%.

pavpanchekha:
Shall we merge this?

chrishtr:
I still want to debug it - next week I will have time.

PR #1440, opened 2025-09-10
Move code and advice about condition.wait() until there are two threads

### LINKED ISSUE #1438: Chapter 12: Waiting issue just after implemented setTimeout
Now I'm stepping through the code.

Chapter 12.2 introduces setTimeout with locking.
My understanding was that the toy browser should run once 12.2 was completed,
because a comment said "Try removing this code and observe. The timers will become quite erratic."

However, the browser didn't behave correctly.

I think this is because the taskRunner isn't running on a separate thread yet (at least up to Chapter 12.2),
and task_runner.run() is called from the main loop.
As a result, calling wait() blocks the main thread.

I'm not sure how to fix this, but I just reported.
Please let me know if I am wrong.

Attached lab12.2.py that implements 12.1 and 12.2.

[lab12.2.py](https://github.com/user-attachments/files/22188913/lab12.2.py)

chrishtr:
Fixes #1438.

### DIFF
--- a/book/scheduling.md
+++ b/book/scheduling.md
@@ -303,24 +303,13 @@ class TaskRunner:
             task.run()
 
         self.condition.acquire(blocking=True)
-        if len(self.tasks) == 0:
-            self.condition.wait()
         self.condition.release()
 ```
-
-It's important to call `wait` at the end of the `run` loop if there is nothing
-left to do. Otherwise that thread will tend to use up a lot of the CPU,
-plus constantly be acquiring and releasing `condition`. This busywork not only
-slows down the computer, but also causes the callbacks from the `Timer` to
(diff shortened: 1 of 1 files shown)


pavpanchekha (approved):
Full agreement

book/scheduling.md
```diff
@@ -303,24 +303,13 @@ class TaskRunner:
             task.run()
 
         self.condition.acquire(blocking=True)
```
pavpanchekha, line 305: Hey @chrishtr, why do we acquire and release the condition here? It's still in mainline.

chrishtr: Looks like an oversight in my PR. Since at this point of the chapter there aren't two threads, we don't need to do this.

PR #1444, opened 2025-09-17
focus_element is used before its introduction in accessibility.md

bunhere:
focus_element is introduced in 14.6, but it is used in 14.5.


chrishtr:
Thanks for the bug fix; I've copied this into another PR that also makes the tests pass: https://github.com/browserengineering/book/pull/1445

Issue #1340, opened 2024-08-18
No "Go Further" blocks for two sections in Chapter 9.


pavpanchekha:
Sections 9.4 (Handling Crashes) and 9.5 (Returning Handles) don't have "Go Further" blocks after them, the only two such sections in the book's main content. We should write some, though not before the first printing.

pavpanchekha:
Fixed by #1455 and #1456

PR #1455, opened 2025-10-07
Add a "Go Further" block about devtools

pavpanchekha:
This addresses half of #1340 and was interesting historically. I learned something doing the research, I didn't realize that Chrome was the first browser to ship them by default.


book/scripts.md
```diff
@@ -362,6 +362,23 @@ Python and JavaScript get pretty complicated. *Because* these bugs are hard,
 it's worth approaching debugging systematically and gathering a lot of
 information before attempting a fix.
 
+::: {.further}
```
chrishtr, line 365: Does this duplicate the Browser UIs and Developer Tools section of skipped.md?

PR #1460, opened 2025-10-30
Update the online book to Skia 138

pavpanchekha:
This has been in the works for a while. The published book (the 1st edition) used Skia 87, basically because the `skia-python` library didn't put out full releases between Skia 87 and Skia 138 (it only put out "pre-releases"). This is important because `skia-python` is a binary package and thus requires a new build for every Python release. At our request, there were releases with Skia 87 out through Python 3.13, but putting out new ones was becoming increasingly untenable because that version of Skia was getting increasingly old.

Recently, thanks to the heroic efforts of @HinTak, there's a full, normal release of Skia 138. @HinTak doesn't plan to release any more `skia-python` builds with Skia 87, so for the most recent Python release (3.14) you can no longer easily install `skia-python` with Skia 87.

The reason this matters to us is that WBE used an "experimental" API in Skia 87, the `FilterQuality` API. Somewhere in between Skia 87 and Skia 138, that API went away and was replaced by a stable `SamplingOptions` API. So if you take a 1st edition printed copy, you can no longer install the correct version of Skia, and if you install a modern one you can no longer get some of the code in the book to work. (There is, luckily, a note in the book to this effect, since we knew this was coming.) The good news is that, now that this API is stable, we probably no longer need to specify a Skia version and don't expect any similar changes in the future.

This PR updates the website to now show the Skia 138 code. I've added footnotes everywhere it matters noting the differences from the printed version (in case someone buys the book and then tries to copy the code from online). The footnotes point to a porting page that shows the old code (for the same use case).

### DIFF
--- a/book/animations.md
+++ b/book/animations.md
@@ -253,7 +253,8 @@ class Browser:
 
         sdl2.SDL_GL_SetAttribute(sdl2.SDL_GL_CONTEXT_MAJOR_VERSION, 3)
         sdl2.SDL_GL_SetAttribute(sdl2.SDL_GL_CONTEXT_MINOR_VERSION, 2)
-        sdl2.SDL_GL_SetAttribute(sdl2.SDL_GL_CONTEXT_FORWARD_COMPATIBLE_FLAG, True)
+        sdl2.SDL_GL_SetAttribute(
+            sdl2.SDL_GL_CONTEXT_FORWARD_COMPATIBLE_FLAG, True)
         sdl2.SDL_GL_SetAttribute(sdl2.SDL_GL_CONTEXT_PROFILE_MASK,
                                  sdl2.SDL_GL_CONTEXT_PROFILE_CORE)
         self.gl_context = sdl2.SDL_GL_CreateContext(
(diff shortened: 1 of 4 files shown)


book/embeds.md
```diff
@@ -244,29 +244,32 @@ higher-quality "bilinear" or even "[Lanczos][lanczos]" algorithms.[^resizing]
 To give web page authors control over this performance bottleneck,
 there's an [`image-rendering`][image-rendering] CSS property that
 indicates which algorithm to use. Let's add that as an argument to
-`DrawImage`:
+`DrawImage`:[^skia-87]
+
+[^skia-87]: The 1^st^ edition of *Web Browser Engineering* used an
```
chrishtr:
```suggestion
[^skia-87]: The 1^st^ printed edition of *Web Browser Engineering* used an
```

pavpanchekha:
Neither here nor there but also the `playsound` library, which we use in the Accessibility chapter, seems to have gone unmaintained in the meanwhile; we'll want, I think, to just remove it from the book, it was kinda annoying anyway and not super central (unlike Skia). If you agree, @chrishtr, I'll make a PR to remove it from the book.

HinTak:
If you could either work on those two issues yourself,  or try to find funding for me to do that paid, would really appreciate. There is a workaround for the pybind11 3.x issue, by adding a little snipplet which limits pybind11 to 2.x and not use the latest 3.x in the history of the m14x pulls where this is turned on and off to test if latest 3.x works.

HinTak:
M142 is already ready in my hard disk, waiting for those two issues.

HinTak:
I don't want to release m142 with a pybind11 workaround.

